PR #1, opened 2017-05-20
add proper context support

thehydroimpulse:
cc @achille-roussel 

I'll add some tests to cover the cancelation.

### DIFF
--- a/kafka.go
+++ b/kafka.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/Shopify/sarama"
@@ -44,7 +43,7 @@ type kafkaReader struct {
 	// Cancel the async task
 	cancel context.CancelFunc
 
-	// The current offset that is synchronized via atomics.
+	// The current offset that is updated by `Read`
 	offset int64
 
 	// kafka fetch configuration
@@ -109,6 +108,9 @@ func (kafka *kafkaReader) Read(ctx context.Context) (Message, error) {
 	case <-ctx.Done():
 		return Message{}, ctx.Err()
 	case msg := <-kafka.events:
+		// Update the local offset to the current message.
+		kafka.offset = msg.Offset
+
 		return msg, nil
 	case err := <-kafka.errors:
 		return Message{}, err
@@ -173,14 +175,12 @@ func (kafka *kafkaReader) Seek(ctx context.Context, offset int64) (int64, error)
 		return 0, err
 	}
 
-	atomic.StoreInt64(&kafka.offset, offset)
-
 	asyncCtx, cancel := context.WithCancel(ctx)
 	kafka.cancel = cancel
 
 	kafka.asyncWait.Add(1)
 
-	go kafka.fetchMessagesAsync(asyncCtx, kafka.events, kafka.errors)
+	go kafka.fetchMessagesAsync(asyncCtx, kafka.events, kafka.errors, offset)
 	kafka.spawned = true
 
 	return offset, nil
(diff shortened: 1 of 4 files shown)


kafka.go
```diff
@@ -194,84 +189,141 @@ func (kafka *kafkaReader) Seek(ctx context.Context, offset int64) (int64, error)
...
-			return offset, errors.Wrap(err, "fetching available offsets failed")
+			errs <- errors.Wrap(err, "fetching available offsets failed")
+			return
 		}
 
 		if block.Offset != 0 {
-			return block.Offset - 1, nil
+			val <- block.Offset - 1
+		} else {
+			val <- block.Offset
```
achille-roussel: It seems like all the sends on channels in this goroutine would block if the context is canceled, causing the goroutine to leak.

kafka.go
```diff
@@ -168,12 +168,7 @@ func (kafka *kafkaReader) Seek(ctx context.Context, offset int64) (int64, error)
 		kafka.closeAsync()
 	}
 
-	broker, err := kafka.client.Leader(kafka.topic, kafka.partition)
-	if err != nil {
-		return 0, err
-	}
-
-	offset, err = kafka.getOffset(broker, offset)
+	offset, err := kafka.getOffset(ctx, offset)
```
achille-roussel: I'm confused about why we need to call getOffset in this method, why not have the cursor initially be -1 (or -2), then have the goroutine fetch the offset if it needs to, then rewrite Read to do something like this:
```go
func (r *kafkaReader) Read(ctx context.Context) (msg Message, err error) {
  // 1. receive the message
  // 2. update the offset
}
```

kafka.go
```diff
@@ -194,84 +189,141 @@ func (kafka *kafkaReader) Seek(ctx context.Context, offset int64) (int64, error)
...
+	}
+}
+
+// Internal asynchronous task to fetch messages from Kafka and send to an internal
+// channel.
+func (kafka *kafkaReader) fetchMessagesAsync(ctx context.Context, eventsCh chan<- Message, errorsCh chan<- error) {
+	defer kafka.asyncWait.Done()
+
+	for {
+		offset := atomic.LoadInt64(&kafka.offset)
```
achille-roussel: Seems like we should pass the initial offset to the fetchMessagesAsync method instead of loading it from the reader.

kafka.go
```diff
@@ -194,84 +189,141 @@ func (kafka *kafkaReader) Seek(ctx context.Context, offset int64) (int64, error)
...
+	errs := make(chan error)
+	val := make(chan *sarama.Broker)
+
+	go func() {
+		broker, err := kafka.client.Leader(kafka.topic, kafka.partition)
+		if err != nil {
+			errs <- errors.Wrap(err, "failed to find leader for topic/partition")
+		} else {
+			val <- broker
+		}
```
achille-roussel: The sends on this goroutine may also block and cause the goroutine to leak if the context is canceled.

kafka.go
thehydroimpulse: `getOffset` here is used to translate the given offset `Seek(ctx, offset)` to a real kafka offset. If you have an offset like `515` then it won't actually make the call and just return the same offset. However, if the offset is `-1` or `-2` we need to ask Kafka to translate that into an actual offset to fetch. `-1` will cause Kafka to give us back the highest offset (high watermark) and `-2` will cause Kafka to give us back the lowest offset. 

However, I need to update the async task logic to move the offset updating logic to the `Read` method. We're only updating the offset after every batch, not every message. So if you do `Read()` then `Offset()` they won't match.

kafka.go
thehydroimpulse: yep, thanks for catching this!

kafka.go
thehydroimpulse: 👍

kafka.go
thehydroimpulse: if we're moving the offset increment to the `Read` method then we'll need to keep this here so that we always load the correct offset to read next.

thehydroimpulse: err, actually, we could remove the atomic offset and have a thread local offset counter for the async task that is increment after we send to the channel. Then we can have a non-synchronized offset on the reader that only the `Seek`, `Read`, and `Offset` method touch.

thehydroimpulse:
Thanks for the review. If you catch anything I'll follow up with another PR.

PR #2, opened 2017-05-23
use consul-go

thehydroimpulse:


consul.go
```diff
@@ -104,24 +90,16 @@ func NewGroupReader(config GroupConfig) (Reader, error) {
...
-		pair := &api.KVPair{
-			Key:     baseKey + reader.name + "/" + strconv.Itoa(partition),
-			Value:   []byte{},
-			Session: reader.session,
-		}
-
-		res, _, err := reader.client.KV().Acquire(pair, &api.WriteOptions{})
-		if err != nil {
-			return 0, err
+		lock, _ := consul.Lock(ctx, baseKey+reader.name+"/"+strconv.Itoa(partition))
```
thehydroimpulse: how do we specify the client or does it automatically grab the session's client from the context? @achille-roussel

achille-roussel: I believe this is documented already:
=> https://github.com/segmentio/consul-go/blob/master/README.md#sessions-and-locks
=> https://godoc.org/github.com/segmentio/consul-go#Locker
=> https://godoc.org/github.com/segmentio/consul-go#WithSession
You can check the centrifuge code as well for reference, let me know if anything is still not clear.

thehydroimpulse: thanks! the api was just a bit different than what I was expecting.

PR #4, opened 2017-05-30
Conn

achille-roussel:
Hey guys, I've spent the last couple of days working on a side project to rewrite a kafka client from scratch and here it is!
The goal I had in mind was to have an API that is much simpler to use than what sarama exposes, better documented, much less resource intensive, and that integrates well with the standard library.

Some highlights:
- the `kafka.Conn` type implements `net.Conn` (and passes the standard [nettest](https://godoc.org/github.com/golang/net/nettest) test suite)
- the `kafka.Dialer` type implements a "leader discovery" mechanism that is pretty essential to how we're consuming from kafka, it can also integrate with different service discovery backends through the use of a `kafka.Resolver`
- the `kafka.Reader` type is a back port of what we have implemented in centrifuge, which is a higher-level API to read messages from kafka (for example it supports asynchronous cancellations via contexts)

There are more optimizations to be done but here's a sexy benchmark already:
```
$ go test -v -run _ -bench 'Conn/ReadBatch$' -benchmem -memprofile /tmp/mem -cpuprofile /tmp/cpu
BenchmarkConn/ReadBatch-4         	   20000	     99455 ns/op	 100.45 MB/s	       1 B/op	       0 allocs/op
PASS
ok  	github.com/segmentio/kafka-go	3.100s
```
(yep, it pulls messages off kafka doing zero dynamic memory allocations 😛)

I gotta say, the kafka protocol specs are by far the worst I've ever dealt with (and I've done some pretty nasty shit in my days), the documentation was misleading in many cases, or sometimes just completely wrong. The protocol is also full of weird design decisions that doesn't get me too excited about the kafka-powered future we envisage... but that's another story.

Things that should be coming up:
- update of the README
- backport of the consul-based partition negotiation that Daniel had implemented (and that we use in centrifuge as well)
- optimizations of the encoding and decoding routines to get rid of as many memory allocations as possible
- maybe a friendlier API to produce messages to kafka?
- support for a wider set of kafka APIs (like consumer groups)?

Please take a look and let me know what you think!


thehydroimpulse:
Dude! This is so sick!

Amir was saying we should get you to use sarama and you'd rewrite it :)

Consumer groups internals were fairly complicated last time I tackled them in sarama but I think that was just because of sarama.

batch.go
```diff
@@ -0,0 +1,212 @@
...
+	)
+
+	batch.mutex.Unlock()
+	msg.Offset = offset
+	msg.Time = timestampToTime(timestamp)
+	return msg, err
+}
+
+func (batch *Batch) readMessage(
+	key func(r *bufio.Reader, size int, nbytes int) (int, error),
```
thehydroimpulse: what do you think of making this a proper type?

error.go
```diff
@@ -0,0 +1,353 @@
...
+
+import (
+	"fmt"
+	"io"
+)
+
+// Error represents the different error codes that may be returned by kafka.
+type Error int
+
+const (
```
thehydroimpulse: Finally no `if err != sarama.ErrNoError`!

tejasmanohar:
Holy moly! Maybe next, we'll have segmentio/kafka :trollface: 

I'm no Kafka expert, but I'll try to review this based on what I know later tomorrow. What were your main gripes with Sarama? Were they fundamental problems that makes it worth rewriting instead of contributing? It does look like a lot less code... does it cover all cases Sarama does? Is it lower level?

reader.go
```diff
@@ -1,17 +1,361 @@
...
+		}
+	}
+
+	return
+}
+
+func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, error) {
+	conn.SetReadDeadline(time.Now().Add(r.maxWait))
+	batch := conn.ReadBatch(r.minBytes, r.maxBytes)
+	conn.SetReadDeadline(time.Now().Add(10 * time.Second))
```
thehydroimpulse: wouldn't this overwrite the `maxWait` deadline?

thehydroimpulse:
@tejasmanohar this doesn't implement some of the higher-level kafka apis (yet) like sarama does (consumer groups with automatic offset management). Sarama was basically impossible to contribute to. They have 0 documentation and they abuse and overuse channels without any clear ownership (they'll just keep passing channel references everywhere). So understanding the flow of logic is pretty hard. It seems like they just kept adding more and more features.

dialer.go
```diff
@@ -0,0 +1,247 @@
...
+// partition for a given topic, returning a Broker value representing it.
+func (d *Dialer) LookupLeader(ctx context.Context, network string, address string, topic string, partition int) (Broker, error) {
+	c, err := d.DialContext(ctx, network, address)
+	if err != nil {
+		return Broker{}, err
+	}
+	defer c.Close()
+
+	brkch := make(chan Broker, 1)
+	errch := make(chan error, 1)
```
yields: should we close those chans when the method returns?

error.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// Timeout returns true if the error was due to a timeout.
+func (e Error) Timeout() bool {
+	return e == RequestTimedOut
+}
+
+// Temporary returns true if the operation that generated the error may succeed
+// if retried at a later time.
+func (e Error) Temporary() bool {
```
yields: 😍

yields:
> I gotta say, the kafka protocol specs are by far the worst I've ever dealt with (and I've done some pretty nasty shit in my days), the documentation was misleading in many cases, or sometimes just completely wrong. The protocol is also full of weird design decisions that doesn't get me too excited about the kafka-powered future we envisage... but that's another story.

same here... :/ -- kafka is full of surprises once you get to use it too.

batch.go
achille-roussel: That's a good idea.

error.go
achille-roussel: Yeah, I had NoError here initially, but then I was feeling like this was just stupid and useless, you just `!= 0` at the protocol level and `!= nil` in the program, this is so much more natural.

reader.go
achille-roussel: It does, and it's the intent. When `ReadBatch` return the response has already been received and the header were read, so the initial `maxWait` value doesn't really matter anymore, I'm just trying to set a hard deadline on how long it may take to read the messages, just in case the connection gets aborted without the client knowing about it.

dialer.go
achille-roussel: It doesn't matter since only one value will ever be sent to either of them, and we receive at most one value as well. The code doesn't need the channels to get closed and eventually the GC collects everything that's not referenced anymore.

achille-roussel:
@tejasmanohar yeah I dug a bunch into sarama, the documentation is the main pain point... The API they expose is a 1:1 representation of the internals of the kafka protocol, but the implementation does tons of async stuff in the background and has plenty of internal concepts, so you get the worst of both worlds (complex API and complex implementation).
On top of that sarama doesn't have support for contexts, and the data structures use pointers all over the place which is a crime from a memory and CPU efficiency perspective.

reader.go
thehydroimpulse: ah, good deal!

thehydroimpulse:
FYI this panics if kafka isn't running:

```
panic: runtime error: invalid memory address or nil pointer dereference
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11403d2]

goroutine 37 [running]:
github.com/segmentio/kafka-go.(*Conn).Close(0x0, 0x3, 0x11c0d89)
	/Users/danielfagnan/dev/src/github.com/segmentio/kafka-go/conn.go:131 +0x22
panic(0x118aa80, 0x1293270)
	/usr/local/Cellar/go/1.8/libexec/src/runtime/panic.go:489 +0x2cf
github.com/segmentio/kafka-go.(*Conn).writeOperation(0x0, 0xc42001c090, 0xc4200180a0, 0x0, 0xc420124000)
	/Users/danielfagnan/dev/src/github.com/segmentio/kafka-go/conn.go:613 +0x22
github.com/segmentio/kafka-go.(*Conn).WriteMessages(0x0, 0xc42012e000, 0x3e8, 0x3e8, 0x11c0d89, 0xe, 0xc4200980a0)
	/Users/danielfagnan/dev/src/github.com/segmentio/kafka-go/conn.go:554 +0x344
github.com/segmentio/kafka-go.BenchmarkConn(0xc4200e4000)
	/Users/danielfagnan/dev/src/github.com/segmentio/kafka-go/conn_test.go:487 +0x2e3
testing.(*B).runN(0xc4200e4000, 0x1)
	/usr/local/Cellar/go/1.8/libexec/src/testing/benchmark.go:140 +0xb2
testing.(*B).run1.func1(0xc4200e4000)
	/usr/local/Cellar/go/1.8/libexec/src/testing/benchmark.go:213 +0x5a
created by testing.(*B).run1
	/usr/local/Cellar/go/1.8/libexec/src/testing/benchmark.go:214 +0x7f
```

achille-roussel:
OK, I've spent a bit of time cleaning up and optimizing the implementation, I'm quite satisfies with the state the package is at right now:
```
$ go test -v -run _ -bench . -benchmem -memprofile /tmp/mem -cpuprofile /tmp/cpu
BenchmarkConn/Seek-4         	    1000	   1085967 ns/op	       0 B/op	       0 allocs/op
BenchmarkConn/Read-4         	    1000	   1148447 ns/op	   8.71 MB/s	     128 B/op	       2 allocs/op
BenchmarkConn/ReadBatch-4    	   20000	     83868 ns/op	 118.05 MB/s	       1 B/op	       0 allocs/op
BenchmarkConn/ReadOffsets-4  	    1000	   1450716 ns/op	       0 B/op	       0 allocs/op
BenchmarkConn/Write-4        	    2000	    889287 ns/op	  11.24 MB/s	       5 B/op	       0 allocs/op
PASS
ok  	github.com/segmentio/kafka-go	8.761s
```
I'm not sure what those 2 allocs on the `Read` benchmark are, I tried tracking them down but really couldn't figure it out. I suspect one of the them is the `return &Batch{ ... }` in `(*Conn).ReadBatch` but the other I have no idea (for some reason `go tool pprof` doesn't show the malloc calls.

Batck to doing real work! 😛

PR #6, opened 2017-06-01
reader improvements

achille-roussel:
I noticed kafka often fails saying the the requested offsets are out of range on the reader, I'm not sure why this is happening because it seems like the tests would never request an offset that's out of range so I suspect some sort of race condition within kafka (AKA a behavior that I may not be aware of).

This PR attempts to make the reader implementation more resilient to those cases by changing the behavior to always gracefully handle out-of-range offset errors:
- if the starting offset is before the first, the reader will start from the first offset
- if the starting offset is after the last, the reader will wait for the last offset to catch up by retrying periodically (with a backoff)

Please take a look and let me know if anything should be changed!

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -332,7 +332,7 @@ func (r *reader) initialize(ctx context.Context, offset int64) (conn *Conn, star
 			continue
 		}
 
-		// This deadline controls how long the offset negatiation may take.
+		// This deadline controls how long the offset negotiation may take.
 		conn.SetDeadline(time.Now().Add(10 * time.Second))
 
 		if first, err = conn.ReadFirstOffset(); err != nil {
--- a/reader.go
+++ b/reader.go
@@ -265,7 +265,13 @@ func (r *reader) run(ctx context.Context, offset int64) {
 		}
 
 		conn, start, err := r.initialize(ctx, offset)
-		if err != nil {
+		switch err {
+		case nil:
+		case OffsetMetadataTooLarge:
+			// This would happen if the requested offset is passed the last
+			// offset on the partition leader. In that case we're just going
+			// to retry later hoping that enough data has been produced.
+		default:
 			// Wait 4 attempts before reporting the first errors, this helps
 			// mitigate situations where the kafka server is temporarily
 			// unavailable.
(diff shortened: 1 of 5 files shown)


reader.go
```diff
@@ -284,39 +290,65 @@ func (r *reader) run(ctx context.Context, offset int64) {
...
+
+		if first, err = conn.ReadFirstOffset(); err != nil {
 			conn.Close()
-			continue
+			break
+		}
+
+		// In case the reader was configured with an offset before the first
+		// offset, skipping directly to the first offset.
+		if offset < first {
```
yields: i'm not sure about this, if the reader was misconfigured, why not return `ErrOffsetOutOfRange` or something?

reader.go
```diff
@@ -284,39 +290,65 @@ func (r *reader) run(ctx context.Context, offset int64) {
...
+			}
+
 			switch offset, err = r.read(ctx, offset, conn); err {
 			case nil:
+				errcount = 0
+				continue
 			case RequestTimedOut:
-			case context.Canceled: // another reader has taken over
+				// Timeout on the kafka side, this can be safely retried.
+			case OffsetOutOfRange:
```
yields: what happens if the reader can't keep up, i imagine the reader may lose messages without us noticing.
this could happen if the reader can't consume fast enough and some messages have been GCed by kafka (expired etc..)

reader.go
yields: also, the reader may not be able to keep up, in that case we should return an error, so that the author of the consumer can fix it.

reader.go
achille-roussel: That's a good point, although if the messages are gone they're already lost and I feel like we're better off trying to consume whatever is available. I do agree we should have a good story around being notified of those things, but getting an offset-out-of-range error doesn't communicate this well.

reader.go
achille-roussel: Hm... what does "misconfigured" means considering the set of rules proposed here?
- `offset <= first offset` means to start at the first offset
- `offset > last offset` means to wait until a message is available at that offset

yields: > offset > last offset means to wait until a message is available at that offset

this sounds awesome to me.



> offset <= first offset means to start at the first offset

what about readers that saved an offset `X` somewhere, so they manage offsets themselves
when they boot the reader with offset `X`, the reader silently adjusts the offset to the first offset in kafka server.

So for example, i saved an offset `100` in a local database, when i start the reader, the reader notices that offset is out of range and jumps to `200`, now i "lost" 100 messages without noticing?

this is my understanding.

achille-roussel: Yes, you lost 100 messages that don't exist anymore anyway. That was my comment above, we should be notified/alerted but I don't think this should prevent reading (and we should be more explicit than just offset-out-of-range).

I'll think of something!

achille-roussel:
Alright! I think I finally got those tests to pass! The code had no issue actually it was the race detector that was getting in the way (it's so expensive it just stalled everything).

I'm making a note of your suggestions Amir and I'll follow up when I have an idea for surfacing skipped offsets.

Thanks for the reviews!

PR #7, opened 2017-06-01
add high offset (watermark) support

thehydroimpulse:
cc @achille-roussel 

This propagates the high watermark offset to the batch. This is used to calculate the lag of a reader by comparing the current offset to the watermark.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -17,21 +17,27 @@ import (
 //
 // Batches are safe to use concurrently from multiple goroutines.
 type Batch struct {
-	mutex    sync.Mutex
-	conn     *Conn
-	lock     *sync.Mutex
-	reader   *bufio.Reader
-	deadline time.Time
-	throttle time.Duration
-	remain   int
-	offset   int64
-	err      error
+	mutex         sync.Mutex
+	conn          *Conn
+	lock          *sync.Mutex
+	reader        *bufio.Reader
+	deadline      time.Time
+	throttle      time.Duration
+	remain        int
+	offset        int64
+	highWaterMark int64
+	err           error
 }
 
 // Throttle gives the throttling duration applied by the kafka server on the
 // connection.
 func (batch *Batch) Throttle() time.Duration {
 	return batch.throttle
+}
+
+// Watermark returns the current highest watermark in a partition.
+func (batch *Batch) HighWaterMark() int64 {
+	return batch.highWaterMark
 }
 
 // Offset returns the offset of the next message in the batch.
--- a/conn.go
+++ b/conn.go
@@ -338,16 +338,17 @@ func (c *Conn) ReadBatch(minBytes int, maxBytes int) *Batch {
 		return &Batch{err: err}
 	}
 
-	throttle, remain, err := readFetchResponseHeader(&c.rbuf, size)
+	throttle, highWaterMark, remain, err := readFetchResponseHeader(&c.rbuf, size)
 	return &Batch{
-		conn:     c,
-		reader:   &c.rbuf,
-		deadline: adjustedDeadline,
-		throttle: duration(throttle),
-		lock:     lock,
-		remain:   remain,
-		offset:   offset,
-		err:      err,
+		conn:          c,
+		reader:        &c.rbuf,
+		deadline:      adjustedDeadline,
+		throttle:      duration(throttle),
+		lock:          lock,
+		remain:        remain,
+		offset:        offset,
+		highWaterMark: highWaterMark,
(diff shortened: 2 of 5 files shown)


conn.go
```diff
@@ -278,6 +278,7 @@ func (c *Conn) Read(b []byte) (int, error) {
 func (c *Conn) ReadMessage(maxBytes int) (Message, error) {
 	batch := c.ReadBatch(1, maxBytes)
 	msg, err := batch.ReadMessage()
+	msg.Watermark = batch.Watermark()
```
thehydroimpulse: I'm not sure about this. Maybe we want to have `Watermark` methods similar to `Offset`?

maybe something like is better:

```
for {
  msg := stream.Read()
  stat.Gauge("watermark", stream.Watermark())
}
```

yields: +1, `stream.Watermark()` seems cleaner :)

batch.go
```diff
@@ -34,6 +35,12 @@ func (batch *Batch) Throttle() time.Duration {
 	return batch.throttle
 }
 
+// Watermark returns the current highest watermark in a partition.
+func (batch *Batch) Watermark() int64 {
```
achille-roussel: Should we call this HighWatermark to be more familiar to Kafka users?

conn.go
achille-roussel: I also don't feel so good about exposing that value in the Message, it's not a property of the message. It's OK if you have to use ReadBatch to get that value, all production code should use this.

read.go
```diff
@@ -273,6 +273,8 @@ func readFetchResponseHeader(r *bufio.Reader, size int) (throttle int32, remain
 		return
 	}
 
+	watermark = p.HighwaterMarkOffset
+
```
achille-roussel: Unnecessary empty line here.

read.go
```diff
@@ -215,7 +215,7 @@ func readSlice(r *bufio.Reader, sz int, v reflect.Value) (int, error) {
 	return sz, nil
 }
 
-func readFetchResponseHeader(r *bufio.Reader, size int) (throttle int32, remain int, err error) {
+func readFetchResponseHeader(r *bufio.Reader, size int) (watermark int64, throttle int32, remain int, err error) {
```
achille-roussel: Should the watermark come after the throttle value since it's given after in the response? Seems more natural to preserve the order.

message.go
```diff
@@ -7,10 +7,11 @@ import (
 
 // Message is a data structure representing kafka messages.
 type Message struct {
-	Offset int64
-	Key    []byte
-	Value  []byte
-	Time   time.Time
+	Offset    int64
+	Watermark int64
```
achille-roussel: This should be removed.

reader.go
```diff
@@ -168,6 +175,10 @@ func (r *Reader) ReadMessage(ctx context.Context) (msg Message, err error) {
 			err = ctx.Err()
 			return
 
+		case watermark := <-r.highWatermarks:
```
achille-roussel: I'm not a big fan of this, it seems like we could pass it as part of the internal `readerMessage` data structure.

reader.go
```diff
@@ -203,6 +214,13 @@ func (r *Reader) Offset() int64 {
 	return offset
 }
 
+func (r *Reader) HighWatermark() int64 {
```
achille-roussel: Instead of exposing the concept of a HighWatermark here, should we replace this with `Lag` which exposes the difference between the offset of the last read message and the high watermark? It seems like this is the concept that the program is gonna be interested in.

thehydroimpulse: I think there are other cases where knowing the high watermark is important but ultimately we do just want the lag. 

Lets settle with a lag method and we can always add back a lower-level watermark later.

batch.go
```diff
@@ -34,6 +35,12 @@ func (batch *Batch) Throttle() time.Duration {
 	return batch.throttle
 }
 
+// Watermark returns the current highest watermark in a partition.
+func (batch *Batch) HighWatermark() int64 {
+	watermark := batch.highWatermark
+	return watermark
```
achille-roussel: Any reason not to just `return batch.highWatermark` here?

thehydroimpulse:
Sarama names it `HighWaterMark` which fits Go's convention better than `HighwaterMark` (which just seems wrong).

achille-roussel:
Alright, I'm down for _HighWaterMark_ if that's what's been used historically 👍

thehydroimpulse:
thanks for the review!

PR #9, opened 2017-06-21
Add the Writer type

achille-roussel:
Hey Kafka boyz, Here's the implementation of a high-level Writer API to produce messages to kafka, a couple of highlights:

- automatically discovers the available partitions for a topic
- balancing messages across partitions using the `Balancer` abstraction
- supports automatic retries on failures
- supports sync and async modes
- supports async cancellation via contexts
- gracefully closes, waits for inflight messages to be flushed before returning

I still have to add more tests for the message balancing algorithms and a section to the README but I think the code is mature enough to be reviewed now.

Please take a look and let me know if anything should be changed.

### DIFF
--- a/balancer.go
+++ b/balancer.go
@@ -1,21 +1,37 @@
 package kafka
 
-import "sync/atomic"
-
+// The Balancer interface provides an abstraction of the message distribution
+// logic used by Writer instances to route messages to the partitions available
+// on a kafka cluster.
+//
+// Instances of Balancer do not have to be safe to use concurrently by multiple
+// goroutines, the Writer implementation ensures that calls to Balance are
+// synchronized.
 type Balancer interface {
-	Balance(key []byte, partitions ...int) (partition int)
+	// Balance receives a message and a set of available partitions and
+	// returns the partition number that the message should be routed to.
+	Balance(msg Message, partitions ...int) (partition int)
 }
 
-type BalancerFunc func([]byte, ...int) int
+// BalancerFunc is an implementation of the Balancer interface that makes it
+// possible to use regular functions to distribute messages across paritions.
+type BalancerFunc func(Message, ...int) int
 
-func (f BalancerFunc) Balance(key []byte, partitions ...int) int {
-	return f(key, partitions...)
+// Balance calls f, satisfies the Balancer interface.
+func (f BalancerFunc) Balance(msg Message, partitions ...int) int {
+	return f(msg, partitions...)
 }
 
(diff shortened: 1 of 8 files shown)


balancer.go
```diff
@@ -0,0 +1,106 @@
...
+	return partitions[offset%length]
+}
+
+// LeastBytes is a Balancer implementation that routes messages to the partition
+// that has received the least amount of data.
+//
+// Note that no coordination is done between multiple producers, having good
+// balancing relies on the fact that each producer using a LeastBytes balancer
+// should produce well balanced messages.
+type LeastBytes struct {
```
yields: 💥

writer.go
```diff
@@ -0,0 +1,533 @@
...
+	"sync"
+	"time"
+)
+
+// The Writer type provides the implementation of a producer of kafka messages
+// that automatically distributes messages across partitions of a single topic
+// using a configurable blancing policy.
+//
+// Instances of Writer are safe to use concurrently from multiple goroutines.
+type Writer struct {
```
yields: it'd be really nice to have `Example` tests for this type :)

writer.go
```diff
@@ -0,0 +1,533 @@
...
+	mutex  sync.RWMutex
+	closed bool
+
+	join sync.WaitGroup
+	msgs chan writerMessage
+	done chan struct{}
+}
+
+// WriterConfig is a configuration type used to create new instances of Writer.
+type WriterConfig struct {
```
yields: is there no way to set required_acks? or is it set on the msg?

writer.go
```diff
@@ -0,0 +1,533 @@
...
+	// This interval defines how often the list of partitions is refreshed from
+	// kafka. It allows the writer to automatically handle when new partitions
+	// are added to a topic.
+	//
+	// The default is to refresh partitions every 15 seconds.
+	RebalanceInterval time.Duration
+
+	// Setting this flag to true causes the WriteMessages method to never block.
+	// It also means that errors are ignored since the caller will not receive
+	// the returned value. Use this only if you don't care about garantees of
```
yields: typo (`garantees`)

achille-roussel:
`required_acks=all` is the default already, and we can't configure it yet => https://github.com/segmentio/kafka-go/blob/master/write.go#L202

It seems like it would be good to make it configurable and not require acks when the writer is used in async mode, that's work for a follow up PR I believe.

achille-roussel:
Thanks for the review Amir!

PR #10, opened 2017-06-22
reduce the number of channels involved in reading messages

achille-roussel:
This pull request addresses a performance issue that was mainly due to the use of unbuffered channels to pass messages between the internal goroutine of the Reader and the caller of ReadMessage.

Here are profiles before and after the change:
```
$ go tool pprof ./kafka-go.test /tmp/base1.cpu
Entering interactive mode (type "help" for commands)
(pprof) top10
3440ms of 4620ms total (74.46%)
Dropped 31 nodes (cum <= 23.10ms)
Showing top 10 nodes out of 88 (cum >= 650ms)
      flat  flat%   sum%        cum   cum%
    1240ms 26.84% 26.84%     2800ms 60.61%  runtime.selectgoImpl
     410ms  8.87% 35.71%      410ms  8.87%  runtime.unlock
     360ms  7.79% 43.51%      500ms 10.82%  runtime.lock
     260ms  5.63% 49.13%      280ms  6.06%  syscall.(*Errno).Timeout
     250ms  5.41% 54.55%      250ms  5.41%  runtime.mach_msg_trap
     210ms  4.55% 59.09%      560ms 12.12%  runtime.selectrecv
     210ms  4.55% 63.64%      210ms  4.55%  runtime.selectrecvImpl
     200ms  4.33% 67.97%      200ms  4.33%  runtime.munmap
     160ms  3.46% 71.43%      160ms  3.46%  runtime.selectrecv2
     140ms  3.03% 74.46%      650ms 14.07%  github.com/segmentio/kafka-go.(*Reader).ReadMessage
```
```
$ go tool pprof ./kafka-go.test /tmp/base2.cpu
Entering interactive mode (type "help" for commands)
(pprof) top10
490ms of 490ms total (  100%)
Showing top 10 nodes out of 105 (cum >= 30ms)
      flat  flat%   sum%        cum   cum%
     170ms 34.69% 34.69%      180ms 36.73%  syscall.Syscall
      80ms 16.33% 51.02%       80ms 16.33%  runtime.mach_semaphore_timedwait
      70ms 14.29% 65.31%       70ms 14.29%  runtime.usleep
      50ms 10.20% 75.51%       50ms 10.20%  runtime.kevent
      50ms 10.20% 85.71%       50ms 10.20%  runtime.mach_semaphore_signal
      30ms  6.12% 91.84%       30ms  6.12%  runtime.mach_semaphore_wait
      20ms  4.08% 95.92%       20ms  4.08%  runtime.freedefer
      10ms  2.04% 97.96%       10ms  2.04%  runtime.memmove
      10ms  2.04%   100%       10ms  2.04%  runtime.readvarint
         0     0%   100%       30ms  6.12%  bufio.(*Reader).Peek
```
Most of the time used to be spent in the `select` statement of ReadMessage, now it's spent in the `read` syscall, which is what we'd expect for an operation that's supposed to be I/O bound (reading from Kafka).

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -90,7 +90,7 @@ func NewReader(config ReaderConfig) *Reader {
 	}
 
 	if config.MaxBytes == 0 {
-		config.MaxBytes = 10e6 // 10 MB
+		config.MaxBytes = 1e6 // 1 MB
 	}
 
 	if config.MinBytes == 0 {
--- a/reader.go
+++ b/reader.go
@@ -19,17 +19,14 @@ type Reader struct {
 
 	// communication channels between the parent reader and its subreaders
 	msgs chan readerMessage
-	errs chan readerError
-	done chan struct{}
 
 	// mutable fields of the reader (synchronized on the mutex)
 	mutex   sync.Mutex
-	ctx     context.Context
+	join    sync.WaitGroup
 	cancel  context.CancelFunc
 	version int64
 	offset  int64
 	lag     int64
-	dirty   bool
 	closed  bool
 }
 
@@ -48,6 +45,10 @@ type ReaderConfig struct {
 	// An dialer used to open connections to the kafka server. This field is
 	// optional, if nil, the default dialer is used instead.
 	Dialer *Dialer
+
+	// The capacity of the internal message queue, defaults to 100 if none is
+	// set.
+	QueueCapacity int
 
 	// Min and max number of bytes to fetch from kafka in each request.
 	MinBytes int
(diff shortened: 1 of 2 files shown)


reader.go
```diff
@@ -122,76 +122,56 @@ func (r *Reader) Config() ReaderConfig {
...
-		close(r.done)
-	}
+	closed := r.closed
 	r.closed = true
-	r.dirty = false
-	r.cancel()
 	r.mutex.Unlock()
+
+	r.cancel()
+	r.join.Wait()
```
yields: we didn't wait before? be nice to have a test showing that we wait for internal goroutines to exit first.

achille-roussel: Yep, but I don't think it's really feasable. `sync.WaitGroup` doesn't expose the value of its internal counter, that would require a bunch of changes to the code to verify it... doesn't seem worth it.

abraithwaite: When I want to write a test like this I generally wrap the WaitGroup in an interface with a local private struct in which you increment a counter and the waitgroup at the same time.  There are ways to do it which are clean.

achille-roussel:
Thanks for the review!

reader.go
achille-roussel: I agree, I didn't say it wouldn't be clean, I feel it increases complexity for no clear benefits. There's always something more we can test but at some point we gain very little value out of it. Here the synchronization is pretty trivial, there's only one spot where goroutines are being started, I think it's not worth adding internal abstractions.

PR #12, opened 2017-06-27
add Reader.ReadLag

achille-roussel:
The intent here is to be able to estimate the lag of a reader without having to read messages. There are times where one of our kafka consumers was failing to make progress and therefore wasn't calling ReadMessage anymore, which means the high watermark wasn't getting updated and the lag value became inaccurate since it was based on outdated information.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -176,6 +176,59 @@ func (r *Reader) ReadMessage(ctx context.Context) (Message, error) {
 	}
 }
 
+// ReadLag returns the current lag of the reader by fetching the last offset of
+// the topic and partition and computing the difference between that value and
+// the offset of the last message returned by ReadMessage.
+//
+// This method is intended to be used in cases where a program may be unable to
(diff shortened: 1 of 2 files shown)


reader.go
```diff
@@ -176,6 +176,59 @@ func (r *Reader) ReadMessage(ctx context.Context) (Message, error) {
 	}
 }
 
+// ReadLag returns the current lag of the reader by fetching the last offset of
+// the topic and partition and computing the difference between that value and
+// the offset of the last message returned by ReadMessage.
+//
+// This method is intended to be used in cases where a program may be unable to
+// call ReadMessage to update the value returned by Lag, but still needs to get
```
abraithwaite: For example, when the consumer is not ready to process the next message, correct?

achille-roussel: Yes

achille-roussel:
Thanks for the review guys!

PR #13, opened 2017-06-27
don't report lag until at least one message is read

achille-roussel:
We've run into a weird situation where when the consumer hasn't read any messages it believes that it has the entire all-time backlog of lag. This PR addresses this issue by reporting no lag until at least one message has been read to set the offset to a positive value.


abraithwaite:
Haha, we had this exact issue!  Can't believe I didn't catch that.

achille-roussel:
Thanks for the review!

PR #14, opened 2017-07-17
Logging + Reader bug fix

achille-roussel:
This PR adds the ability to set a logger on a kafka reader to get insights into the internals, while also probably fixing a bug with offset management.
I haven't been able to really identify the root cause of the issue but this code has been running in production for a week now and the bug hasn't triggered (where it did consistently before).

The problem we were seeing was the reader being unable to make progress, its offset was always being reset to a previous value.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"math"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


achille-roussel:
Thanks for the review!

abraithwaite:
The changes look good.  i can't see any obvious issues which would cause the offset to be reset here.

In an internal library I've used in the past, we were accidentally committing the offset on Close of the consumer, but we're clearly not doing that here.

PR #17, opened 2017-08-28
add higher level interface to lookup partitions of a topic

achille-roussel:
This was available on the connection object but it's a low level interface, with this change it becomes a context-aware single method call to lookup partitions of a topic.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -169,6 +169,35 @@ func (d *Dialer) LookupLeader(ctx context.Context, network string, address strin
 	return brk, err
 }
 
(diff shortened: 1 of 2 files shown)


dialer.go
```diff
@@ -169,6 +169,35 @@ func (d *Dialer) LookupLeader(ctx context.Context, network string, address strin
 	return brk, err
 }
 
+// LookupPartitions returns the list of partitions that exist for the given topic.
+func (d *Dialer) LookupPartitions(ctx context.Context, network string, address string, topic string) ([]Partition, error) {
```
abraithwaite: 🥇

PR #19, opened 2017-09-26
Fix NotLeaderForPartition error

thehydroimpulse:
This adds some logic to re-establish a connection with the new leader if the current one is no longer valid.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -507,7 +507,24 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
 	var err error
 
 	for {
-		if msg, err = batch.ReadMessage(); err != nil {
+		if msg, err = batch.ReadMessage(); err != nil && err == NotLeaderForPartition {
+			if err = batch.Close(); err != nil {
+				break
+			}
+
+			// Close the connection ignoring any errors it may cause.
+			conn.Close()
+
+			// The current connection is invalid for the selected partition likely caused
+			// by leader election. Try and re-initialize the connection
+			conn, _, err = r.initialize(ctx, offset)
+			if err != nil {
+				break
+			}
+
+			// read will re-initialize the batch with the new connection.
+			return r.read(ctx, offset, conn)
+		} else if err != nil {
 			err = batch.Close()
 			break
 		}
--- a/reader.go
+++ b/reader.go
@@ -424,7 +424,19 @@ func (r *reader) run(ctx context.Context, offset int64, join *sync.WaitGroup) {
 			switch offset, err = r.read(ctx, offset, conn); err {
 			case nil:
 				errcount = 0
+			case NotLeaderForPartition:
+				if err = conn.Close(); err != nil {
+					r.withLogger(func(log *log.Logger) {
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -507,7 +507,24 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
 	var err error
 
 	for {
-		if msg, err = batch.ReadMessage(); err != nil {
+		if msg, err = batch.ReadMessage(); err != nil && err == NotLeaderForPartition {
```
yields: nit: is `err != nil` needed here?

reader.go
```diff
@@ -507,7 +507,24 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
...
+			if err = batch.Close(); err != nil {
```
yields: do we/should we wrap errors?

reader.go
thehydroimpulse: yep, I'll remove it!

achille-roussel:
I'm not sure how this will address the issue, it seems to me that this error would previously go through the `default` case which closes the connection and goes back to `initialize`. The special case here does the same.
I don't think we have a good sense of why the consumer started failing to find the right leader for the partition it was consuming from, getting to the bottom of that will help us understand how to address the issue.

achille-roussel:
Oh actually now that I re-read the code this is the problem: https://github.com/segmentio/kafka-go/blob/d3acaa5f3474c8447b8d6f16ec510e78656d9bb2/reader.go#L458

we keep looping here when we get this error.

OK that makes sense now.

reader.go
```diff
@@ -424,7 +424,18 @@ func (r *reader) run(ctx context.Context, offset int64, join *sync.WaitGroup) {
...
+					r.withLogger(func(log *log.Logger) {
+						log.Printf("failed to close connection for partition %d of %s at offset %d", r.partition, r.topic, offset)
+					})
+				}
 
+				conn, start, err = r.initialize(ctx, offset)
+				if err != nil {
+					r.sendError(ctx, err)
+					break readLoop
+				}
```
achille-roussel: I don't think we should bubble up this error by calling `sendError`, the reader should abstract this problem away and pretend that everything is fine.

How about just doing this?
- log the error
- close the connection
- `break readLoop`

The code will flow back normally to the initialization phase.

reader.go
```diff
@@ -424,7 +424,18 @@ func (r *reader) run(ctx context.Context, offset int64, join *sync.WaitGroup) {
 			switch offset, err = r.read(ctx, offset, conn); err {
 			case nil:
 				errcount = 0
+			case NotLeaderForPartition:
+				if err = conn.Close(); err != nil {
+					r.withLogger(func(log *log.Logger) {
```
achille-roussel: This is overkill, closing a network connection won't fail.

thehydroimpulse:
thanks for the review! Updated to remove the redundant code and re-use the initialization that's already there.

calvinfo:
Boom thanks guys!

PR #21, opened 2017-09-29
add reader stats

achille-roussel:
This is a long due pull request. It adds metrics collection to the internal implementation of the kafka reader, and exposes them to the application through a new `kafka.(*Reader).Stats` method.

The `kafka.ReaderStats` structure is designed to work with [stats](https://github.com/segmentio/stats) and can be directly passed to the stats engine's `Report` method.

I've tried to add as many metrics as I could think of but if you notice anything that is missing please let me know.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log"
 	"math"
+	"strconv"
 	"sync"
 	"time"
 )
@@ -30,10 +31,14 @@ type Reader struct {
 	mutex   sync.Mutex
 	join    sync.WaitGroup
 	cancel  context.CancelFunc
+	stop    context.CancelFunc
 	version int64
 	offset  int64
 	lag     int64
 	closed  bool
+
+	// reader stats are all made of atomic values, no need for synchronization.
+	stats readerStats
 }
 
 // ReaderConfig is a configuration object used to create new instances of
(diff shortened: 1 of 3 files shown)


reader.go
```diff
@@ -64,11 +69,59 @@ type ReaderConfig struct {
...
+	Bytes      int64 `metric:"kafka.reader.message.bytes"   type:"counter"`
+	Rebalances int64 `metric:"kafka.reader.rebalance.count" type:"counter"`
+	Timeouts   int64 `metric:"kafka.reader.timeout.count"   type:"counter"`
+	Errors     int64 `metric:"kafka.reader.error.count"     type:"counter"`
+
+	DialTime DurationStats `metric:"kafka.reader.dial.seconds"`
+	ReadTime DurationStats `metric:"kafka.reader.read.seconds"`
+	WaitTime DurationStats `metric:"kafka.reader.wait.seconds"`
+
+	Offset        int64 `metric:"kafka.reader.offset"          type:"gauge"`
```
abraithwaite: Should offset be a gauge or counter?

achille-roussel: Gauge seems better because the reader exposes method to change the offset that it's reading from, so it may go backward sometimes.

achille-roussel:
Thanks for the review!

reader.go
abraithwaite: I see, makes sense.  I wish Datadog were better at graphing gauge types then :-(.

PR #22, opened 2017-09-29
fix stats

achille-roussel:
Follow up of #21, fixes broken average when the count of observed values is zero, add summaries for number of messages and sizes of each fetch to kafka.


achille-roussel:
Thanks for the review!

PR #23, opened 2017-10-03
fix error handling on writes and fix invalid partition condition

thehydroimpulse:


writer.go
```diff
@@ -306,8 +313,21 @@ func (w *Writer) run() {
 			}
 
 			if len(partitions) != 0 {
-				writers[w.config.Balancer.Balance(wm.msg, partitions...)].msgs <- wm
+				selectedPartition := w.config.Balancer.Balance(wm.msg, partitions...)
```
thehydroimpulse: This is used to allow us to test the error case.

calvinfo:
Though the circleCI log isn't loading for me, why do the tests fail?

thehydroimpulse:
Not sure but the way I was testing it was a little wacky. Re-working the tests now.

thehydroimpulse:
@calvinfo  mind looking it over once more after some changes?

writer.go
```diff
@@ -27,6 +28,7 @@ type Writer struct {
 
 // WriterConfig is a configuration type used to create new instances of Writer.
 type WriterConfig struct {
+	NewPartitionWriter func(partition int, config WriterConfig) PartitionWriter
```
achille-roussel: Is the reason we need this only to support mocking during tests? If that's the case I'd advocate for making it an unexported field so we don't expose implementation details to the public API.

writer.go
```diff
@@ -231,7 +240,11 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 		timer := time.NewTimer(backoff(attempt+1, 100*time.Millisecond, 1*time.Second))
 		select {
 		case <-timer.C:
-			err = nil
+			// Only clear the error (so we retry the loop) if we have more retries, otherwise
+			// we risk silencing the error.
+			if attempt != w.config.MaxAttempts-1 {
```
achille-roussel: How about replacing this condition by `attemp < w.config.MaxAttempts`? It is easier to read and more robust (if for any reason attempt happened to be greater than MaxAttempts the current code would break).

writer.go
```diff
@@ -364,7 +384,12 @@ func diffp(new []int, old []int) (diff []int) {
 	return
 }
 
-type writer struct {
+type PartitionWriter interface {
```
achille-roussel: Same comment as for NewPartitionWriter in the configuration object, if all we're solving with this abstraction is to mock tests let's not make it part of the public API.

thehydroimpulse:
Updated to address the feedback!

writer.go
```diff
@@ -364,6 +384,11 @@ func diffp(new []int, old []int) (diff []int) {
 	return
 }
 
+type partitionWriter interface {
+	Messages() chan<- writerMessage
+	Close()
```
achille-roussel: There is no need to export those methods, they are only used internally within the package. To make the intention clear I would lowercase them.

writer.go
```diff
@@ -27,6 +28,7 @@ type Writer struct {
 
 // WriterConfig is a configuration type used to create new instances of Writer.
 type WriterConfig struct {
+	newPartitionWriter func(partition int, config WriterConfig) partitionWriter
```
achille-roussel: It's considered good practice in Go to put unexported symbols after exported ones, in order to first present to the reader the API that it can use.

writer.go
```diff
@@ -240,6 +253,7 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 		timer.Stop()
 
 		if err != nil {
+			err = fmt.Errorf("failed to write message: %s", err)
```
achille-roussel: What is the purpose of overriding this error? It seems like it would discard important type information about the problem (for example if the original error is a `*net.OpError` or `context.DeadlineExceeded` it coudl be introspected by the program).

writer.go
thehydroimpulse: 👍 I'll open a new PR to fix these.

writer.go
thehydroimpulse: good point!

PR #26, opened 2017-10-06
add error logger

achille-roussel:
This PR adds the ability to configure a logger specifically to report errors on kafka readers. This is useful to avoid the flood of common logs under normal situation and only surface the exceptional events that occur within the readers.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -79,6 +79,10 @@ type ReaderConfig struct {
 	// If not nil, specifies a logger used to report internal changes within the
 	// reader.
 	Logger *log.Logger
+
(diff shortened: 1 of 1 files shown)


f2prateek:
> This is useful to avoid the flood of common logs under normal situation 

Shouldn't those normal logs be "debug"? Can you highlight an example of a log for which you'd want to use an error logger, and one for which you wouldn't?

f2prateek:
Ah, this isn't using `events.Logger` so we don't have levels here.

achille-roussel:
Yeah, for better compatibility this package uses the standard library only.

achille-roussel:
Thanks for the review!

PR #32, opened 2017-11-09
EOF on closed reader

achille-roussel:
This PR changes the error returned by ReadMessage to EOF when the reader has been closed, there are two reasons why we make this change:

- io.EOF is the _correct_ error for indicating that a read operation has reached the end of the data stream it was reading
- io.ErrClosedPipe maybe returned by the inner reader if the connection is lost, and would confuse the application into thinking that the reader has reached the end

This may be considered a breaking change, however it was not previously documented and doesn't break the compilation either.

Please take a look and let me know if you have any concerns.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -254,7 +254,7 @@ func (r *Reader) ReadMessage(ctx context.Context) (Message, error) {
 
 		case m, ok := <-r.msgs:
 			if !ok {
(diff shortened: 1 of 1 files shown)


achille-roussel:
Thanks for the reviews!

PR #34, opened 2017-11-13
fix reading before the first offset

achille-roussel:
We noticed an issue where consumers would stop making progress if they were falling behind and attempted to read from an offset that was before the first offset of a partition.

This PR addresses the issue.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -626,7 +626,7 @@ func (r *reader) run(ctx context.Context, offset int64, join *sync.WaitGroup) {
 
 				if err != nil {
 					r.withErrorLogger(func(log *log.Logger) {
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -618,15 +620,42 @@ func (r *reader) run(ctx context.Context, offset int64, join *sync.WaitGroup) {
...
+					break readLoop
+				}
+
+				switch {
+				case offset < first:
+					r.withErrorLogger(func(log *log.Logger) {
+						log.Printf("the kafka reader is reading before the first offset for partition %d of %s, skipping from offset %d to %d (%d messages)", r.partition, r.topic, offset, first, first-offset)
+					})
+					offset, errcount = first, 0
+					continue // retry immediately so we don't keep falling behind due to the backoff
```
abraithwaite: Nice.  Ran into this quite frequently in the past.  I like this solution.

PR #36, opened 2017-11-21
readLag call in SetOffset

maxence-charriere:
This PR make the lag reading start also when a call to SetOffset is performed.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -451,7 +451,7 @@ func (r *Reader) withErrorLogger(do func(*log.Logger)) {
 	}
 }
 
-func (*Reader) activateReadLag() {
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -384,6 +384,11 @@ func (r *Reader) Lag() int64 {
 // The method fails with io.ErrClosedPipe if the reader has already been closed.
 func (r *Reader) SetOffset(offset int64) error {
 	var err error
+
+	if r.config.ReadLagInterval > 0 && atomic.CompareAndSwapUint32(&r.once, 0, 1) {
```
achille-roussel: Let’s put this operation in its own menthod now that it’s used in two places.

PR #39, opened 2017-12-20
Goroutine locking on writer.go:WriteMessages method

josledp:
writer.go:write method return nothing on resch channels when conn==nil and w.dial() fails, so the WriteMessages method locks waiting forever some answer in the res channel


josledp:
Hi, 

  I've been trying to add the test, but the problem is that when calling Writer.partitions() it connects to the broker the get the partitions information. If it cannot connect it does not open the partition, which is the place when the writer is created. So I do not see any way to add the test without modifying the writer.go code.

achille-roussel:
Alright, the change looks good to me, it was definitely a miss. 

Thanks for the fix 👍

PR #41, opened 2017-12-23
added TLS support to Dialer

savaki:
Added

`TLS *tls.Config`

to Dialer.  When present, Dialer will attempt to connect via TLS.  Otherwise,  standard net.Conn will be used.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -203,6 +203,29 @@ func (d *Dialer) LookupPartitions(ctx context.Context, network string, address s
 	return prt, err
 }
 
+// connectTLS returns a tls.Conn that has already completed the Handshake
+func (d *Dialer) connectTLS(ctx context.Context, conn net.Conn) (tlsConn *tls.Conn, err error) {
+	tlsConn = tls.Client(conn, d.TLS)
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+		err = tlsConn.Handshake()
+	}()
+
+	select {
+	case <-ctx.Done():
+		conn.Close()
+		tlsConn.Close()
+		err = ctx.Err()
+		return
+
+	case <-done:
+	}
+
+	return
+}
+
 func (d *Dialer) dialContext(ctx context.Context, network string, address string) (net.Conn, error) {
 	if r := d.Resolver; r != nil {
 		host, port := splitHostPort(address)
@@ -230,7 +253,7 @@ func (d *Dialer) dialContext(ctx context.Context, network string, address string
 	}
 
 	if d.TLS != nil {
-		return tls.Client(conn, d.TLS), nil
+		return d.connectTLS(ctx, conn)
 	}
 
 	return conn, nil
--- a/dialer.go
+++ b/dialer.go
@@ -2,6 +2,7 @@ package kafka
 
 import (
 	"context"
+	"crypto/tls"
 	"net"
 	"strconv"
 	"time"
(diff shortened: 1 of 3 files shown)


dialer.go
```diff
@@ -213,12 +218,19 @@ func (d *Dialer) dialContext(ctx context.Context, network string, address string
...
+	dialer := &net.Dialer{
 		LocalAddr:     d.LocalAddr,
 		DualStack:     d.DualStack,
 		FallbackDelay: d.FallbackDelay,
 		KeepAlive:     d.KeepAlive,
-	}).DialContext(ctx, network, address)
+	}
+
+	if d.TLS != nil {
+		return tls.DialWithDialer(dialer, network, address, d.TLS)
```
achille-roussel: It’s sad that we have to choose between supporting context cancellations and TLS support. 

Do you think we could keep context cancellation by doing a basic dialer.DialContext, then call tls.Client, and tlsConn.Handshake in a separate goroutine so the main one can select in the context cancellation and completion of the handshake.

savaki: Sure.  About to head out to the movies with family.  I'll make that change when I get back.

achille-roussel: No hurry :+1:

savaki:
now with context support

dialer.go
```diff
@@ -213,12 +218,22 @@ func (d *Dialer) dialContext(ctx context.Context, network string, address string
...
 		DualStack:     d.DualStack,
 		FallbackDelay: d.FallbackDelay,
 		KeepAlive:     d.KeepAlive,
 	}).DialContext(ctx, network, address)
+	if err != nil {
+		return nil, err
+	}
+
+	if d.TLS != nil {
+		return tls.Client(conn, d.TLS), nil
```
achille-roussel: Doing this delays the TLS handshake until a message is first read or written, which may have unexpected consequences (like timeouts if the server expects the handshake to come quickly, but the first message may be produced too late).

This is what I suggested to do the handshake in a separate goroutine, and select on the context and the completion of that goroutine to support proper cancellation during this step.

savaki: Ah, gotcha.  Didn't realize this delayed the handshake.  Let me take another stab at it.  

On a separate note, I took at crack at consumer groups yesterday.  It's far from finished, but it's probably far enough along to get your feedback on it.  https://github.com/savaki/kafka-go/tree/consumer-groups

Currently, it handles in a very basic form: find coordinator, join group, and sync group

dialer.go
```diff
@@ -198,6 +203,29 @@ func (d *Dialer) LookupPartitions(ctx context.Context, network string, address s
...
 }
 
+// connectTLS returns a tls.Conn that has already completed the Handshake
+func (d *Dialer) connectTLS(ctx context.Context, conn net.Conn) (tlsConn *tls.Conn, err error) {
+	tlsConn = tls.Client(conn, d.TLS)
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+		err = tlsConn.Handshake()
```
achille-roussel: There's a race condition between `err = ctx.Err()` and `err = tlsConn.Handshake()` here.

One way to solve this would be to change the `done` channel to be of type `chan error` and publish the error returned by `tlsConn.Handshake()`. Then the select statement looks like this:

```go
select {
case <-ctx.Done():
    ...
case err = <-done:
}
```
That way there's only one goroutine writing to the `err` variable.

dialer.go
```diff
@@ -198,6 +203,29 @@ func (d *Dialer) LookupPartitions(ctx context.Context, network string, address s
 	return prt, err
 }
 
+// connectTLS returns a tls.Conn that has already completed the Handshake
+func (d *Dialer) connectTLS(ctx context.Context, conn net.Conn) (tlsConn *tls.Conn, err error) {
+	tlsConn = tls.Client(conn, d.TLS)
+	errch := make(chan error)
+
+	go func() {
+		defer close(errch)
```
achille-roussel: Closing the channel is unnecessary since a value is always produced to the channel, there's no risk to block waiting for a value. It has no impact on the correctness of the code tho.

PR #46, opened 2017-12-27
Add topic partition to message

savaki:
kafka.Message now includes topic and partition

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -144,7 +144,10 @@ func (batch *Batch) Read(b []byte) (int, error) {
 // it is less memory-efficient than Read, but has the advantage of never
 // failing with io.ErrShortBuffer.
 func (batch *Batch) ReadMessage() (Message, error) {
-	msg := Message{}
+	msg := Message{
+		Topic:     batch.conn.topic,
+		Partition: batch.conn.partition,
+	}
 	batch.mutex.Lock()
 
 	offset, timestamp, err := batch.readMessage(
--- a/batch.go
+++ b/batch.go
@@ -144,10 +144,7 @@ func (batch *Batch) Read(b []byte) (int, error) {
 // it is less memory-efficient than Read, but has the advantage of never
 // failing with io.ErrShortBuffer.
 func (batch *Batch) ReadMessage() (Message, error) {
-	msg := Message{
-		Topic:     batch.conn.topic,
-		Partition: batch.conn.partition,
-	}
+	msg := Message{}
 	batch.mutex.Lock()
 
 	offset, timestamp, err := batch.readMessage(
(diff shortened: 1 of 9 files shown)


batch.go
```diff
@@ -144,7 +144,10 @@ func (batch *Batch) Read(b []byte) (int, error) {
 // it is less memory-efficient than Read, but has the advantage of never
 // failing with io.ErrShortBuffer.
 func (batch *Batch) ReadMessage() (Message, error) {
-	msg := Message{}
+	msg := Message{
+		Topic:     batch.conn.topic,
+		Partition: batch.conn.partition,
+	}
```
achille-roussel: Can we set this only on success? Otherwise when an error is returned it’s not the zero-value if the message which is being returned anymore.

conn.go
```diff
@@ -13,6 +14,11 @@ import (
 	"time"
 )
 
+var (
+	errInvalidWriteTopic     = errors.New("writes should NOT set Topic on kafka.Message")
+	errInvalidWritePartition = errors.New("writes should NOT set Partition on kafka.Message")
```
achille-roussel: How about rephrasing to say “cannot” or “must not” so it’s more assertive? “should” is a conditional which communicates that it may be OK under some conditions.

conn.go
```diff
@@ -507,6 +513,17 @@ func (c *Conn) WriteMessages(msgs ...Message) (int, error) {
 		return 0, nil
 	}
 
+	// users may believe they can set the Topic and/or Partition
+	// on the kafka message.
```
achille-roussel: Let’s add this to te WriteMessages documentation so we leave no room for interpretation here.

createtopics_test.go
```diff
@@ -0,0 +1,241 @@
+package kafka
```
achille-roussel: Is this file expected to be here? It seems weird to have those structures in a test file, and it’s unclear how it’s related to the kafka.Message changes.

message.go
```diff
@@ -7,10 +7,12 @@ import (
 
 // Message is a data structure representing kafka messages.
 type Message struct {
-	Offset int64
-	Key    []byte
-	Value  []byte
-	Time   time.Time
+	Topic     string
+	Partition int32
```
achille-roussel: The package uses int for partition numbers everwhere, let’s keep it consistent here.

createtopics_test.go
savaki: yes.  The existing test cases read and publishes messages to partition 0.  Consequently, there seemed to be no way to test whether or not partition was being set correctly or if it was a byproduct of partition 0 sharing go's zero value.

My solution was to implement createTopics in the test package only so I could ensure I was working on a topic with more than 1 partition.

message.go
savaki: Sure.  I should have looked further.  I saw Conn declared partition int32 and followed.

createtopics_test.go
savaki: And to clarify, I left them in test because no production code had need for createTopics.  I figured they can always be pulled in if the need arises.

savaki:
That should be everything.   I appreciate all the feedback over the holidays.

PR #47, opened 2017-12-27
expanding Conn to handle kafka api calls required to support consumer groups

savaki:
added raw Kafka calls to conn.  rebalance will be added to Reader.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -136,19 +136,6 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 	return c
 }
 
-func writeHeader(w *bufio.Writer, clientID string, apiKey apiKey, apiVersion apiVersion, correlationID, size int32) {
-	h := requestHeader{
-		ApiKey:        int16(apiKey),
-		ApiVersion:    int16(apiVersion),
-		CorrelationID: correlationID,
-		ClientID:      clientID,
-	}
-	h.Size = h.size() - 4 + size
-
-	// write message
-	h.writeTo(w)
-}
-
 // describeGroups retrieves the specified groups
 //
 // See http://kafka.apache.org/protocol.html#The_Messages_DescribeGroups
@@ -157,10 +144,7 @@ func (c *Conn) describeGroups(request describeGroupsRequestV1) (describeGroupsRe
 
 	err := c.readOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, describeGroupsRequest, v1, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(describeGroupsRequest, v1, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -188,10 +172,7 @@ func (c *Conn) findCoordinator(request findCoordinatorRequestV1) (findCoordinato
 
 	err := c.readOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, groupCoordinatorRequest, v1, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(groupCoordinatorRequest, v1, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -217,10 +198,7 @@ func (c *Conn) heartbeat(request heartbeatRequestV1) (heartbeatResponseV1, error
 
 	err := c.writeOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, heartbeatRequest, v1, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(heartbeatRequest, v1, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -246,10 +224,7 @@ func (c *Conn) joinGroup(request joinGroupRequestV2) (joinGroupResponseV2, error
 
 	err := c.writeOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, joinGroupRequest, v2, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(joinGroupRequest, v2, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -275,10 +250,7 @@ func (c *Conn) leaveGroup(request leaveGroupRequestV1) (leaveGroupResponseV1, er
 
 	err := c.writeOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, leaveGroupRequest, v1, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(leaveGroupRequest, v1, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -304,10 +276,7 @@ func (c *Conn) listGroups(request listGroupsRequestV1) (listGroupsResponseV1, er
 
 	err := c.readOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, listGroupsRequest, v1, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(listGroupsRequest, v1, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -333,10 +302,7 @@ func (c *Conn) offsetCommit(request offsetCommitRequestV3) (offsetCommitResponse
 
 	err := c.writeOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, offsetCommitRequest, v3, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(offsetCommitRequest, v3, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -366,10 +332,7 @@ func (c *Conn) offsetFetch(request offsetFetchRequestV3) (offsetFetchResponseV3,
 
 	err := c.readOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, offsetFetchRequest, v3, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(offsetFetchRequest, v3, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
@@ -402,10 +365,7 @@ func (c *Conn) syncGroups(request syncGroupRequestV1) (syncGroupResponseV1, erro
 
 	err := c.readOperation(
 		func(deadline time.Time, id int32) error {
-			w := &c.wbuf
-			writeHeader(w, c.clientID, syncGroupRequest, v1, id, request.size())
-			request.writeTo(w)
-			return w.Flush()
+			return c.writeRequest(syncGroupRequest, v1, id, request)
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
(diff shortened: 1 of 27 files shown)


conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 	return c
 }
 
+func writeHeader(w *bufio.Writer, clientID string, apiKey apiKey, apiVersion apiVersion, correlationID, size int32) {
```
achille-roussel: Any reason you couldn't reuse the `(*Conn).writeRequest` method?

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
+				return (&response).readFrom(&c.rbuf, size)
+			}())
+		},
+	)
+	if err != nil {
+		return response, err
+	}
+	for _, group := range response.Groups {
+		if group.ErrorCode != 0 {
+			return response, Error(group.ErrorCode)
```
achille-roussel: Can we return `describeGroupsResponseV1{}` instead of `response` when there is a non-nil error? It seems less error-prone to return the zero-value when an error occurs instead of leaking values.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
+			return expectZeroSize(func() (remain int, err error) {
+				return (&response).readFrom(&c.rbuf, size)
+			}())
+		},
+	)
+	if err != nil {
+		return response, err
+	}
+	if response.ErrorCode != 0 {
+		return response, Error(response.ErrorCode)
```
achille-roussel: Same comment here on returning the zero-value on error.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
```
achille-roussel: Same comment here on returning the zero-value on error.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
```
achille-roussel: Same comment here on returning the zero-value on error.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
```
achille-roussel: Same comment here on returning the zero-value on error.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
```
achille-roussel: Same comment here on returning the zero-value on error.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
+			}())
+		},
+	)
+	if err != nil {
+		return response, err
+	}
+	for _, r := range response.Responses {
+		for _, pr := range r.PartitionResponses {
+			if pr.ErrorCode != 0 {
+				return response, Error(pr.ErrorCode)
```
achille-roussel: Same comment here on returning the zero-value on error.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
+	if err != nil {
+		return response, err
+	}
+	if response.ErrorCode != 0 {
+		return response, Error(response.ErrorCode)
+	}
+	for _, r := range response.Responses {
+		for _, pr := range r.PartitionResponses {
+			if pr.ErrorCode != 0 {
+				return response, Error(pr.ErrorCode)
```
achille-roussel: Same comment here on returning the zero-value on error.

conn.go
```diff
@@ -136,6 +136,293 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
+			return expectZeroSize(func() (remain int, err error) {
+				return (&response).readFrom(&c.rbuf, size)
+			}())
+		},
+	)
+	if err != nil {
+		return response, err
+	}
+	if response.ErrorCode != 0 {
+		return response, Error(response.ErrorCode)
```
achille-roussel: Same comment here on returning the zero-value on error.

conn_test.go
```diff
@@ -398,6 +452,252 @@ func testConnReadWatermarkFromBatch(t *testing.T, conn *Conn) {
...
+			switch err {
+			case GroupCoordinatorNotAvailable:
+				continue
+			default:
+				t.Fatalf("bad findCoordinator: %s", err)
+			}
+		}
+
+		if response.Coordinator.NodeID == 0 {
+			t.Fatalf("bad NodeID")
```
achille-roussel: Could you replace the use of Fatalf with Errorf? In general Fatal* is useful for errors that occur within the initialization phase of a test (because there's no way the test can succeed if a connection cannot be established for example).

However when verifying the state of the system after the execution of whatever code the test exercise, it's more interesting to report all that went wrong instead of just the first problem.

describegroups.go
```diff
@@ -0,0 +1,176 @@
...
+		return
+	}
+	if remain, err = readString(r, remain, &t.Protocol); err != nil {
+		return
+	}
+
+	remain, err = readArrayWith(r, remain, func(r *bufio.Reader, size int) (remain int, err error) {
+		item := describeGroupsResponseMemberV1{}
+		remain, err = (&item).readFrom(r, size)
+		t.Members = append(t.Members, item)
```
achille-roussel: Should we check for `err == nil` before appending the value here? Otherwise the last member appended to the list is very likely meaningless.

joingroup.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	if remain, err = readString(r, remain, &t.MemberID); err != nil {
+		return
+	}
+
+	remain, err = readArrayWith(r, remain, func(r *bufio.Reader, size int) (remain int, err error) {
+		var item joinGroupResponseMemberV2
+		remain, err = (&item).readFrom(r, size)
+
+		t.Members = append(t.Members, item)
```
achille-roussel: Same comment here on checking for `err == nil` before appending.

listgroups.go
```diff
@@ -0,0 +1,82 @@
...
+		return
+	}
+	if remain, err = readInt16(r, remain, &t.ErrorCode); err != nil {
+		return
+	}
+
+	remain, err = readArrayWith(r, remain, func(withReader *bufio.Reader, withSize int) (withRemain int, withErr error) {
+		var item ListGroupsResponseGroupV1
+		withRemain, withErr = (&item).readFrom(withReader, withSize)
+		t.Groups = append(t.Groups, item)
```
achille-roussel: Same comment here on checking for `withErr == nil` before appending.

Also, why name the return values `withRemain` and `withErr` instead of `remain` and `err` like it's done in other places?

syncgroup.go
```diff
@@ -0,0 +1,168 @@
...
+		var partitions []int32
+
+		if remain, err = readString(r, remain, &topic); err != nil {
+			return
+		}
+
+		remain, err = readArrayWith(r, remain, func(r *bufio.Reader, size int) (remain int, err error) {
+			var partition int32
+			remain, err = readInt32(r, size, &partition)
+			partitions = append(partitions, partition)
```
achille-roussel: Same comment here on checking for `err == nil` before appending.

describegroups.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (t *describeGroupsResponseV1) readFrom(r *bufio.Reader, size int) (remain int, err error) {
+	if remain, err = readInt32(r, size, &t.ThrottleTimeMS); err != nil {
+		return
+	}
+
+	remain, err = readArrayWith(r, remain, func(r *bufio.Reader, size int) (remain int, err error) {
+		item := describeGroupsResponseGroupV1{}
+		remain, err = (&item).readFrom(r, size)
+		t.Groups = append(t.Groups, item)
```
achille-roussel: Same comment here on checking for `err == nil` before appending.

joingroup.go
```diff
@@ -0,0 +1,213 @@
...
+
+func (t *protocolMetadata) readFrom(r *bufio.Reader, size int) (remain int, err error) {
+	if remain, err = readInt16(r, size, &t.Version); err != nil {
+		return
+	}
+
+	remain, err = readArrayWith(r, remain, func(r *bufio.Reader, size int) (remain int, err error) {
+		var topic string
+		remain, err = readString(r, size, &topic)
+		t.Topics = append(t.Topics, topic)
```
achille-roussel: Same comment here on checking for `err == nil` before appending.

joingroup.go
```diff
@@ -0,0 +1,213 @@
...
+		return
+	})
+
+	if remain, err = readBytes(r, remain, &t.UserData); err != nil {
```
achille-roussel: This `if` check seems unnecessary, you could just assign `remain` and `err` and return.

joingroup.go
```diff
@@ -0,0 +1,213 @@
...
+	if remain, err = readInt16(r, size, &t.Version); err != nil {
+		return
+	}
+
+	remain, err = readArrayWith(r, remain, func(r *bufio.Reader, size int) (remain int, err error) {
+		var topic string
+		remain, err = readString(r, size, &topic)
+		t.Topics = append(t.Topics, topic)
+		return
+	})
```
achille-roussel: Should the code check for `err != nil` here before moving forward?

conn.go
savaki: Nope.  I looked through the write.go and it appeared the headers were being written by hand, but that might have been for performance.  I'll use writeRequest instead.

conn.go
savaki: I'm open to that.  The consideration was the loss of information.  If I return the zero value, you would love the knowledge of which group failed.  I'm happy to remove it for now since I'm not using the return value.

savaki: I believe this line of reason will also apply to a number of the comments below.

conn_test.go
savaki: Sure, happy to change to that.

listgroups.go
savaki: I must have put down code and come back to it.  If you're good, I'd probably change it to withRemain and withErr everyplace.  I think shadowing of non-err variables can be confusing and difficult to read.

describegroups.go
savaki: Whoops.  May have accidentally deleted my comment.

I can change this if you would like me to, but I'd argue that when err != nil the entire contents of describeGroupsResponseMemberV1 are meaningless with the possible exception of ErrorCode and ErrorMessage which aren't impacted by this pattern.

My guess is that you'd probably question code that tried to make use of a value associated with err != nil.  For example, consider ```err := json.Unmarshal(data, &content)```.  If err != nil, I would consider the entire value of ```content``` suspect.

Happy to change if you feel strongly about this.

achille-roussel: Honestly I don’t feel too strong about either ways, if we don’t leak the value at a higher level it’s kind of OK, but there are places where you’ve checked for the error before appending and places where you don’t. I care more about consistency so it doesn’t mislead the reader into questioning why errors are sometimes checked and sometimes not. I guess my advise would be to pick one programming style and apply it everywhere there’s a similar code pattern.

listgroups.go
achille-roussel: Picking a different name to avoid shadowing variables is fine by me, let’s just aim for consistency and take the same approach everywhere then :+1:

describegroups.go
savaki: It's easy enough to change.

savaki:
@achille-roussel Let me know if there's an additional changes you'd like me to make.

joingroup.go
```diff
@@ -0,0 +1,212 @@
...
+		sizeofBytes(t.UserData)
+}
+
+func (t protocolMetadata) writeTo(w *bufio.Writer) {
+	writeInt16(w, t.Version)
+	writeStringArray(w, t.Topics)
+	writeBytes(w, t.UserData)
+}
+
+func (t protocolMetadata) bytes() []byte {
```
achille-roussel: Those `bytes() []byte` methods don't seem to be used anywhere, which means they are also untested. Can we get rid of them and only introduce them if we need to?

joingroup.go
```diff
@@ -0,0 +1,212 @@
+package kafka
+
+import (
+	"bufio"
+	"bytes"
+)
+
+type memberProtocolMetadata struct {
+	// MemberID assigned by the group coordinator or null if joining for the
```
achille-roussel: Is `null` the `"null"` string here? Or the zero-value of a `string`?

joingroup.go
savaki: ```bytes()``` is used by newJoinGroupRequestV2 and is needed as part of the consumer group handshake to declare the participation in the consumer group.

joingroup.go
savaki: It's the zero-value of a string.  I can update the docs to clarify.

joingroup.go
achille-roussel: Ah! I searched the page but Github hid some larger files so I missed it. All good 👍

achille-roussel: Actually I can't find `newJoinGroupRequestV2`, the only place I see it used is in a test https://github.com/segmentio/kafka-go/pull/47/files#diff-9d43d24e28e45a8fd8a0ffe03a0e925aR17

Am I missing something?

savaki: Ah right, I'm looking on different branch. I'll go ahead and backport the tests.

PR #48, opened 2017-12-27
handle edge case where Batch.close called during readMessage

savaki:
Topic and Partition are assigned to message prior to releasing mutex


savaki:
Still one more edge case

PR #50, opened 2017-12-28
Reduce test times

savaki:
reduce test time by converting time.Sleep(...) to createTopic


savaki:
Shaved about 30-40% off the test times on my local machine.

PR #52, opened 2017-12-29
created Hash balancer for producer compatibility with sarama

savaki:
by default, sarama uses the fnv1a algorithm to determine which partition to route messages to.  

To ensure that messages are routed to the same partition, I've added a Hash balancer that uses the same logic.

Even if one is not coming from sarama, having a partitioner that routes messages with the same key to the same partition is critical.

### DIFF
--- a/balancer.go
+++ b/balancer.go
@@ -131,7 +131,7 @@ type Hash struct {
 
 func (h *Hash) Balance(msg Message, partitions ...int) (partition int) {
 	if msg.Key == nil {
-		return (&h.rr).Balance(msg, partitions...)
+		return h.rr.Balance(msg, partitions...)
 	}
 
 	hasher := fnv1aPool.Get().(hash.Hash32)
@@ -139,7 +139,7 @@ func (h *Hash) Balance(msg Message, partitions ...int) (partition int) {
 
 	hasher.Reset()
 	if _, err := hasher.Write(msg.Key); err != nil {
-		return (&h.rr).Balance(msg, partitions...)
+		return h.rr.Balance(msg, partitions...)
 	}
 
 	// uses same algorithm that Sarama's hashPartitioner uses
--- a/balancer.go
+++ b/balancer.go
@@ -1,6 +1,11 @@
 package kafka
 
-import "sort"
+import (
+	"hash"
+	"hash/fnv"
+	"sort"
+	"sync"
+)
 
 // The Balancer interface provides an abstraction of the message distribution
 // logic used by Writer instances to route messages to the partitions available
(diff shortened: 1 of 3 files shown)


balancer.go
```diff
@@ -104,3 +109,44 @@ func (lb *LeastBytes) makeCounters(partitions ...int) (counters []leastBytesCoun
...
+// Note that this is the same algorithm used by the Sarama Producer and ensures
+// that messages produced by kafka-go will be delivered to the same topics that
+// the Sarama producer would be delivered to
+type Hash struct {
+	rr RoundRobin
+}
+
+func (h *Hash) Balance(msg Message, partitions ...int) (partition int) {
+	if msg.Key == nil {
+		return (&h.rr).Balance(msg, partitions...)
```
achille-roussel: Sarama's implementation uses a random distribution if the key is `nil` (see https://github.com/Shopify/sarama/blob/8093021a9bebdd8046c88bbc2c6f747e5a03cf66/partitioner.go#L114-L116).

I don't know what's best here, round robin likely would give a better balancing but would also not be exactly the same behavior as Sarama's so switching to kafka-go would result in a behavior change.

We should document the behavior when the key is `nil` and if it's different from the one implemented in Sarama we should also make it clear (since the documentation currently claims that it's the same algorithm).

Another alternative is to allow the program to configure the fallback balancer in the `Hash` structure by exporting the field.

balancer.go
```diff
@@ -104,3 +109,44 @@ func (lb *LeastBytes) makeCounters(partitions ...int) (counters []leastBytesCoun
...
+	}
+
+	hasher := fnv1aPool.Get().(hash.Hash32)
```
achille-roussel: Sarama allows setting a custom hashing function, I feel like this is almost a requirement because FNV is not a good fit for all keys.

I'd advocate for allowing a hashing function to be configured on the `Hash` structure, something like this:
```go
type Hash struct {
    Func func(key []byte) uint32
}
```
Then a program can set whatever it wants (doing optimizations with a memory pool if needed), and kafka-go defaults Saram's algorithm.

achille-roussel:
I'm looking more into Sarama's Partitioner and in the `ProducerMessage` keys are of type `Encoder` to be transformed before being hashed. I don't see a need for implementing this if we allow programs to set the hash function but we should likely mention in the documentation that it's a voluntary choice and what the alternative is.
(see https://github.com/Shopify/sarama/blob/8093021a9bebdd8046c88bbc2c6f747e5a03cf66/partitioner.go#L117-L120)

balancer.go
achille-roussel: Alternatively we could use a field of type `hash.Hash32` since the `Balancer` documentation states that calls to `Balance` are synchronized, we can safely reuse the same hash, calling Reset to reinitialize it.

balancer.go
savaki: True, but since the random distribution isn't something that can be counted on by consumer, I felt that using one of the Balancers that kafka-go already provided was acceptable.  The driving issue for me was having interoperability between sarama and kafka-go.

Happy to create a random balancer if you feel this is important.

balancer.go
savaki: Sure, happy to do it the sarama way.  I saw what sarama had done, but opted not to expose the hasher to minimize the surface area I was touching.

savaki:
It's true, that transformations are optional.  But in practice, the common encoders, sarama.ByteEncoder and sarama.StringEncoder are just plain type wrappers that do a noop transform to []byte.

balancer.go
savaki: The use case I was using as a driver is, if I had a heterogeneous environment with both kafka-go and sarama would I be surprised.  If I set the message key, then I expect messages with the same key to arrive on the same partition.  If I don't set the key, then I have no expectation as to which partition the message will arrive.

balancer.go
achille-roussel: Cool, I really think it makes sense here to allow the algorithm to be configurable on the `Hasher` type, it's just one field and doesn't even require making an extra function.

balancer.go
achille-roussel: You're right, random cannot be relied on, and round robin will give a more balanced distribution. Let's keep it this way 👍

balancer.go
```diff
@@ -104,3 +109,44 @@ func (lb *LeastBytes) makeCounters(partitions ...int) (counters []leastBytesCoun
...
+	defer fnv1aPool.Put(hasher)
+
+	hasher.Reset()
+	if _, err := hasher.Write(msg.Key); err != nil {
+		return (&h.rr).Balance(msg, partitions...)
```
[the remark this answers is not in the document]
savaki: Works for me.

PR #55, opened 2018-01-05
added consumer group support to Reader

savaki:
I thought I would open a pull request early to get your initial feedback on the request.  This is a working version of consumer group support for the Reader.  I've manually verified interoperability with Sarama.  In addition, I've included an automated  compatibility test with sarama.

Almost all the changes are additive.  The main change was around (*Reader).start that now supports multiple readers and an initial offset.  That said, I tried to preserve the spirit of what I believe start was doing.

Would love to get your thoughts.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1093,6 +1093,15 @@ func (r *Reader) Close() error {
 	r.stop()
 	r.join.Wait()
 
+	if r.useConsumerGroup() {
+		// gracefully attempt to leave the consumer group on close
+		if generationID, membershipID := r.membership(); generationID > 0 && membershipID != "" {
+			if conn, err := r.coordinator(); err == nil {
+				_ = r.leaveGroup(conn)
+			}
+		}
+	}
+
 	<-r.done
 
 	if !closed {
--- a/reader.go
+++ b/reader.go
@@ -1787,7 +1787,7 @@ func (r *reader) withErrorLogger(do func(*log.Logger)) {
 // extractTopics returns the unique list of topics represented by the set of
 // provided members
 func extractTopics(members []memberGroupMetadata) []string {
-	visited := map[string]struct{}{}
+	var visited = map[string]struct{}{}
 	var topics []string
 
 	for _, member := range members {
--- a/reader.go
+++ b/reader.go
@@ -72,7 +72,7 @@ type Reader struct {
 	stop         context.CancelFunc
 	done         chan struct{}
 	commits      chan Message
-	version      int64
+	version      int64 // version holds the generation of the spawned readers
 	offset       int64
 	lag          int64
 	closed       bool
@@ -495,7 +495,7 @@ func (r *Reader) fetchOffsets(subs map[string][]int32) (map[int]int64, error) {
 	return offsetsByPartition, nil
 }
 
-func (r *Reader) subscribe(subs map[string][]int32, ready func()) error {
+func (r *Reader) subscribe(subs map[string][]int32) error {
 	// always clear prior subscriptions
 	r.unsubscribe()
 
@@ -511,8 +511,6 @@ func (r *Reader) subscribe(subs map[string][]int32, ready func()) error {
 	r.mutex.Lock()
 	r.start(offsetsByPartition)
 	r.mutex.Unlock()
-
-	ready() // mark the Reader as ready to go
 
 	r.withErrorLogger(func(l *log.Logger) {
 		l.Printf("subscribed to partitions: %+v", offsetsByPartition)
@@ -812,9 +810,8 @@ func (r *Reader) commitLoop(stop <-chan struct{}, done chan<- struct{}) {
 	}
 }
 
-func (r *Reader) run(ready func()) {
+func (r *Reader) run() {
 	defer close(r.done)
-	defer ready() // ensures ready called when GroupID == ""
 
 	const (
 		backoffDelayMin = 100 * time.Millisecond
@@ -853,7 +850,7 @@ func (r *Reader) run(ready func()) {
 		go r.commitLoop(commitStop, commitDone)
 
 		// subscribe to assignments
-		if err := r.subscribe(assignments, ready); err != nil {
+		if err := r.subscribe(assignments); err != nil {
 			r.withErrorLogger(func(l *log.Logger) {
 				l.Printf("subscribe failed for consumer group, %v: %v\n", r.config.GroupID, err)
 			})
@@ -1111,6 +1108,13 @@ func NewReader(config ReaderConfig) *Reader {
 		readerStatsPartition = -1
 	}
 
+	// when configured as a consume group, start version as 1 to ensure that only
+	// the rebalance function will start readers
+	version := int64(0)
+	if config.GroupID != "" {
+		version = 1
+	}
+
 	stctx, stop := context.WithCancel(context.Background())
 	r := &Reader{
 		config:  config,
@@ -1131,21 +1135,10 @@ func NewReader(config ReaderConfig) *Reader {
 			// once when the reader is created.
 			partition: strconv.Itoa(readerStatsPartition),
 		},
+		version: version,
 	}
 
-	// When using consumer groups, we want to ensure that the rebalancer
-	// has first dibs on starting reader goroutines.  To that end, we
-	// provide the rebalancer a ready func() that it will call to indicate
-	// its done what it needs to do (which may be nothing when GroupID == "")
-	readCtx, ready := context.WithCancel(context.Background())
-	defer ready()
-
-	go r.run(ready)
-
-	select {
-	case <-stctx.Done():
-	case <-readCtx.Done():
-	}
+	go r.run()
 
 	return r
 }
--- a/reader.go
+++ b/reader.go
@@ -1,11 +1,15 @@
 package kafka
 
 import (
+	"bufio"
+	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
 	"math"
+	"sort"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -15,6 +19,39 @@ import (
 const (
 	firstOffset = -1
 	lastOffset  = -2
+)
+
+var (
+	// errNotAvailable returned when func called that is not supported in cluster mode
+	errNotAvailable = errors.New("not available when GroupID is set")
+)
+
+const (
+	// defaultProtocolType holds the default protocol type documented in the
+	// kafka protocol
+	//
+	// See https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-GroupMembershipAPI
+	defaultProtocolType = "consumer"
+
+	// defaultHeartbeatInterval contains the default time between heartbeats.  If
+	// the coordinator does not receive a heartbeat within the session timeout interval,
+	// the consumer will be considered dead and the coordinator will rebalance the
+	// group.
+	//
+	// As a rule, the heartbeat interval should be no greater than 1/3 the session timeout
+	defaultHeartbeatInterval = 3 * time.Second
+
+	// defaultSessionTimeout contains the default interval the coordinator will wait
+	// for a heartbeat before marking a consumer as dead
+	defaultSessionTimeout = 30 * time.Second
+
+	// defaultRebalanceTimeout contains the amount of time the coordinator will wait
+	// for consumers to issue a join group once a rebalance has been requested
+	defaultRebalanceTimeout = 10 * time.Second
+
+	// defaultRetentionTime holds the length of time a the consumer group will be
+	// saved by kafka
+	defaultRetentionTime = time.Hour * 24
 )
 
 // Reader provides a high-level API for consuming messages from kafka.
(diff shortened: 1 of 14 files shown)


reader.go
```diff
@@ -484,28 +1462,47 @@ func (r *Reader) readLag(ctx context.Context) {
...
+				maxWait:     r.config.MaxWait,
+				version:     r.version,
+				msgs:        r.msgs,
+				stats:       &r.stats,
+			}).run(ctx, offset)
+		}(ctx, partition, offset, &r.join)
+	}
+}
+
+func (r *Reader) CommitMessage(m Message) error {
```
achille-roussel: Should we do this automatically as part of the `ReadMessage` method? I feel like otherwise a program has keep awareness of whether a Reader is part of a consumer group to know if it must call `CommitMessage` or not.

I feel like managing the commits automatically would better be inline with the current design where the offsets are already automatically managed when directly addressing partitions.

If this is a critical feature for one of your systems I think we should introduce a new method (say `PeekMessage`) which doesn't do automatic offset management and leaves it up to the program to explicitly commit messages.

reader.go
```diff
@@ -824,3 +1819,25 @@ func (r *reader) withErrorLogger(do func(*log.Logger)) {
 		r.withLogger(do)
 	}
 }
+
+// extractTopics returns the unique list of topics represented by the set of
+// provided members
+func extractTopics(members []memberGroupMetadata) []string {
+	visited := map[string]struct{}{}
```
achille-roussel: Style: keeping one variable declaration style within the local scope helps with readability, for example here you could rewrite this as
```go
var visited = map[string]struct{}{}
var topics []string
```

strategy.go
```diff
@@ -0,0 +1,182 @@
...
+
+// strategy encapsulates the client side rebalancing logic
+type strategy interface {
+	// ProtocolName of strategy
+	ProtocolName() string
+
+	// ProtocolMetadata returns the metadata for the protocol
+	ProtocolMetadata(topics []string) (groupMetadata, error)
+
+	// DefineMemberships returns which members will be consuming
```
achille-roussel: This documentation seems incorrect.

strategy.go
```diff
@@ -0,0 +1,182 @@
+package kafka
+
+import "sort"
+
+// strategy encapsulates the client side rebalancing logic
+type strategy interface {
+	// ProtocolName of strategy
+	ProtocolName() string
```
achille-roussel: Any reason not to non-exported method names in this interface? (since the interface is non-exported itself)

syncgroup.go
```diff
@@ -37,6 +37,10 @@ func (t groupAssignment) writeTo(w *bufio.Writer) {
 }
 
 func (t *groupAssignment) readFrom(r *bufio.Reader, size int) (remain int, err error) {
+	if size == 0 {
```
achille-roussel: Are there valid use cases where there will be no more bytes left to read even tho we expect to read a `groupAssignment` structure?

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+			}
+		}
+	}
+
+	return offsetsByPartition, nil
+}
+
+func (r *Reader) subscribe(subs map[string][]int32) error {
+	// always clear prior subscriptions
+	r.unsubscribe()
```
achille-roussel: It seems like `subscribe` is only called it one place, can we move this call out of this method so we keep a single responsibility per method?

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+
+	if len(subs[r.config.Topic]) == 0 {
```
achille-roussel: Would this be better expressed as
```go
if _, exists := subs[r.config.Topic]; !exists {
    return nil
}
```

Checking the length assuming to get the zero-value if it doesn't exist seems less straight forward.

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+func (r *Reader) waitThrottleTime(throttleTimeMS int32) {
+	if throttleTimeMS == 0 {
+		return
+	}
+
+	select {
+	case <-r.stctx.Done():
+		return
+
+	case <-time.After(time.Duration(throttleTimeMS) * time.Millisecond):
```
achille-roussel: Can we replace the use of `time.After` for a `timer.Timer`? That way we can stop it on return and we don't leak resources if `stctx` is canceled.

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+				// offsets should only increase
+				continue
+			}
+
+			offsetsByPartition[m.Partition] = m.Offset
+		}
+	}
+}
+
+func (r *Reader) commitLoop(stop <-chan struct{}, done chan<- struct{}) {
```
achille-roussel: Instead of a `stop` channel here we could use a context, that way it could inherit from `r.stctx` and we would get cancellation propagation if either the base context is canceled or only the one we created for the commit loop.

The same logic applies for the heartbeat loop.

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+			return err
+		}
+	}
+
+	r.waitThrottleTime(resp.ThrottleTimeMS)
+
+	return nil
+}
+
+func (r *Reader) hbLoop(stop <-chan struct{}, done chan<- struct{}) {
```
achille-roussel: How about naming this method `heartbeatLoop`? `hb*` is a bit hard to read.

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+		assignments, err := r.rebalance()
+		if err != nil {
+			r.withErrorLogger(func(l *log.Logger) {
+				l.Printf("rebalance failed for consumer group, %v: %v", r.config.GroupID, err)
+			})
+			continue
+		}
+
+		// start the heartbeat
+		hbStop, hbDone := make(chan struct{}), make(chan struct{})
```
achille-roussel: It seems like a `sync.WaitGroup` would be a better construct to wait for all sub-goroutines to complete instead of waiting on multiple "done" channels.

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+
+		break
+	}
+
+	return nil
+}
+
+// newJoinGroupRequestV2 handles the logic of constructing a joinGroup
+// request
+func (r *Reader) newJoinGroupRequestV2() (joinGroupRequestV2, error) {
```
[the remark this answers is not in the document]
savaki: Good to know.

reader.go
savaki: Perhaps we could go through some use cases.  

For us, it is critical that messages are processed before we commit the message off the Reader which requires an explicit commit.  Consider what would happen if one were performing db replication or event sourcing if a message that was read for reason failed to be processed.  Instead of INSERT > UPDATE > DELETE we could end up with just INSERT > UPDATE

That said, I can see some cases where keeping track of the offsets might not be as critical and a lossy, automatic commit would be acceptable.  Perhaps an AutoCommit flag on ReaderConfig?

strategy.go
savaki: Two reasons, neither strong.  

The first is I have a personal pattern where I extend the Go style to apply to struct > struct communication.  Even if the type is lower case, I prefer the func to be uppercase to clearly indicate that it should be shared.

The second is that strategy is similar to Balancer, you can easily see others wanting to implement their own Strategies.  For example, a rack-aware consumer strategy.  In that case, uppercasing strategy is all that's required.

syncgroup.go
savaki: I came across this case when testing for compatibility with bsm/sarama-cluster.  It appears in some cases, sarama-cluster can send a nil array entry.  Admittedly, I didn't look too closely at it.

reader.go
savaki: Sure, happy to change it. 

I figured it wouldn't be a problem since the loop is slow and resources are freed once time.After fires.  It's not so much a leak as holding resources longer than necessary.

reader.go
savaki: Happy to change it.  Just my idiosyncrasy.  Feels a little awkward to leave an edge case open that someone in the future could stumble upon.

reader.go
savaki: yep.  felt a little awkward to me as well, but I saw

```
r.cancel() // always cancel the previous reader
```
 
in start which seemed conceptually similar.

reader.go
```diff
@@ -29,31 +66,847 @@ type Reader struct {
...
+			})
+			return nil, err
+
+			// all possible error codes returned by join group
+		//case GroupLoadInProgress:
+		//case GroupCoordinatorNotAvailable:
+		//case NotCoordinatorForGroup:
+		//case InconsistentGroupProtocol:
+		//case InvalidSessionTimeout:
+		//case GroupAuthorizationFailed:
```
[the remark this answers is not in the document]
savaki: Makes sense.

reader.go
```diff
@@ -29,31 +66,770 @@ type Reader struct {
...
+			ProtocolName:     strategy.ProtocolName(),
+			ProtocolMetadata: meta.bytes(),
+		})
+	}
+
+	return request, nil
+}
+
+// newMemberProtocolMetadata maps encoded member metadata ([]byte) into memberGroupMetadata
+func (r *Reader) newMemberProtocolMetadata(in []joinGroupResponseMemberV2) ([]memberGroupMetadata, error) {
```
achille-roussel: Should we rename this “make” as well? It returns a value from a make call, seems like it would make sense.

reader.go
```diff
@@ -29,31 +66,770 @@ type Reader struct {
...
+	}
+
+	r.withLogger(func(l *log.Logger) {
+		l.Printf("joinGroup succeeded for response, %v.  generationID=%v, memberID=%v\n", r.config.GroupID, response.GenerationID, response.MemberID)
+	})
+
+	return assignments, nil
+}
+
+func (r *Reader) newSyncGroupRequestV1(memberAssignments memberGroupAssignments) syncGroupRequestV1 {
```
achille-roussel: Same here, seems like this method should be named “make*”

reader.go
```diff
@@ -297,6 +1178,10 @@ func (r *Reader) ReadMessage(ctx context.Context) (Message, error) {
 // The function returns a lag of zero when the reader's current offset is
 // negative.
 func (r *Reader) ReadLag(ctx context.Context) (lag int64, err error) {
+	if r.config.GroupID != "" {
```
achille-roussel: This is tested plenty of times, how about adding a method to help make the code more readable (something like r.useConsumerGroup())

reader.go
achille-roussel: Do you have a concern about adding a new method that doesn’t do the auto-commit?

I feel like that would make it explicit whether CommitMessage has to be called or not. I would really prefer if we can keep the auto-commit behavior in ReadMessage. 

Also what do you think about changing the commit method to be called CommitMessages and accept an arbitrary number of messages as argument? That way the program acquired some ownership over the batching of commits (if we can make the commits synchronized I can see this being useful for some of our use cases at Segment, but we must get the API right to allow this).

strategy.go
achille-roussel: Fair, I can see us exposing the strategy interface at some point :+1:

syncgroup.go
achille-roussel: How about leaving a comment about it here? So anyone reading this code knows why we’re making the check.

reader.go
achille-roussel: PS: what do you think about passing a context as well to the CommitMeasages method? That way we also give control to the application on how long it may wait for the commit to be published?

savaki: I think I'm just very used to thinking of read messages as being a singular activity and autocommit as a property of the connection rather than having ReadMessage vs PeekMessage/CommitMessage.  How would you feel about FetchMessage/CommitMessage for naming which would be more in sync with Kafka's naming.

savaki: Here are the commit use cases that I'm thinking of.

1. fetch message, process message, commit message used by most of our use cases.  Since we require all services support at least once processing, async commits are fine since replays will be ignored.

2. fetch/process/commit for exactly once processing.  I'm not sure this is a thing with the api as we have it.  I believe Confluent created a new api to support these semantics, but admittedly I haven't read up on it.  Exactly once processing would really have to be at least once since there's no two phase commit.

3. fetch/process/batch-commit for micro-batch processing.

4. fetch/process/no-commit e.g. live statistics.  

4 doesn't require a commit, so that's easy.  1 & 2 only require a single message commit where 3 requires an arbitrarily sized batch.  

What about having two commit methods?  Sync/async could still be controlled by the CommitInterval field.

```
Commit(ctx context.Context, offsets Offsets) error
CommitMessage(m kafka.Message)
```

1 & 2 seems like they would almost always want to an async commit with a single message.  Having a ctx seems like it would be confusing.  

3 seems like it could benefit from a sync commit commit.  By separating the calls, we could also separate the shape of the data being sent.

For comparison, sarama-cluster uses the following:

```
func (c *Consumer) MarkOffset(msg *sarama.ConsumerMessage, metadata string)
func (c *Consumer) MarkOffsets(s *OffsetStash)
```

syncgroup.go
savaki: Sure, that makes sense.

reader.go
achille-roussel: What I really want to avoid is creating a dependency between how the Reader is configured and how it is used, which is why I would prefer to have auto commits on ReadMessage. That way, no matter what the configuration is, the code does not need to be changed. Otherwise, the code using the Reader needs to be aware of its configuration to know whether it has to commit or not, or it has to maybe ignore the error, or we make the commit a no-op if it’s not needed? All those patterns seem less robust than splitting the APIs to have the program explicitly control how it consumes messages when it needs to.

FetchMessage sounds good to me (I actually like it better than Peek).

I feel like making the commit sync or async can be configured on the Reader, because in this case it doesn’t impact the way the program uses the code, it only makes a blocking operation non-blocking. This is actually the approach we took in the Writer to configure whether the writes are done synchronously it asynchronously. 

I’m not concerned about passing a context for cases where commits are asynchronous, very likely the program would use the same context that it passed to FetchMessage.

I’d like to avoid introducing a new Offset concept here, I liked your proposal of passing the message to commit, it makes it trivial to use the API, otherwise the program needs to extract the offsets etc...

It seems to me that a method with this signature would cover all use cases:
```go
CommitMessages(ctx context.Context, msg ...Message) error
```
(Plus it’s consistent with the model used with WriteMessages in the Writer)

savaki: Works for me.

savaki:
Sure, would be happy to.  There's additional tests I'd like to create on handling consumer groups under various edge cases, but I'd be good with merging as well.

achille-roussel:
Cool, let's not block the merge then, feel free to send follow up PRs to add more tests 👍 

Thanks a ton for your contribution on this, it's massive 🙇

PR #58, opened 2018-01-09
add CommitMessages documentation

achille-roussel:
@savaki this PR adds documentation on the `CommitMessages` method, and moves it close to `FetchMessage` in the file to make it easier to read the code (since those are closely related).


savaki:
Looks good to me.

PR #59, opened 2018-01-09
Revisit sync commits

achille-roussel:
@savaki I modified the `CommitMessages` implementation in the case where the reader is configured to use synchronous commits, in which case the method now actually waits for the commit to complete and any error is propagated back to the program.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1085,7 +1085,7 @@ func NewReader(config ReaderConfig) *Reader {
 		msgs:    make(chan readerMessage, config.QueueCapacity),
 		cancel:  func() {},
 		done:    make(chan struct{}),
-		commits: make(chan commitRequest, config.QueueCapacity),
+		commits: make(chan commitRequest),
 		stop:    stop,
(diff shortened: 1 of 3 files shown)


reader.go
```diff
@@ -740,44 +722,37 @@ func (r *Reader) commitLoopInterval(conn offsetCommitter, stop <-chan struct{})
...
-			}
-
-			r.offsetStash.merge(msgs...)
+		case req := <-r.commits:
+			r.offsetStash.merge(req.commits)
 		}
 	}
 }
 
+func (r *Reader) commitOffsetStash(conn offsetCommitter) error {
```
savaki: given commitOffsetStash is only called from one place and it only performs a single task and is already wrapped in the commit func, is there a reason not to inline it above?

achille-roussel: Not really, I can change that 👍

savaki:
Looks good to me.

achille-roussel:
Thanks for the review 👍

PR #62, opened 2018-01-10
resolves off by 1 compatibility issue with sarama.

savaki:
I'm re-submitting this as it seems to have gotten lost in the shuffle.

Internally, kafka-go offset is consistent with itself. Unfortunately, not with the rest of the world. The commit offset should be the offset of the next message to read and NOT the last message read. Verified by running sarama and kafka-go sequentially to verify they picked up each others offsets.

Seeing as you were working in commits, I thought I would hop on your commit.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -14,7 +14,7 @@ func makeCommit(msg Message) commit {
 	return commit{
 		topic:     msg.Topic,
 		partition: msg.Partition,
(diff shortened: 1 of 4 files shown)


achille-roussel:
Ah! I missed that it was not branched off of master. 

Thanks.

Issue #67, opened 2018-01-31
proposal: expose ability to create topics


jonasrmichel:
Currently, the only way to create a topic is to ensure kafka's `auto.create.topics.enable` is set to `true` (which it is by default) and then to publish a message on a non-existent topic (which will cause a panic). 

This is workable, but kind of dirty feeling to me. If I want to ensure I don't pollute an existing topic with throwaway "create" messages, I need to first check if a topic exists before attempting to create it using this approach.

It would be much nicer if the internal `createTopics` mechanism were exposed to support this use case.

achille-roussel:
This is a great idea, would you be able to send a pull request that implements this change?

PR #68, opened 2018-02-02
expose ability to create topics (#67)

jonasrmichel:
* exposes `CreateTopics` on a connection
* "hides" internal protocol request types with aliases


createtopics.go
```diff
@@ -0,0 +1,218 @@
...
+	return sizeofInt32(t.Partition) +
+		sizeofInt32(t.Replicas)
+}
+
+func (t createTopicsRequestV2ReplicaAssignment) writeTo(w *bufio.Writer) {
+	writeInt32(w, t.Partition)
+	writeInt32(w, t.Replicas)
+}
+
+type TopicConfig = createTopicsRequestV2Topic
```
achille-roussel: Can we make this its own type instead of aliasing to an internal type? Go type aliases were designed to facilitate code migrations, this could mislead the reader into interpreting that the code is being refactored.

Unexported types also aren't exposed by godoc, which means the reader needs to go look at the source to understand how it can use the code.

jonasrmichel: Good points. I'll see if I can clean this up a bit.

My intent was to preserve the `*RequestV2Topic` types because there's API compatibility information encoded in the type names. However, I can probably accomplish the same thing with an alias in the opposite direction after making the exposed types first class.

jonasrmichel:
@achille-roussel -- Thoughts? I'd love to get this feature into `master`.

jonasrmichel:
@achille-roussel -- All good points. I've tried to address your comments in my most recent commit. I've done away with the type aliases in favor of concrete types and non-exported conversion methods. The exported types don't expose the low level details of the Kafka protocol.

PR #71, opened 2018-02-08
writer stats test fix

jonasrmichel:


jonasrmichel:
@achille-roussel -- I neglected to update a function signature in the `writer_test`, which caused tests to fail. Fixed now.

achille-roussel:
Do you know why the diff against master is showing so many changes? Seems like it should only touch writer_test.go

jonasrmichel:
Sorry, the PR included both the initial `WriterStats` work and the fix to the test file. Fixed now.

PR #73, opened 2018-02-22
fix error message for errNotAvailable

jsteenb2:
The error message was a bit confusing. Read opposite to what the intended behavior. Small change to better represent the invalid use case it describes.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -29,7 +29,8 @@ const (
 
 var (
 	// errNotAvailable returned when func called that is not supported in cluster mode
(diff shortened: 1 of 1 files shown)


jsteenb2:
I like that idea, quick question, me being unaware here, where is the `errNotAvailableWithGroup` used?

jsteenb2:
Good catch, intelliJ was ignoring the entire repo 😬.  Added in the separate err types and changed the error messages in line to use them.

jsteenb2:
👍 my pleasure

PR #75, opened 2018-02-23
Automatically set Time when publishing the messages

Pryz:
If not set at the creation of the message, we need to set the message timestamp.

By not setting this timestamp, messages will be deleted at the next log cleaner iteration.

### DIFF
--- a/message.go
+++ b/message.go
@@ -15,7 +15,10 @@ type Message struct {
 	Offset    int64
 	Key       []byte
 	Value     []byte
-	Time      time.Time
+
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -15,7 +15,10 @@ type Message struct {
 	Offset    int64
 	Key       []byte
 	Value     []byte
-	Time      time.Time
+
+	// If not set at the creation, Time will be automatically set when
+	// writing the message
```
achille-roussel: Finish with a period (`.`)?

PR #80, opened 2018-02-28
Add ApiVersions

Pryz:
This PR adds the `ApiVersions` function to the API.

### DIFF
--- a/apiversions.go
+++ b/apiversions.go
@@ -1,0 +1,166 @@
+package kafka
+
+import (
+	"bufio"
+)
+
+type apiVersionsRequestV1 struct {
+}
+
+func (a apiVersionsRequestV1) size() int32 {
+	return 0
+}
+
+func (a apiVersionsRequestV1) writeTo(w *bufio.Writer) {
+}
+
+type apiVersionsResponseV1 struct {
+	ErrorCode      int16
+	ApiVersions    []apiVersionsV1
+	ThrottleTimeMs int32
+}
+
+func (a apiVersionsResponseV1) size() int32 {
+	return sizeofInt16(a.ErrorCode) +
+		sizeofArray(len(a.ApiVersions), func(i int) int32 { return a.ApiVersions[i].size() }) +
+		sizeofInt32(a.ThrottleTimeMs)
+}
+
+func (a apiVersionsResponseV1) writeTo(w *bufio.Writer) {
+	writeInt16(w, a.ErrorCode)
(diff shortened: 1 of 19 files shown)


apiversions.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if remain, err = readInt32(r, size, &a.ThrottleTimeMs); err != nil {
+		return
+	}
+
+	return
+}
+
+type ApiVersionsV1 struct {
```
achille-roussel: This is a protocol type, it shouldn't be exported.

conn.go
```diff
@@ -989,6 +989,31 @@ func (c *Conn) requestHeader(apiKey apiKey, apiVersion apiVersion, correlationID
 	}
 }
 
+// ApiVersions return the list of API versions supported by the broker.
+func (c *Conn) ApiVersions() (versions []ApiVersionsV1, err error) {
```
achille-roussel: Is there a reason why we would need to call this from a program? If not we shouldn't export it.

conn.go
```diff
@@ -989,6 +989,31 @@ func (c *Conn) requestHeader(apiKey apiKey, apiVersion apiVersion, correlationID
...
+			versions = response.ApiVersions
+
+			return nil
+		},
+	)
+	if err != nil {
+		return []ApiVersionsV1{}, err
+	}
+
+	return
```
achille-roussel: You're mixing names and unnamed return values, this is a bit confusing to the reader.

docker-compose.yml
```diff
@@ -5,7 +5,7 @@ kafka:
   ports:
     - "9092:9092"
   environment:
-    KAFKA_VERSION: "0.10.1.0"
+    KAFKA_VERSION: "0.11.0.1"
```
achille-roussel: I double-checked and it looks like this environment variable isn't used anymore, we should use an explicit tag on the image instead.

Pryz: I think it's still in used : https://github.com/wurstmeister/kafka-docker/blob/master/Dockerfile#L8

conn.go
Pryz: I would say that it depends if the strategy is to let the user defining which protocol version to use or if we do that automatically in the library.

achille-roussel: Let's do it automatically, kafka-go aims to be simple to use, this kind of implementation complexity is worth abstracting.

Pryz:
I'm going to close this PR. The first idea was just to introduce the ability to use the ApiVersion API and it then became the "dealing with all the Kafka API versions".

Instead of doing that, we are going to rethink the internals of the library and use the latest message formats.

PR #83, opened 2018-03-19
Fix alignment issues with stats on 32-bit systems

nathanielc:
Go's atomic package requires that the 64bit operations, are 64-bit aligned on 32-bit systems. See the last note here: https://golang.org/pkg/sync/atomic/#pkg-note-BUG

This PR comments about the issue and fixes two instances where atomic operations were being performed on non-aligned int64s resulting in a panic. The fix was simply to allocate the stats structs since the start of structs are guaranteed to be 64-bit aligned when allocated.

The PR also removes the unsafe usage in the stats.go file since it was not necessary.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -93,7 +93,9 @@ type Reader struct {
 	// reader stats are all made of atomic values, no need for synchronization.
 	once  uint32
 	stctx context.Context
-	stats readerStats
+	// reader stats are all made of atomic values, no need for synchronization.
+	// Use a pointer to ensure 64-bit alignment of the values.
+	stats *readerStats
 }
 
 // useConsumerGroup indicates whether the Reader is part of a consumer group.
@@ -936,6 +938,7 @@ type ReaderStats struct {
 	Partition string `tag:"partition"`
 }
 
+// readerStats is a struct that contains statistics on a reader.
 type readerStats struct {
 	dials      counter
 	fetches    counter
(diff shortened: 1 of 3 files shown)


achille-roussel:
Thanks for the fix 👍 

`unsafe.Pointer` conversions not being necessary, is this due to recent change in conversion rules in Go? (can't remember if it was 1.9 or 1.10 that changed some of those).

nathanielc:
> unsafe.Pointer conversions not being necessary, is this due to recent change in conversion rules in Go? (can't remember if it was 1.9 or 1.10 that changed some of those).

I am not sure, honestly. I haven't ever needed to use unsafe to cast a specific type back to its more generic type (i.e. *counter -> *int64), so it was surprising to me, that is was there. I have tested the code with both 1.10 and 1.9 and it works fine.

Looking through the release notes for the past several releases didn't turn up anything either.

I can submit the unsafe changes as a secondary PR if there is concern that its not a correct change.

nathanielc:
FWIW I have tested this small program on Go 1.4 and it works as expected:

```
package main

import "sync/atomic"

type counter int64

func (c *counter) ptr() *int64 {
        return (*int64)(c)
}

func (c *counter) inc() {
        atomic.AddInt64(c.ptr(), 1)
}

func main() {
        var c = new(counter)
        c.inc()
        println(c) // prints pointer location
        println(*c) // prints 1
}

```

achille-roussel:
Alright, tha’s good enough for me.

Issue #44, opened 2017-12-26
[savaki/kafka-go/consumer-groups] Discussion


achille-roussel:
@savaki

I'm opening this issue as communication thread on the consumer groups implementation that you're working on there https://github.com/savaki/kafka-go/tree/consumer-groups

First of all, thanks a lot of taking time to work on this, it's a highly demanded feature that I'm sure a lot will benefit from.

Here are some feedback I had on your work:

- All types that represent structured protocol messages (like DescribeGroupsRequestV1 for example) must be non-exported. One of the main goals of the kafka-go project is to hide away implementation details of the kafka protocol into high-level concepts that fit well into Go programs. We must aim for exposing the minimal public API so the package can focus on useful and programmer-friendly concepts. Take a look at how the other protocol messages are implemented for example.

- The file naming that you used is not following the current scheme, aim for shorter names with no underscores (the way it was done for all other files). For example, `api_describe_groups.go` could likely be renamed `describegroups.go` to match the current style.

- The naming scheme you used with the `api` prefix for method names like `apiListGroups`, `apiLeaveGroup` seems unnecessary. Simply naming those methods `listGroups`, `leaveGroups` would make the code shorter and more expressive.

- It is more common in Go to declare duration constants as `<value> * <unit>` instead of the opposite. The reason for this is better readability because `30 * time.Second` kins of read like natural English "30 seconds". For example `defaultSessionTimeout = time.Second * 30` would be better written `defaultSessionTimeout = 30 * time.Second`.

- This is a considerable amount of work, what do you think about splitting it into two steps? First adding all the protocol messages with tests, then adding high-level APIs and actual consumer group implementation. That way we can better focus on the changes being made, ease review and chance of producing quality software.

- What do you think about supporting consumer groups in the `Reader` type instead of introducing a new public API (`ConsumerGroup`). It seems like we could simply add a `Group string` property to the `ReaderConfig` to configure the reader as being part of a group instead of directly reading from a partition. This may be less flexible, for example we lose the ability to read from multiple topics, or from multiple static partitions, but as a first step I think we're better off focusing on  reading from a single topic and keeping the API as simple as possible. This would have the benefit of reusing the existing public API and a bunch of the `Reader` implementaiton, making it easier for programs based on kafka-go to introduce support for consumer groups without having to create abstractions if they need to support both.

savaki:
Thanks for opening this thread.  Here was my thinking on each of those:

[...]

PR #88, opened 2018-03-27
Add Latest option to readerconfig to start reading from latest offset

errnoh:


errnoh:
( discussion at https://github.com/segmentio/kafka-go/issues/44 )

reader.go
```diff
@@ -824,6 +828,9 @@ func (r *Reader) run() {
 // ReaderConfig is a configuration object used to create new instances of
 // Reader.
 type ReaderConfig struct {
+	// If the reading should be started from the latest offset instead of the earliest one
+	Latest bool
```
errnoh: "Latest" might not be descriptive enough though. StartFromLatest? ReadFromLatest? Any ideas?

abraithwaite:
Bringing the conversation here.  As an operator, I would expect a bit more flexibility over being able to set the offsets at startup on a per-partition basis even when using consumer groups.

In my experience there are plenty of times where this is necessary and desirable.  Is there any reason we shouldn't allow for arbitrary setting of offsets at startup @achille-roussel @savaki ?

errnoh:
@abraithwaite Easy way out would be to allow setting the `whence` and `offset` both, which would allow one to have relative offsets (from both beginning or the end). This would still leave out the option to set the exact offset value though, which would indeed need a map (or another struct) to be passed.

Can't see a reason why that wouldn't be an option. The underlying logic does build a map like that anyways, you'd just force it to use the one you pass on the config. Only issue with that is the bloating of readerconfig with options 99% of use cases don't need, but then again this is a Kafka library and use cases like that can be production critical.

reader.go
errnoh: Agreed that the option shouldn't be as the first field.

For the naming the issue with just `Offset` is that while it makes sense in the `FirstOffset` and `LastOffset` sense but setting it to anything other than those two wouldn't do what you'd expect it to do when using consumer groups.

Second issue with `Offset` is that in regard of what the current context is the parameter is more like `Whence` instead of `Offset`. (Which then again is more related to the other discussion happening in this same PR)

abraithwaite: > setting it to anything other than those two wouldn't do what you'd expect it to do when using consumer groups.

I'm not sure I understand.  I'd expect it to start consuming from that offset and continue as normal.

One risk posed when doing this sort of thing is that the consumer can die and restart with the old config.  The I've handled this at $JOB-1 is by running a one off job to reprocess events with since and until offests.  Otherwise we have to figure out the restart story.

I'm not sure what works the best here and I'm open to other ideas, but that seems like a reasonable way forward.

errnoh: The problem is that `offset` only has context in a `topic/partition` pair. This ReaderConfig field was added to be used with a consumer group that handles all partitions in a single topic.

Of course I have a bias here as well since I'm using ReaderConfig for the consumer group while others may use it only to read a single partition for example. In that case it would indeed then work as you'd expect. But in that case setting it to anything other than FirstOffset or LastOffset when using consumer group should probably throw an error.

For generic solution that covers all cases there is the other discussion happening in this same PR. Basically it would mean passing `[partition]offset` map and an optional `whence` which probably would default to absolute value.

abraithwaite: > The problem is that offset only has context in a topic/partition pair. This ReaderConfig field was added to be used with a consumer group that handles all partitions in a single topic.

Correct.  In my experience operating kafka, it doesn't make a lot of sense to abstract away partition details when using consumer groups.  We might need to make some updates to the implementation in this library to reflect that eventually but we're not using the groups implementation widely internally yet so we haven't hit those cases.

achille-roussel:
@errnoh @abraithwaite what's the state of this change? Do you still want to see it happen?

achille-roussel:
I'm going to close this due to inactivity, but feel free to re-open there's interest in getting this change merged again!

cretz:
I am interested in this feature (essentially auto.reset.offset as latest in other clients). Can this now be implemented? If this particular patch is no longer applicable, shall I submit a new PR?

stevevls:
Keep an eye on #277.  There will be an option there for consumer groups to start from the `FirstOffset` (default) or `LastOffset`.  I have a couple more test cases to button up, then I will be merging it.

cretz:
Nice. Please confirm I have the ability to apply that option only for first time the consumer group is seen and to use last committed offset otherwise.

stevevls:
That's correct!

PR #89, opened 2018-03-29
fix bug that when time is zero, writeTime doesn't set to the msgs

CodeRushing:
This bug cause the producer message will be deleted very soon

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -771,7 +771,7 @@ func (c *Conn) WriteMessages(msgs ...Message) (int, error) {
 	}
 
 	writeTime := time.Now()
-	for _, msg := range msgs {
+	for i, msg := range msgs {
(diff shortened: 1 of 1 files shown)


abraithwaite:
Can you provide a little more context?  What's the problem you're observing, and how does this fix it?

As far as I can tell, this won't change the behavior at all (unless I'm missing something).

abraithwaite:
> I think the problem is msg is a local variable, it's not actually modifying the slice (unless I'm missing something).

idiot facepalm gif-source

Makes a copy, and it's not a pointer.  Duh.

CodeRushing:
@achille-roussel  Done

achille-roussel:
Thanks for your contribution!

PR #95, opened 2018-04-29
Make topic and partition read-only

cyberdelia:
This make `Topic` and `Partition` read-only by un-exporting the fields and exposing them only trough the `Topic()` and `Partition` methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ for higher level abstractions, like the `Reader` for example.
 
 A `Reader` is another concept exposed by the `kafka-go` package, which intends
 to make it simpler to implement the typical use case of consuming from a single
-topic-partition pair.  
+topic-partition pair.
 A `Reader` also automatically handles reconnections and offset management, and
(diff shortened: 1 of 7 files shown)


cyberdelia:
I'm not facing any problem, I just saw an opportunity to fix a potential "foot-gun".

It's a "breaking" change, but it will prevent bad code to work, and good code will get a compiler error indicating the type change from `string` to `func() string`. The latter seems a small price to pay compare to any alternative that comes to mind to deprecate this behaviour.

achille-roussel:
You may be right, but there’s also potential value about having the topic and partition exposed on the message. For example, one demanded feature has been to allow the writer to publish to multiple topics, in which case the destination topic could be configured directly on the message.

Unless we have a solid example of code that would have been used improperly due to thus design decision I think we’ll be better off keeping our options open and avoiding a breaking change here. 

Th is for opening the discussion on this topic anyway, it’s always welcome.

PR #97, opened 2018-05-04
Add constants for Seek's whence and introduce a fourth strategy

jnjackins:
This change is intended to be non-breaking.

Since the meaning of `whence` differs subtly from `os.Seek`, name them to make the meanings obvious. Also, add the missing "seek relative to current offset" (previously whence=1 was documented as doing this, but it actually seeks to an absolute offset, whereas 0 is relative to the beginning of the partition.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -463,21 +463,21 @@ func (c *Conn) Offset() (offset int64, whence int) {
 }
 
 const (
-	SeekRelativeStart   = 0 // Seek relative to the first offset available in the partition.
-	SeekAbsolute        = 1 // Seek to an absolute offset.
-	SeekRelativeEnd     = 2 // Seek relative to the last offset available in the partition.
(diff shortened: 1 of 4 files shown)


jnjackins:
Thanks. After sleeping on it, I'm having second thoughts about the names. Maybe this would be better:
```
+const (
+	SeekStart    = 0 // Seek relative to the first offset available in the partition.
+	SeekAbsolute = 1 // Seek to an absolute offset.
+	SeekEnd      = 2 // Seek relative to the last offset available in the partition.
+	SeekCurrent  = 3 // Seek relative to the current offset.
+)
```

Then the names match up better with the `io.Seek*` constants, and I think the meaning is still pretty clear (maybe clearer), and shorter is always nice.

Any thoughts? I think whatever we decide, we're going to be stuck with it for a while. :)

achille-roussel:
Consistency with the `io` package sounds great to me 👍

PR #98, opened 2018-05-04
Fast forward offset when consuming too slow

zorkian:
If Kafka expires a log segment while we're in the middle of reading it, we would correctly detect that this had happened and update the Reader's offset, but the Connection would continue consuming from wherever it was. This adds a Seek call to fast forward to the first offset.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1637,6 +1637,8 @@ func (r *reader) run(ctx context.Context, offset int64) {
 					r.withErrorLogger(func(log *log.Logger) {
 						log.Printf("the kafka reader is reading before the first offset for partition %d of %s, skipping from offset %d to %d (%d messages)", r.partition, r.topic, offset, first, first-offset)
 					})
+					// seek connection to first offset
+					conn.Seek(0, 0)
 					offset, errcount = first, 0
 					continue // retry immediately so we don't keep falling behind due to the backoff
 


zorkian:
FWIW, this is a bug that currently causes the client to get stuck reading invalid offsets forever. The fix seems fairly obviously and straightforward, any chance of merging it?

achille-roussel:
Apologies, I missed this PR.

@stevevls @abraithwaite @Pryz @jnjackins any concerns about this fix?

stevevls:
To me, this feels like it should be wrapped with a configuration setting.  I agree that this is a useful change, but I'm not sure that I'd want to force this behavior on all consumers of the library because it may silently succeed where some would want it to fail.

reader.go
```diff
@@ -1637,6 +1637,8 @@ func (r *reader) run(ctx context.Context, offset int64) {
 					r.withErrorLogger(func(log *log.Logger) {
 						log.Printf("the kafka reader is reading before the first offset for partition %d of %s, skipping from offset %d to %d (%d messages)", r.partition, r.topic, offset, first, first-offset)
 					})
+					// seek connection to first offset
+					conn.Seek(0, 0)
```
jnjackins:
```suggestion
					conn.Seek(0, SeekStart)
```

reader.go
```diff
@@ -1637,6 +1637,8 @@ func (r *reader) run(ctx context.Context, offset int64) {
 					r.withErrorLogger(func(log *log.Logger) {
 						log.Printf("the kafka reader is reading before the first offset for partition %d of %s, skipping from offset %d to %d (%d messages)", r.partition, r.topic, offset, first, first-offset)
 					})
+					// seek connection to first offset
```
jnjackins:
```suggestion
```

jnjackins:
Can you explain why this seek is necessary? Line 1640 shows that the offset passed to `read` is already being updated to the first offset. Or better yet, can you add a test that is fixed by this change?

abraithwaite:
Generally, I'm against this feature.  I'm not opposed to adding it if there's enough support for it.

However, here's the problem: if a consumer is at the back of the queue, it generally means it's severely lagging, for whatever requirements have been set for the topic retention.  When this happens, it's probably likely to happen again.   When a reader is performing this badly, it also affects the rest of the consumers due to kafka's design (caching using the OS pagecache).  The slow consumer will be reading from disk into memory, taking up valuable space in the pagecache which should be reserved for the front-of-the-queue consumers.

Personally, I'd like to hear of a solid use-case for this before it gets implemented.

>  we correctly detect that this had happened and update the Reader's offset

Seeing as it's already (half?) implemented, perhaps it's worth doing.

@achille-roussel @rjenkins @stevevls Thoughts?

Pryz:
Fast forwarding is a really specific behavior. Like @abraithwaite explained if you have to reset the offset of a consumer this indicates something is wrong with the application or something bad happened.

I would not be against having this implemented as an option for the `Reader` but not the default behavior. By default, I will much prefer the application to fail/crash.

zorkian:
The current behavior is definitely a bug. When it consumes too slowly it ends up never consuming again as it continues to fetch the same invalid offset forever.

The approach taken by `kafka-python` and `librdkafka` is to provide a user configuration that lets you select what to do when the offset is out of range. Both libraries provide three options: beginning, end, or error. They both default to pick the largest/end offset and fast forward to that.

Which default is best depends on your POV and use case. I don't really have a horse in that race.

Please feel free to close the PR if you don't need it, but I might recommend filing an issue to at least return some kind of error in this situation so that users can see that something has gone wrong.

Pryz:
Definitively not against fast forwarding. I just thing the user should have a choice between fast-forwarding and fail.
As a first solution, we can make the behavior clear in the doc, merge this PR and fill an issue to add the option in the reader.

achille-roussel:
I think there are a few things that I'd like to see added to land this feature:

- We should add a counter to the reader stats that tracks the number of offsets that have been skipped, otherwise we would be potentially discarding data without knowing about it, which seems worse than the current behavior.

- Skipping to the first offset is still racy, the Kafka broker could have deleted the first offset by the time the reader attempts to read it. I feel like we should track those kinds of situations and seek to a position like `firstOffset + delta` to minimize the risks of a race condition by skipping further than the last offset. The delta could start at zero, and increase by "2 x (new first offset - previous first offset)" to try adapt to the rate at which the broker is discarding offsets.

- We definitely want to add a test for this, a bug here could have dramatic consequences. In comparison the current approach (even if it's annoying) seems safer because it requires human intervention to fix the broken state of the system.

- We probably want to make it a configurable option, so programs have the choice of using the feature or not.

achille-roussel:
I'm going to close this due to inactivity, but feel free to re-open if there's interest in getting this change merged again!

Issue #82, opened 2018-03-08
proposal: support for snappy


matthax:
I've done some digging through the docs, and while I've had better luck with this library than a few others but one thing I've noticed it seems to be lacking is support for snappy.

Most of the Kafka clusters I work with are using snappy by default for compression, so the ability to decompress messages coming from the reader and compress outgoing messages from the writer directly through this library would be great.

I'm just getting started with Go, so I'm not sure I'd be the best person to implement it, but I know the c headers are pretty readily available so it probably should not be terribly difficult...

Thoughts on this?

achille-roussel:
This is a very good point, and it would be a nice feature to add.

Would you have the cycles to work on this?

matthax:
I can give it a shot

errnoh:
I've been using github.com/eapache/go-xerial-snappy `Decode()` for decompressing `Message.Value` without running into any problems.

matthax:
@errnoh When I decode using that library I still seem to get some sort of header in the value. Any idea what might cause that or have you not experienced that problem?

matthax:
I noticed these messages also have no `Key` the `nbytes` being read for it is `-1`. Perhaps related? Not sure.

Pryz:
Hey @matthax sorry for the late response here. See my last comment here : https://github.com/segmentio/kafka-go/issues/100#issuecomment-406177864

achille-roussel:
@matthax snappy compression is fully supported in this package and we are using it in production. I'm going to close this issue, feel free to re-open if you have any issues getting snappy compression to work.

PR #106, opened 2018-06-27
Add compression

### LINKED ISSUE #87: Support for Gzip Compression
Related to #82 

I have a Kafka stream that I need to read from that is compressed with gzip. I would prefer to continue using this package as it's API is much nicer than the others and it doesn't require C Go. Would there be any interest in getting this added?

Pryz:
This PR is an attempt to add compression into `kafka-go`. I've only implemented `gzip` and `snappy` right now but I'm planning to add `lz4` as well.

I'm not quite sure if I'm doing to encoding/decoding at the right level. Let me know what you think.

Fixes #87 #82


message.go
```diff
@@ -2,9 +2,44 @@ package kafka
 
 import (
 	"bufio"
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"io/ioutil"
 	"time"
+
+	"github.com/eapache/go-xerial-snappy"
```
achille-roussel: Any reasons not to use [github.com/golang/snappy](https://godoc.org/github.com/golang/snappy) ?

Pryz: The Kafka encoding of snappy is using a specific Java library which does its own framing. If we really want to use `golang/snappy` I will have to dig into https://github.com/xerial/snappy-java

abraithwaite:
> We could also adopt a model like what the standard image package uses, where importing a kafka-go subpackage for compression automatically installs the codecs in the kafka-go package, so all you have to do to support compression on the consumer side is:

+1

Pryz:
Love the idea ! I will rework that PR :)

abraithwaite:
One note on that though: I'd opt for having the import register the codec, but still require specifying it when instantiating a client.  Similar to how the `sql` package requires both import and specifying the driver.

I'm not a fan of it being chosen automatically on import as some packages do.

compression.go
```diff
@@ -0,0 +1,40 @@
...
+	CompressionGZIP
+	CompressionSnappy
+	CompressionLZ4
+)
+
+type CompressionCodec struct {
+	str    func() string
+	encode func(src []byte, level int) ([]byte, error)
+	decode func(src []byte) ([]byte, error)
+}
```
achille-roussel: There's a weird dependency on the gzip compression level exposed in this abstraction. Neither lz4 nor snappy have such configuration.

How about something like this:
```go
package kafka

type CompressionCodec interface {
    Encode(dst, src []byte) (int, error)
    Decode(dst, src []byte) (int, error)
}

type Message struct {
    ...
    CompressionCodec CompressionCodec // no need to carry extra options if we carry the codec directly
}

func RegisterCompressionCodec(code int8, newCodec func() CompressionCodec) {
    ...
}

func NewCompressionCodec(code int8) CompressionCodec {
    // lookup the codec registered with `code`
    ...
}
```
```go
package gzip

type CompressionCodec struct {
    // use the zero-value, then initialize with the Reset method in the Encode/Decode
    r gzip.Reader
    w gzip.Writer
}

func NewCodec() *CompressionCodec {
    return NewCodec(gzip.DefaultCompression)
}

func NewCodecLevel(level int) *CompressionCodec {
    ...
}

const Code = 1 // gzip compression codec code in the kafka protocol

func init() {
    kafka.RegisterCompressionCodec(Code, func() kafka.CompressionCodec {
        return NewCodec()
    })
}
```
Here are a couple of reasons why I'm suggesting those changes:
- the compression level on gzip is only necessary on the encoder side, and the codec registry is mostly useful for decompression, so no reason to expose those implementation details in the generic interface
- using `Encode(dst, src []byte) (int, error)` instead of `Encode(src []byte) ([]byte, error)` means that the API gives an opportunity for optimizing the allocation of the destination buffer, which is otherwise not possible if the codec is in charge of allocating it.
- using the kafka codec ID instead of a codec name means we're not re-inventing yet another way to identify the compression codec, we're just using the values that are defined by the protocol.
- in each sub-package providing the implementation of a compression codec, the codec constructors return a concrete type instead of an interface, this is considered a best practice in Go (that's what the `compress/gzip` package does for example).

gzip/gzip.go
```diff
@@ -0,0 +1,83 @@
...
+
+// Code implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Code() int8 {
+	return 1
+}
+
+// Encode implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Encode(dst, src []byte) (int, error) {
+	var buf bytes.Buffer
+	writer := gzip.NewWriter(&buf)
```
achille-roussel: Creating a gzip writer is not a cheap operation, the constructor has to allocate multiple buffers internally, which is why I was suggesting you to keep the writer as a field of the compression codec, and use its `Reset` method to set the writer that it will be outputting to compressed data to.

gzip/gzip.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Code implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Code() int8 {
+	return 1
+}
+
+// Encode implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Encode(dst, src []byte) (int, error) {
+	var buf bytes.Buffer
```
achille-roussel: Why use an intermediary buffer instead of an instance of your `buffer` type here? It causes an unnecessary copy where you could instead have the gzip writer output directly to the destination byte slice.

gzip/gzip.go
```diff
@@ -0,0 +1,83 @@
...
+
+	n, err := buf.WriteTo(&buffer{
+		data: dst,
+	})
+	return int(n), err
+}
+
+// Decode implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Decode(dst, src []byte) (int, error) {
+	reader, err := gzip.NewReader(bytes.NewReader(src))
```
achille-roussel: Same comment as for the writer, reusing the reader and calling it's `Reset` method to configure its input would be much more efficient here.

writer.go
```diff
@@ -106,6 +106,12 @@ type WriterConfig struct {
 	// whether the messages were written to kafka.
 	Async bool
 
+	// CompressionCodec set the codec to be used to compress Kafka messages.
+	CompressionCodec int8
+
+	// CompressionLevel set the compression level if the codec supports it.
+	CompressionLevel int
```
achille-roussel: Is this still necessary? How about having a `CompressionCodec` field instead so the program can set the compression algorithm to whatever it pleases.

message.go
```diff
@@ -19,6 +21,12 @@ type Message struct {
 	// If not set at the creation, Time will be automatically set when
 	// writing the message.
 	Time time.Time
+
+	// Compression codec used to encode the message value
+	CompressionCodec int8
+
+	// Compression level for the codec if supported (only gzip)
+	CompressionLevel int
```
achille-roussel: Should those field be changed to a `CompressionCodec` instead? Having the compression level part of the message here is really mixing abstraction layers that should be decoupled. For example, of we needed to support new compression options, should they all be carried in the message as well?

compression.go
```diff
@@ -0,0 +1,71 @@
...
+const compressionCodecMask int8 = 0x03
+const DefaultCompressionLevel int = -1
+
+func init() {
+	RegisterCompressionCodec(func() CompressionCodec {
+		return CompressionCodecNone{}
+	})
+}
+
+type CompressionCodecNone struct{}
```
achille-roussel: Is this type necessary? Always applying Encode/Decode on messages that have no compression means we're going to double the number of dynamic memory allocations and require a copy of the message values that was not necessary before.

Having a way to represent "no compression" is important but we should just get away with not encoding/decoding in this case so we don't pay an extra performance penalty when no compression is applied.

compression.go
```diff
@@ -0,0 +1,71 @@
...
+	codecs[c.Code()] = c
+}
+
+// CompressionCodec represents a compression codec to encode and decode
+// the messages.
+// See : https://cwiki.apache.org/confluence/display/KAFKA/Compression
+type CompressionCodec interface {
+	Code() int8
+	Encode(dst, src []byte) (int, error)
+	Decode(dst, src []byte) (int, error)
```
achille-roussel: Those methods would deserve to be documented to define how valid compression codec implementations must behave. For example, explaining the error that must be returned if the destination buffer is too small, or what the value returned by the `Code` function should be.

compression.go
```diff
@@ -0,0 +1,71 @@
...
+	CompressionGZIP
+	CompressionSnappy
+	CompressionLZ4
+)
+
+var codecs map[int8]CompressionCodec
+
+// RegisterCompressionCodec registers a compression codec so it can be used by a Writer.
+func RegisterCompressionCodec(codec func() CompressionCodec) {
+	if codecs == nil {
```
achille-roussel: What's the point of this lazy-allocation here? Why not simply create the map when it's declared?
```go
var codecs = make(map[int8]CompressionCodec)
```

compression.go
```diff
@@ -0,0 +1,71 @@
+package kafka
+
+const (
+	CompressionNone int8 = iota
+	CompressionGZIP
+	CompressionSnappy
+	CompressionLZ4
+)
+
+var codecs map[int8]CompressionCodec
```
achille-roussel: It would be save to synchronize access to this map with a `sync.RWMutex` because we can't predict when a program will call `RegisterCompressionCodec`.

compression.go
```diff
@@ -0,0 +1,71 @@
+package kafka
+
+const (
+	CompressionNone int8 = iota
+	CompressionGZIP
+	CompressionSnappy
+	CompressionLZ4
```
achille-roussel: I had suggested you to move those constants in the packages that implement the compression codecs with a value like this:
```go
package gzip
// Having the code called the same in every package makes the code very programmer friendly
// as it becomes intuitive what it's gonna be named in each package. It also gives a clear
// guideline to future maintainers about how they should structure packages that provide
// compression codec implementations.
const Code = 1
```
Any reason why you'd prefer to keep those constants here?

snappy/snappy.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Code implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Code() int8 {
+	return 2
+}
+
+// Encode implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Encode(dst, src []byte) (int, error) {
+	buf := buffer{data: dst}
+	return buf.Write(snappy.Encode(src))
```
achille-roussel: Honestly the framing header is so simple in https://github.com/eapache/go-xerial-snappy/blob/master/snappy.go that I feel like we should just bring the code in and not take a dependency to the go-xerial-snappy package here.

This would allow you to use a similar implementation to the one you used for gzip and lz4 based on the snappy reader and writer in https://godoc.org/github.com/golang/snappy, which would be more efficient than forcing the allocation of a temporary buffer like what's currently done.

reader.go
```diff
@@ -1203,7 +1203,11 @@ func (r *Reader) FetchMessage(ctx context.Context) (Message, error) {
 					m.error = io.ErrUnexpectedEOF
 				}
 
-				return m.message, m.error
+				msg, err := m.message.Decode()
+				if err != nil {
+					return m.message, err
+				}
+				return msg, m.error
```
achille-roussel: I think the behavior of this code is being changed if an error occurred before decoding, it used to return `m.error` but now it will likely return the decoder's own error instead.

It seems to me that the correct approach would be to check for `m.error != nil` before decoding the message, and if non-nil, return the error without attempting to decode the message.

compression.go
achille-roussel: If the constant is in the snappy package it would be snappy.Code

Pryz:
Did a quick benchmark : 

```
$ benchcmp old.txt new.txt                                                                                                                                                                                                                          
benchmark                          old ns/op     new ns/op     delta                                                                                                                                                                                
BenchmarkConn/Seek-4               153827        153954        +0.08%                                                                                                                                                                               
BenchmarkConn/Read-4               210917        206446        -2.12%
BenchmarkConn/ReadBatch-4          24629         24945         +1.28%
BenchmarkConn/ReadOffsets-4        154487        152966        -0.98%
BenchmarkConn/Write-4              125715        127393        +1.33%
BenchmarkReader-4                  233107        237784        +2.01%
BenchmarkSyncGroupResponseV1-4     112           113           +0.89%

benchmark                     old MB/s     new MB/s     speedup
BenchmarkConn/Read-4          47.41        48.44        1.02x
BenchmarkConn/ReadBatch-4     401.99       396.90       0.99x                                                                                                                                                                                       
BenchmarkConn/Write-4         79.54        78.50        0.99x                                                                                                                                                                                       
BenchmarkReader-4             8.79         68.90        7.84x
```

Pryz:
Current status : 
```
$ go test -v -run=XXX -bench=BenchmarkCompression -test.benchmem
goos: linux
goarch: amd64
BenchmarkCompression/None-4              5000000               286 ns/op               0 B/op          0 allocs/op
BenchmarkCompression/GZIP-4                 3000            461296 ns/op         2484425 B/op         71 allocs/op
BenchmarkCompression/Snappy-4            2000000               616 ns/op             136 B/op          6 allocs/op
BenchmarkCompression/LZ4-4                  1000           1841563 ns/op        16977090 B/op         43 allocs/op
PASS
ok      github.com/segmentio/kafka-go   7.055s
```

Pryz:
Benchmark with multiple payload sizes : 

```
$ go test -v -run=XXX -bench=BenchmarkCompression -test.benchmem
goos: linux
goarch: amd64
BenchmarkCompression/None1024-4                  5000000               332 ns/op               0 B/op          0 allocs/op
BenchmarkCompression/None4096-4                  3000000               480 ns/op               0 B/op          0 allocs/op
BenchmarkCompression/None8192-4                  2000000               680 ns/op               0 B/op          0 allocs/op
BenchmarkCompression/None16384-4                 1000000              1068 ns/op               0 B/op          0 allocs/op
BenchmarkCompression/GZIP1024-4                    10000            190830 ns/op          904577 B/op         39 allocs/op
BenchmarkCompression/GZIP4096-4                     5000            327800 ns/op          939246 B/op         43 allocs/op
BenchmarkCompression/GZIP8192-4                     3000            519551 ns/op          986500 B/op         45 allocs/op
BenchmarkCompression/GZIP16384-4                    2000            911897 ns/op         1080995 B/op         48 allocs/op
BenchmarkCompression/Snappy1024-4                 500000              2439 ns/op            7809 B/op          6 allocs/op
BenchmarkCompression/Snappy4096-4                 200000              5942 ns/op           30981 B/op          6 allocs/op
BenchmarkCompression/Snappy8192-4                 200000             10466 ns/op           61706 B/op          6 allocs/op
BenchmarkCompression/Snappy16384-4                100000             19027 ns/op          124948 B/op          6 allocs/op
BenchmarkCompression/LZ41024-4                      1000           1338378 ns/op        12720875 B/op         35 allocs/op
BenchmarkCompression/LZ44096-4                      1000           1362444 ns/op        12733860 B/op         36 allocs/op
BenchmarkCompression/LZ48192-4                      1000           1397017 ns/op        12750790 B/op         36 allocs/op
BenchmarkCompression/LZ416384-4                     1000           1400856 ns/op        12784318 B/op         36 allocs/op
PASS
ok      github.com/segmentio/kafka-go   27.055s
```

Pryz:
With SetBytes on encoded messages : 
```
$ go test -v -run=XXX -bench=BenchmarkCompression -test.benchmem                                                                                                                                                   
goos: linux                                                                                                                                                                                                        
goarch: amd64
BenchmarkCompression/None1024-4                  5000000               332 ns/op        3077.37 MB/s           0 B/op          0 allocs/op                                                                         
BenchmarkCompression/None4096-4                  3000000               480 ns/op        8522.58 MB/s           0 B/op          0 allocs/op                                                                         
BenchmarkCompression/None8192-4                  2000000               678 ns/op        12066.28 MB/s          0 B/op          0 allocs/op                                                                         
BenchmarkCompression/None16384-4                 1000000              1065 ns/op        15371.09 MB/s          0 B/op          0 allocs/op                                                                         
BenchmarkCompression/GZIP1024-4                    10000            184983 ns/op           4.38 MB/s      904578 B/op         39 allocs/op                                                                         
BenchmarkCompression/GZIP4096-4                     3000            333677 ns/op           9.30 MB/s      939247 B/op         43 allocs/op                                                                         
BenchmarkCompression/GZIP8192-4                     3000            496374 ns/op          12.42 MB/s      986501 B/op         45 allocs/op                                                                         
BenchmarkCompression/GZIP16384-4                    2000            872557 ns/op          14.08 MB/s     1077026 B/op         48 allocs/op                                                                         
BenchmarkCompression/Snappy1024-4                 500000              2303 ns/op         446.74 MB/s        7809 B/op          6 allocs/op                                                                         
BenchmarkCompression/Snappy4096-4                 300000              5817 ns/op         704.95 MB/s       30981 B/op          6 allocs/op                                                                         
BenchmarkCompression/Snappy8192-4                 200000             10289 ns/op         796.67 MB/s       61706 B/op          6 allocs/op                                                                         
BenchmarkCompression/Snappy16384-4                100000             18335 ns/op         893.87 MB/s      124948 B/op          6 allocs/op                                                                         
BenchmarkCompression/LZ41024-4                      1000           1255731 ns/op           0.83 MB/s    12720863 B/op         35 allocs/op                                                                         
BenchmarkCompression/LZ44096-4                      1000           1279243 ns/op           3.22 MB/s    12733848 B/op         36 allocs/op                                                                         
BenchmarkCompression/LZ48192-4                      1000           1294811 ns/op           6.34 MB/s    12750780 B/op         36 allocs/op                                                                         
BenchmarkCompression/LZ416384-4                     1000           1318437 ns/op          12.44 MB/s    12784320 B/op         36 allocs/op                                                                         
PASS
ok      github.com/segmentio/kafka-go   26.216s
```

message.go
```diff
@@ -19,6 +21,9 @@ type Message struct {
 	// If not set at the creation, Time will be automatically set when
 	// writing the message.
 	Time time.Time
+
+	// Compression codec used to encode the message value
+	CompressionCodec int8
```
achille-roussel: What's your take on using a `CompressionCodec` here instead of the code?

message.go
```diff
@@ -32,15 +37,63 @@ func (msg Message) item() messageSetItem {
...
+		return msg, fmt.Errorf("codec %d not imported.", msg.CompressionCodec)
+	}
+
+	var err error
+	msg.Value, err = transform(msg.Value, codec.Encode)
+	return msg, err
+}
+
+// Decode decodes the Message using the CompressionCodec.
+func (msg Message) Decode() (Message, error) {
```
achille-roussel: Is there a use-case for exporting those methods?

message.go
Pryz: You mean the type CompressionCodec ?

achille-roussel: Yep, basically carrying an instance of `CompressionCodec` instead of the code.

compression.go
```diff
@@ -0,0 +1,31 @@
+package kafka
+
+var codecs = make(map[int8]CompressionCodec)
```
achille-roussel: You still need a mutex to protect reads and writes to this map.

message.go
Pryz: With the actual implementation it means that we need a type of the "None" codec case.

message.go
Pryz: mmm I don't have any. I will remove that.

Pryz: This is going to screw up my tests ! :D

message.go
achille-roussel: Can we use nil to mean there is no compression?

snappy/snappy.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Code implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Code() int8 {
+	return Code
+}
+
+func (c CompressionCodec) Encode(dst, src []byte) (int, error) {
+	buf := buffer{data: dst}
+	return buf.Write(snappy.Encode(nil, src))
```
nussjustin: Instead of using `buffer` and passing `nil` to `snappy.Encode` wouldn't it make more sense to just use `snappy.MaxEncodedLen` to check if the data fits? This way `dst` won't be changed if it's too small and if it fits you can just pass `dst` to `snappy.Encode`, avoiding an allocation.

E.g. 

```go
if n := snappy.MaxEncodedLen(len(src)); n > len(dst) {
    return 0, bytes.ErrTooLarge
}

return snappy.Encode(dst, src)
```

message.go
abraithwaite: > This is going to screw up my tests ! :D

You can un-export the method, then put an exported method in a file called `export_test.go` under the same package and your tests will still work.

snappy/snappy.go
achille-roussel: That's a good suggestion. I think we're gonna try to focus on an OK implementation in this PR, then we can think about ways to make it faster. I'll leave it up to @Pryz to make the call here tho since he's leading the effort.

message.go
```diff
@@ -31,16 +36,68 @@ func (msg Message) item() messageSetItem {
...
 	return m
 }
 
+func (msg Message) encode() (Message, error) {
+	//if msg.CompressionCodec.Code() == CompressionNoneCode {
+	if msg.CompressionCodec == nil {
+		return msg, nil
+	}
+
+	codec, ok := codecs[msg.CompressionCodec.Code()]
```
achille-roussel: Is this necessary? The message already carries a codec, we could use it directly instead of forcing a check on the registered codecs.

message.go
```diff
@@ -31,16 +36,68 @@ func (msg Message) item() messageSetItem {
...
+		Value:      msg.Value,
+		Timestamp:  timestamp(msg.Time),
+		Attributes: attrs,
 	}
 	m.CRC = m.crc32()
 	return m
 }
 
+func (msg Message) encode() (Message, error) {
+	//if msg.CompressionCodec.Code() == CompressionNoneCode {
```
achille-roussel: This comment should probably go away.

message.go
```diff
@@ -31,16 +36,68 @@ func (msg Message) item() messageSetItem {
...
+
+func (msg Message) decode() (Message, error) {
+	c := msg.message().Attributes & compressionCodecMask
+	if c == CompressionNoneCode {
+		return msg, nil
+	}
+
+	codec, ok := codecs[c]
+	if !ok {
+		return msg, fmt.Errorf("codec %d not imported.", msg.CompressionCodec)
```
achille-roussel: I wonder if an error is the right thing to do here. Maybe we just want to let the message go through in a compressed format?

The challenge is to detect whether the message is compressed or not at the application level then. We could introduce a "UnknownCompressionCodec" implementation of of the `CompressionCodec` interface that programs can test against, for example:
```go
if _, unknown := msg.CompressionCodec.(kafka.UnknownCompressionCodec); unknown {
   // maybe do something?
}
```

Let me know what you think? (I'm also OK if we make this part of a follow up PR)

writer.go
```diff
@@ -262,6 +265,8 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 		}
 
 		for _, msg := range msgs {
+			msg.CompressionCodec = w.config.CompressionCodec
```
achille-roussel: Should we check for `msg.CompressionCodec == nil` here? That way we allow individual messages to overwrite the compression codec they use, and otherwise default to the one configured on the writer.

Pryz: Does it mean we are going to allow multiple codec in the same topic ? Feels weird even so Kafka should be able to handle that.

Pryz:
Latest benchmarck : 
```
$ go test -v -run=XXX -bench=BenchmarkCompression -test.benchmem
goos: linux
goarch: amd64
BenchmarkCompression/None1024-4                  3000000               403 ns/op        2538.81 MB/s           0 B/op          0 allocs/op
BenchmarkCompression/None4096-4                  3000000               551 ns/op        7430.57 MB/s           0 B/op          0 allocs/op
BenchmarkCompression/None8192-4                  2000000               765 ns/op        10697.87 MB/s          0 B/op          0 allocs/op
BenchmarkCompression/None16384-4                 1000000              1149 ns/op        14252.97 MB/s          0 B/op          0 allocs/op
BenchmarkCompression/GZIP1024-4                    10000            236904 ns/op           3.42 MB/s      904575 B/op         39 allocs/op
BenchmarkCompression/GZIP4096-4                     5000            372801 ns/op           8.32 MB/s      939247 B/op         43 allocs/op
BenchmarkCompression/GZIP8192-4                     3000            548940 ns/op          11.23 MB/s      986501 B/op         45 allocs/op
BenchmarkCompression/GZIP16384-4                    2000            924999 ns/op          13.29 MB/s     1080995 B/op         48 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/Snappy1024-4                1000000              2630 ns/op         391.12 MB/s        6529 B/op          5 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/Snappy4096-4                 200000              6317 ns/op         649.10 MB/s       26116 B/op          5 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/Snappy8192-4                 100000             10706 ns/op         765.64 MB/s       51976 B/op          5 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/Snappy16384-4                100000             21057 ns/op         778.36 MB/s      104464 B/op          5 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/LZ41024-4                      1000           1715081 ns/op           0.61 MB/s    12720789 B/op         35 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/LZ44096-4                      1000           1450152 ns/op           2.84 MB/s    12733794 B/op         35 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/LZ48192-4                      1000           1573210 ns/op           5.22 MB/s    12750553 B/op         35 allocs/op                                                                                                                                                                                                                
BenchmarkCompression/LZ416384-4                     1000           1723310 ns/op           9.52 MB/s    12784119 B/op         35 allocs/op                                                                                                                                                                                                                
PASS                                                                                                                                                                                                                                                                                                                                                      
ok      github.com/segmentio/kafka-go   29.977s
```

abraithwaite:
Boom

PR #107, opened 2018-07-06
fixed bug with unknown topic/partition

bo0rsh201:
If you try to connect to unexistent partition/topic, you will get stuck into infinite retry loop with
ReadPartitions result [], nil
seems like error should be on other place

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -159,9 +159,9 @@ func (d *Dialer) LookupLeader(ctx context.Context, network string, address strin
 					return
 				}
 			}
+			errch <- UnknownTopicOrPartition
+			return
 		}
-
-		errch <- UnknownTopicOrPartition
 	}()
 
 	var brk Broker


abraithwaite:
I believe this is working as intended.  If the topic gets created after some time, it should start working.

If you think otherwise, can you please elaborate on why this is not desired behavior and reopen?   Thanks.

bo0rsh201:
@abraithwaite and how can I figure out that I should create something?
There is no message or error, program just stucks in DialLeader and doesn't cancel even on context.
This peace of code looks really strange
`go func() {
		for attempt := 0; true; attempt++ {
			if attempt != 0 {
                                //  maybe we should check sleep result and exit on false???
				sleep(ctx, backoff(attempt, 100*time.Millisecond, 10*time.Second))
			}

			partitions, err := c.ReadPartitions(topic)
			if err != nil {
				if isTemporary(err) {
					continue
				}
				errch <- err
				return
			}

			for _, p := range partitions {
				if p.ID == partition {
					brkch <- p.Leader
					return
				}
			}
		}
                //  it's complely unreachable
		errch <- UnknownTopicOrPartition
	}()
`
I have put few comments above.
I think - really expected behaviour is to explicitly return error if we don't have topic or partition (even worse)
and return control to parent code. If someone decides that it's ok to stuck forever silently waiting for topic to be created without any notifications - keep it outside of library.

abraithwaite:
I can see either behavior being valid.  Perhaps @Pryz, or @achille-roussel can chime in on what they think the behavior should be.

abraithwaite:
> Perhaps we can create topic on demand

This is a server-side config.  `auto.create.topics.enable`.  Perhaps this satisfies your use case?

bo0rsh201:
Current behavior seems strange for me, but if it breaks BC and someone relies on it (I hope not), I will just keep it in my own fork, but maybe we should at least cancel this loop on ctx cancel (sleep(...) result is ignored now)

achille-roussel:
How about adding a flag on the dialer to control yeh behavior? Something like “FailOnMissingTopic”, or maybe a “Strict” mode to force the dialer to fail faster?

bo0rsh201:
@achille-roussel thanks for feedback, but it seems like an overkill for this case :)
if it's not a common problem (nobody except me cares), I prefer to keep it simple.

PR #112, opened 2018-07-19
Read attributes from message

Pryz:
Fix the missing attributes while consuming compressed payload.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -107,7 +107,7 @@ func (batch *Batch) Read(b []byte) (int, error) {
 	batch.mutex.Lock()
 	offset := batch.offset
 
-	_, _, err := batch.readMessage(
+	_, _, _, err := batch.readMessage(
 		func(r *bufio.Reader, size int, nbytes int) (int, error) {
 			if nbytes < 0 {
 				return size, nil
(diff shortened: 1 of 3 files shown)


batch.go
```diff
@@ -107,7 +107,7 @@ func (batch *Batch) Read(b []byte) (int, error) {
 	batch.mutex.Lock()
 	offset := batch.offset
 
-	_, _, err := batch.readMessage(
+	_, _, _, err := batch.readMessage(
```
achille-roussel: Is it OK to discard the attributes here? It seems like we should decompress the content that has been written to the output buffer passed to `Read([]byte)` isn't it?

message.go
```diff
@@ -22,6 +22,9 @@ type Message struct {
 	// writing the message.
 	Time time.Time
 
+	// Message attributes.
+	Attributes int8
```
achille-roussel: Why do we need this? Isn't the `CompressionCodec` enough?

Pryz: We need it when we decompress the message we've just consumed. `CompressionCodec` is populated when we create a message to be produced.

batch.go
Pryz: Yes but I think we will need to implement Message Set for the decompression to really make sense.

message.go
achille-roussel: Decoding is done here right? https://github.com/segmentio/kafka-go/blob/c087815a5658518dd07e2bab939de0db8921adea/batch.go#L173

Seems like we could pass the attributes to `decode` instead of exposing an implementation detail in the public API of `Message`.

Pryz: We are also calling the `decode` method in the `Reader` where we don't have access to the attributes (I think, I will double check).

achille-roussel:
@Pryz @stevevls Does #135 supersets this PR?

Issue #120, opened 2018-09-26
-1 and -2 are used inconsistently to mean "first offset" or "last offset"


jnjackins:
See:
```
 // ReadFirstOffset returns the first offset available on the connection.
 func (c *Conn) ReadFirstOffset() (int64, error) {
       return c.readOffset(-2)
 }
```

And:
```
const (
       firstOffset = -1
       lastOffset  = -2
)
```

And:
```
// SetOffset changes the offset from which the next batch of messages will be
// read.
//
// Setting the offset ot -1 means to seek to the first offset.
// Setting the offset to -2 means to seek to the last offset.
```

And sometimes `-1` means something else entirely:
```
// Offset returns the current offset of the reader.
func (r *Reader) Offset() int64 {
	if r.useConsumerGroup() {
		return -1
	}
```

jnjackins:
Also, -1 and -2 are hardcoded all over the place, and the intent is almost never clear: do they mean first offset or last offset? According to our private constants, they mean one thing, and according to the Kafka protocol, they mean the opposite thing. Kafka claims that -2 is the first (oldest) offset, and -1 is the last (newest) offset.

jnjackins:
Whatever the correct values are, these constants need to be exported and used everywhere. We had the same issue with seek constants, too: the values were used inconsistently and incorrectly. Named constants makes it clear.

jnjackins:
#121 illustrates the problem in more detail.

PR #121, opened 2018-09-26
Use named constants for special offsets

jnjackins:
Note: This PR breaks backward compatibility.

This change replaces hardcoded -1 and -2 offsets with named constants, and fixes them to be consistent with other libraries / kafka docs.

For #120.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1368,10 +1368,11 @@ func (r *Reader) ReadLag(ctx context.Context) (lag int64, err error) {
 	return
 }
 
-// Offset returns the current offset of the reader.
+// Offset returns the current absolute offset of the reader, or -1
+// if r is backed by a consumer group
 func (r *Reader) Offset() int64 {
 	if r.useConsumerGroup() {
-		return 0
+		return -1
 	}
 
 	r.mutex.Lock()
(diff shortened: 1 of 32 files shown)


reader.go
```diff
@@ -17,8 +17,8 @@ import (
 )
 
 const (
```
jnjackins: I reversed these. Sarama and kafkacat both treat them this way (last == -1 and first == -2). Neither seem to be mentioned in the Kafka protocol, so I'm unclear whether this is a library convention or a fundamental Kafka thing.

jnjackins:
More data: the kafka python library agrees with sarama:
```
class OffsetResetStrategy(object):
    LATEST = -1
    EARLIEST = -2
    NONE = 0
```

reader.go
abraithwaite: FWIW, I prefer 'Newest' and 'Oldest', but being in line with other libraries is probably the most important.

achille-roussel:
@jnjackins would you have time to fix the tests? Or maybe someone else can take a crack at it?

jnjackins:
I'll look at it again today and let you know if I need help, how does that sound. :)

PR #122, opened 2018-09-26
Reader nows allows setting offset with timestamp

kngu9:
Previously the library allowed to set the offset given the offset number. However the reader does not allow setting the offset by finding the next available offset by timestamp.

This change will allow you to change the reader's offset by specifying the timestamp.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1388,6 +1388,32 @@ func (r *Reader) SetOffset(offset int64) error {
 	return err
 }
 
+// SetOffsetWithTimestamp changes the offset from which the next batch of messages will be
+// read given the timestamp t.
+//
+// The method fails if the unable to connect to the coordinator, or unable to read the offset
+// given the ts, or if the reader has been closed.
+func (r *Reader) SetOffsetWithTimestamp(t time.Time) error {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if r.closed {
+		return io.ErrClosedPipe
+	}
+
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -1388,6 +1388,32 @@ func (r *Reader) SetOffset(offset int64) error {
...
+// given the ts, or if the reader has been closed.
+func (r *Reader) SetOffsetWithTimestamp(t time.Time) error {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if r.closed {
+		return io.ErrClosedPipe
+	}
+
+	conn, err := r.coordinator()
```
achille-roussel: I'm confused about why we're looking for the coordinator here. The coordinator is supposed to be used when the reader is part of a consumer group, but after we're callign `SetOffset` which is a method that cannot be used in a consumer group.

Can you describe a bit more what your intent was here?

achille-roussel:
@kngu9 do you still have interest in getting this PR merged?

kngu9:
> @kngu9 do you still have interest in getting this PR merged?

Hi @achille-roussel, it seems to me that the last message that you sent did not get to me. I apologize for this. My intentions were that, given a timestamp 2019-01-15T09:00:00.00, I want to get the offset from that timestamp, however I can't seem to be able to set it with the current functions that the library implemented.

achille-roussel:
I think that what you're looking for is the leader of the partition, not the coordinator isn't it?

PR #124, opened 2018-10-15
Exported Consumer Group Strategy and made it configurable

stevevls:


strategy.go
```diff
@@ -2,8 +2,8 @@ package kafka
 
 import "sort"
 
-// strategy encapsulates the client side rebalancing logic
-type strategy interface {
+// Strategy encapsulates the client side rebalancing logic
+type Strategy interface {
```
achille-roussel: I wonder if we should rename this type now that it's going to be part of the public API. "Strategy" is a bit vague.

What do you think about something like `ConsumerStrategy` / `GroupStrategy` / `GroupBalancer` / etc... ?

stevevls: Makes sense.  `Balancer` is already used for partitioning, so might not want to overload it even though `GroupBalancer` is otherwise a good choice.  What do you think of `AssignmentStrategy`?

achille-roussel: I think "Assignment" isn't used anywhere in the API right now, so it's makes it harder to intuitively figure out what it is. I was thinking "Group" because it reuses the wording of "GroupID" used in the reader configuration.

stevevls: Okay.  I'd propose `GroupBalancer` then unless you think that causes confusion with the `Balancer` for partitioning.

stevevls:
Oops...I had renamed it to `group_balancer`.  Will remove the extra underscore.

Pryz:
We should fix the CI before pushing more changes.

groupbalancer.go
```diff
@@ -93,31 +98,28 @@ func (r rangeStrategy) AssignGroups(members []memberGroupMetadata, topicPartitio
 // 		C1: [1, 4]
 // 		C2: [2, 5]
 //
-type roundrobinStrategy struct{}
+type RoundRobinGroupBalancer struct{}
```
stevevls: @achille-roussel I was thinking about maybe leaving this type unexported and having a constant instead:
```suggestion
var RoundRobin GroupBalancer = roundRobinGroupBalancer{}
```

Does that make for a clearer API for the client, or do you think it's better to have the type exported?

Pryz:
Thanks for fixing the tests !

achille-roussel:
@stevevls regarding RoundRobinGroupBalancer I think having an exported type is better than a variable in general, here are a couple of reasons:

- If/when we need to extend it, adding properties to a type can be done without breaking backward compatibility. If it was a global variable we'll have to break the code to change it to a type, or introduce new APIs.

- A global variable is not a constant, it may be overwritten (hopefully by mistake but still), so a program could assign `kafka.RoundRobin` to something that's not actually a round-robin balancing strategy. Types cannot be overwritten.

- A global variable makes discoverability through the Go documentation harder than an exported type, which will appear that the top level.

If you're concerned about increasing the amount of exported symbols that show at the top-level of the documentation, an approach I've taken before is to have a constructor returning the interface type. Go doc will group those under the documentation of the interface type, which makes the doc easy to navigate.

groupbalancer.go
```diff
@@ -2,33 +2,41 @@ package kafka
 
 import "sort"
 
-// strategy encapsulates the client side rebalancing logic
-type strategy interface {
-	// ProtocolName of strategy
+// GroupMember describes a single participant in a consumer group.
+type GroupMember struct {
+	// ID is the unique ID for this member as taken from the JoinGroup response.
+	ID       string
```
achille-roussel: Are you running an old-school version of `go fmt`? Or maybe this is a new formatting rule? The spacing here is unusual.

groupbalancer.go
```diff
@@ -2,33 +2,41 @@ package kafka
...
+	// Topics is a list of topics that this member is consuming.
+	Topics   []string
+
+	// UserData contains any information that the GroupBalancer sent to the
+	// consumer group coordinator.
+	UserData []byte
+}
+
+// MemberGroupAssignments holds MemberID => topic => partitions
+type MemberGroupAssignments map[string]map[string][]int
```
achille-roussel: What's your take here on "MemberGroup" vs "GroupMember"?

groupbalancer.go
stevevls: My on-save action got unhooked, so `go fmt` wasn't running.  Good catch!

groupbalancer.go
stevevls: I like it.  Changed!

stevevls:
> Thanks for fixing the tests !

One of them is still racy. 😿

stevevls:
@achille-roussel Thanks for the explanation on type v. variable...that's super helpful.

conn_test.go
```diff
@@ -58,12 +58,10 @@ func (c *connPipe) Write(b []byte) (int, error) {
 	// goroutines a chance to start and set the deadline.
 	time.Sleep(time.Millisecond)
 
-	// Some tests set very short deadlines which end up aborting requests and
-	// closing the connection. To prevent this from happening we check how far
-	// the deadline is and if it's too close we timeout.  The 100 ms value
-	// is chosen to coincide with the timeout set in nettest/TestFutureTimeout.
-	// That value is subject to change in the future.
-	if t := c.wconn.writeDeadline(); !t.IsZero() && t.Sub(time.Now()) <= (100*time.Millisecond) {
+	// The nettest code only sets deadlines when it expects the write to time
```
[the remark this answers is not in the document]
achille-roussel: It actually looks less hacky than what I had put in place, looking good to me 👍

PR #135, opened 2018-10-31
Added support for message set compression

stevevls:
This PR adds support for message set compression...a Kafka feature where batches of messages can be compressed together and produced as a single message.  This changes allows kafka-go to interoperate with compression as implemented by other APIs.

While there is already code in place for message-by-message compression, it doesn't quite work.  On the produce side, the `decode` function operates on a value instead of a pointer (https://github.com/segmentio/kafka-go/blob/c6db9435477f3cb658e2dd0fa93e02118c870251/message.go#L55), so the compressed value is lost.  Furthermore, the produce request uses an optimized function to write, and it does not set the compression attributes (https://github.com/segmentio/kafka-go/blob/c6db9435477f3cb658e2dd0fa93e02118c870251/write.go#L225).  As a result, kafka-go *always* writes uncompressed messages.  On the consume side, it does not correctly deserialize compressed message sets.

In order to facilitate this change, I upgraded fetch and produce requests from v1 to v2 so that we would get support for the magic byte value of 1 in the message set (see https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets).  These API versions are compatible with Kafka 0.10.0.0+.

Some areas where feedback would be appreciated:
1. I broke compatibility with the `CompressionCodec` interface to better fit the needs of the read/write code.  Since adding new `CompressionCodec` implementations would require hacking Kafka as well, I'm not terribly concerned that users have implemented their own.  LMK if you think this is not a valid assumption!
1. I am tempted to remove the `CompressionCodec` field from the `Message` struct even though that would be a breaking change.  I currently have marked it as deprecated.


Pryz:
> I broke compatibility with the CompressionCodec interface to better fit the needs of the read/write code. Since adding new CompressionCodec implementations would require hacking Kafka as well, I'm not terribly concerned that users have implemented their own. LMK if you think this is not a valid assumption!

Seems fine to me.

> I am tempted to remove the CompressionCodec field from the Message struct even though that would be a breaking change. I currently have marked it as deprecated.

Why ?

stevevls:
In order to get better compression ratios, I moved the compression up a level.  Now the `Writer` passes a `CompressionCodec` to the `Conn.WriteMessagesCompressed` function and the entire slice of messages gets compressed together as a single message set.  I think it makes sense to have the compression configured at the `Writer` level anyway, since I don't see a case where you would want to mix and match compression algorithms or send some messages compressed but others not.  What do you think?

message.go
```diff
@@ -23,6 +22,9 @@ type Message struct {
 	Time time.Time
 
 	// Compression codec used to encode the message value
+	//
+	// Deprecated: this field is ignored.  If compression is desired, then
+	// configure it on the Writer.
```
achille-roussel: I wonder if we should just remove the field. We'd be breaking programs at compile time, which is sad but could be better than silently changing the behavior at runtime?

write.go
```diff
@@ -112,10 +113,10 @@ func write(w *bufio.Writer, a interface{}) {
 // allocations that occur when building the data structures representing the
 // kafka protocol requests.
 
-func writeFetchRequestV1(w *bufio.Writer, correlationID int32, clientID, topic string, partition int32, offset int64, minBytes, maxBytes int, maxWait time.Duration) error {
+func writeFetchRequestV2(w *bufio.Writer, correlationID int32, clientID, topic string, partition int32, offset int64, minBytes, maxBytes int, maxWait time.Duration) error {
```
achille-roussel: Do you know which version of Kafka we're dropping backward compatibility with by moving to FetchRequest v2?

write.go
```diff
@@ -221,28 +233,47 @@ func writeProduceRequestV2(w *bufio.Writer, correlationID int32, clientID, topic
...
-		writeBytes(w, msg.Key)
-		writeBytes(w, msg.Value)
+		writeMessage(w, msg.Offset, attributes, msg.Time, msg.Key, msg.Value)
 	}
 
 	return w.Flush()
 }
+
+func compress(codec CompressionCodec, msgs ...Message) ([]Message, error) {
+	buf := &bytes.Buffer{}
```
achille-roussel: How about calling `buf.Grow` with an estimate of the final size here? It would save lots of copies and reallocations of the internal buffer.

conn.go
```diff
@@ -775,13 +774,22 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
...
 func (c *Conn) WriteMessages(msgs ...Message) (int, error) {
+	return c.WriteMessagesCompressed(nil, msgs...)
+}
+
+// WriteMessages writes a batch of messages to the connection's topic and
+// partition, returning the number of bytes written. The write is an atomic
+// operation, it either fully succeeds or fails.
+//
+// If the compression codec is not nil, the messages will be compressed.
+func (c *Conn) WriteMessagesCompressed(codec CompressionCodec, msgs ...Message) (int, error) {
```
achille-roussel: How about `WriteCompressedMessages`? So it reads like normal english :)

I don't have a strong preference, I like the common prefix too...

compression.go
```diff
@@ -1,6 +1,22 @@
...
+)
+
+type Compressor interface {
+	io.Writer
+	io.Closer
+}
+
+type Decompressor interface {
+	io.Reader
+}
```
achille-roussel: Any reason not to use `io.WriteCloser` and `io.Reader` here?

compression.go
```diff
@@ -1,6 +1,22 @@
 package kafka
 
-import "sync"
+import (
+	"io"
+	"sync"
+
+	"github.com/pkg/errors"
```
achille-roussel: I don't think we have a dependency on this package right now, and it doesn't look like we're using much of the features it provides.

Any reasons not to use the standard `errors` package instead?

batch.go
```diff
@@ -180,47 +178,36 @@ func (batch *Batch) readMessage(
...
+		err = batch.msgs.discard()
+		switch {
+		case err != nil:
+			batch.err = err
+		case batch.msgs.remaining() == 0:
+			// Because we use the adjusted deadline we could end up returning
+			// before the actual deadline occurred. This is necessary otherwise
+			// timing out the connection for real could end up leaving it in an
+			// unpredictable state, which would require closing it.
+			// This design decision was main to maximize the changes of keeping
```
achille-roussel: > This design decision was main to maximize the changes of keeping

- "main" looks like a typo maybe?
- "changes" should be "chances"?

conn.go
stevevls: That sounds good...I like normal English now that I think about it a little more.  The `Write` prefix should be plenty to indicate the functions are logically grouped.

message.go
stevevls: Removed!

PR #137, opened 2018-11-02
feat(writer): add logger config, use to log write errors

dominicbarnes:
Currently, when `WriterConfig.Async == true` you cannot capture write errors in order to diagnose problems. The only way to detect if errors are happening is via metrics, but obviously that is only an aggregate count. (since no additional tags are included)

This change adds `WriterConfig.Logger` and `WriterConfig.ErrorLogger` (mirroring `ReaderConfig`) to add support for logging within `Writer` in order to log write errors.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log"
 	"math/rand"
 	"sort"
 	"sync"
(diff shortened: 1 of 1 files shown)


AlexStocks:
I can not agree with use log to record error if I use other logger which do not compatible with the the go logger. In my opinion, the sarama's callback function style may be a better choice.

achille-roussel:
Maybe, maybe not. Logging is a tough problem. The design decision we made was to use the standard library to avoid inventing yet another way of doing logging. 

In my opinion, it is also easy enough to write a parser for the standard log format and extract the data, that introducing more abstractions isn’t necessary.

AlexStocks:
Yeah, I agree with u. However, I insist on the callback function which is more flexible than log.

Issue #138, opened 2018-11-02
Supply an rdkafka-compatible hash function too


samv:
This is relatively trivial, but for folk who are also using eg confluent_kafka or other librdkafka-based libraries in other places, they might appreciate a quick way to select this.

This code uses the test cases from `github.com/edenhill/librdkafka/tests/0048-partitioner.c` to confirm that this library/polynomial pair produces the same CRC32's.

The Java world uses a different partitioning function again, [murmur](https://en.wikipedia.org/wiki/MurmurHash).  Fun.  And none of are [consistent hashes](https://en.wikipedia.org/wiki/Consistent_hashing). 

```go
package main

import (
	"hash/crc32"
	"log"
)

var hashTable = crc32.MakeTable(0xEDB88320)

func main() {
	cases := []string{
		"this is another string with more length to it perhaps",
		"hejsan",
		"23456",
	}
	expected := []uint32{
		0xb0150df7,
		0xd077037e,
		0xb1b451d7,
	}

	for testCaseIdx := range cases {
		hasher := crc32.New(hashTable)
		hasher.Write([]byte(cases[testCaseIdx]))
		if hasher.Sum32() != expected[testCaseIdx] {
			log.Printf("mismatch on %s!  %x != %x", cases[testCaseIdx], hasher.Sum32(), expected[testCaseIdx])
		} else {
			
			log.Printf("%#v hashes to %x, good", cases[testCaseIdx], hasher.Sum32())
		}
	}
}
```

stevevls:
Fixed by #334

PR #142, opened 2018-11-11
Add CRC32 and Murmur2 balancers

martinxsliu:
This PR adds the `CRC32Balancer` and `Murmur2Balancer` types that provides the same partitioning behaviour as the librdkafka and Java clients respectively.

The default partitioner strategy for librdkafka is the `consistent_random` strategy which uses CRC32 if the length of the key is not zero, otherwise it randomly selects a partition ([see here](https://github.com/edenhill/librdkafka/blob/95a56e160a3da2018e694dbd74d315163afe90b9/src/rdkafka_msg.c#L706-L725)). This is implemented by the `CRC32Balancer`.

The default partitioner strategy for the Java client is to first use the provided partition of the produced message if it exists. Otherwise, if the key is null then use a round robin strategy, and if the key is provided then use its murmur2 algorithm to calculate the partition. Using a user provided partition value doesn't seem permitted by this library (going by the warning in the readme) so it was not implemented, but the round robin and murmur2 hashing is implemented by the `Murmur2Balancer`.

I have taken unit tests from the librdkafka and [kafka-python](https://github.com/dpkp/kafka-python) libraries to verify that the implementations are consistent.

### DIFF
--- a/balancer.go
+++ b/balancer.go
@@ -2,7 +2,9 @@ package kafka
 
 import (
 	"hash"
+	"hash/crc32"
 	"hash/fnv"
+	"math/rand"
 	"sort"
 	"sync"
 )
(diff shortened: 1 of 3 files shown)


martinxsliu:
Addresses https://github.com/segmentio/kafka-go/issues/138

balancer.go
```diff
@@ -158,3 +160,98 @@ func (h *Hash) Balance(msg Message, partitions ...int) (partition int) {
...
+
+	idx := (murmur2(msg.Key) & 0x7fffffff) % uint32(len(partitions))
+	return partitions[idx]
+}
+
+// Go port of the Java library's murmur2 function.
+// https://github.com/apache/kafka/blob/1.0/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L353
+func murmur2(data []byte) uint32 {
+	length := len(data)
+	var seed uint32 = 0x9747b28c
```
achille-roussel: Should this be a `const`? It doesn't seem to be modified anywhere.

balancer.go
```diff
@@ -158,3 +160,98 @@ func (h *Hash) Balance(msg Message, partitions ...int) (partition int) {
...
+	// 'm' and 'r' are mixing constants generated offline.
+	// They're not really 'magic', they just happen to work well.
+	var m uint32 = 0x5bd1e995
+	var r uint32 = 24
```
achille-roussel: Same for `m` and `r`, they could be constants if they aren't modified anywhere.

achille-roussel:
The changes look good to me 👍 

@Pryz @stevevls @yields do you have any concerns? Otherwise I think this can be merged.

balancer.go
```diff
@@ -158,3 +160,98 @@ func (h *Hash) Balance(msg Message, partitions ...int) (partition int) {
...
+//
+// This balancer strategy is the default partitioner used by the librdkafka library
+// and the language bindings that are built on top of it, including the
+// github.com/confluentinc/confluent-kafka-go Go package.
+type CRC32Balancer struct {
+	rb randomBalancer
+}
+
+func (b CRC32Balancer) Balance(msg Message, partitions ...int) (partition int) {
+	if len(msg.Key) == 0 {
```
stevevls: Just to double check, does the librdkafka library differentiate between a key that is the empty string and a nil key?  A quick comment here to clarify would be appreciated.

achille-roussel:
@martinxsliu Do you want to follow up on the last comments? Then we can merge this PR.

Pryz:
@martinxsliu are you still working on this PR ? If so can you followup on @stevevls comment and rebase your change please. Thanks :)

balancer.go
stevevls: Looks like this matches the default `consistent_random` partitioner (empty or nil key is randomly assigned).  There's also a `consistent` algo that does assigns empty or nil keys to a single partition.  It might make sense to decouple the `randomBalancer` so that we could support both modes.

See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

balancer.go
```diff
@@ -158,3 +160,98 @@ func (h *Hash) Balance(msg Message, partitions ...int) (partition int) {
...
+}
+
+// Murmur2Balancer is a Balancer that uses the Murmur2 hash function to determine
+// which partition to route messages to.  This ensures that messages with the same
+// key are routed to the same partition.  A round robin strategy will be used if
+// the key is nil.
+//
+// This balancer strategy is the default partitioner used by the Java library.
+type Murmur2Balancer struct {
+	rr RoundRobin
```
stevevls: Similarly, this balancer comes in two flavors, `murmur2` and `murmur2_random`.  Might make sense to decouple the RR.

See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

stevevls:
Fixed via #334

PR #143, opened 2018-11-15
Updated discardN to return errShortRead if N bytes not available

stevevls:


stevevls:
This addresses a lag issue that we found at production scale.  If a consumer receives truncated messages from the broker due to hitting max fetch request size, then if unlucky enough for the message to be truncated at the stage where we call `discardN` to dump the CRC and message size, it could end up blocking while attempting to read more bytes.  Eventually the request would time out, and it would proceed to the next fetch request.

While waiting to time out, the backlog continues to grow, which exacerbates the underlying issue, and consumers would likely make a small bit of progress after each timeout only to hit the `discardN` issue, timeout, have a larger backlog, fall further behind, ad nauseam.

This fix addresses that issue by doing a check against the remaining bytes.  If not enough bytes are available, we return immediately with an `errShortRead`.

discard.go
```diff
@@ -3,6 +3,9 @@ package kafka
 import "bufio"
 
 func discardN(r *bufio.Reader, sz int, n int) (int, error) {
+	if n > sz {
```
achille-roussel: Should we still do `r.Discard(sz)` here?

stevevls:
I'm giving up on the unit test for now...I realized I was running locally against kafka 0.10.1.0 and not 0.11.0.0 like in our unit tests.  Sadly, the test doesn't work correctly against 0.11.0.0.  I've left the code for reference w/ a todo to adapt it later.  Tests will skip it for now. 😞

achille-roussel:
@stevevls seems like the test is timing out, is it possible that 0.10 and 0.11 behave slightly differently here regarding priorities of MaxBytes vs MaxWait?

If MaxWait takes precedence then 500 iterations with a max wait of 100ms will exceed the 30s time limit on the test.

If MaxBytes takes precedence the responses will come almost instantly as soon as 100 bytes are available, and the test will likely complete with the 30s timeout.

I wonder if differences between 0.10 and 0.11 are also the reason why we failed to trigger the issue in our staging environment.

achille-roussel:
(I'm not super sure about my last comment, but it wouldn't be the first time we see those kinds of behavior changes between kafka versions)

stevevls:
I ran into a couple of issues with 0.11.  The first was that if we publish a big batch of messages, Kafka was returning them with magic byte equal to 2 (i.e. as a `RecordBatch` instead of a `MessageSet`).  This would cause a test timeout b/c reading messages would fail due to unsupported version.  Oddly enough, breaking down the initial produce into groups of 10 or fewer avoided this server-side behavior...they would come back as message sets.  However, once I did that, I was having trouble getting the truncation to occur.  I can spend a few more cycles attempting to reproduce today.  

Another thought I had was simply to add some unit tests to `discardN` to ensure that it's not reading too many bytes.  Given that it was the root cause of the problem, we could do some relatively simple tests there to ensure correct behavior.  What do you think of that instead?

I also had the same thought re: ability to reproduce in stage...I think there's something to it.

PR #145, opened 2018-11-19
Split compressed when needed to keep msg size under broker limit

stevevls:


stevevls:
What do you guys think about trying to make the max message bytes a configurable value?  I didn't see an easy way to do it with introducing API b/w incompatibility or adding a new function to `Conn`.

achille-roussel:
Introducing a new method on `*kafka.Conn` seems OK to me, what concerns do you have about it?

stevevls:
Actually, I think I'm going to abandon this.  The code still fails in cases where we exceed the 1MB limit on the produce request.  After further reflection, I believe that the burden is on the higher-level API to ensure that they are breaking up messages prior to passing them to the low-level API.  Otherwise, the `Write` function would no longer be atomic unit b/c it would have to make multiple produce requests to break up large message batches.  That seems very undesirable!

Issue #128, opened 2018-10-17
"unsupported message version 2 found in fetch response" errror


frranck:
Hello,
On big messages after increasing the KAFKA_MESSAGE_MAX_BYTES variable, I'm getting
"unsupported message version 2 found in fetch response" errors while doing a conn.ReadBatch,
Any idea what it mean ?

the code is there:
https://github.com/diadata-org/diadata/blob/master/pkg/dia/helpers/kafkaHelper/Kafka.go#L275

achille-roussel:
Hello @frranck Are you using kafka-go to both produce and consume into Kafka?

It could help to know which version of Kafka you are running as well!

VictorDenisov:
The problem can occur in every version from 0.11 and up:
https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-FetchResponse

Read paragraph that starts with:
In Kafka 0.11, the structure of the 'MessageSet' and 'Message' were significantly changed. 

I don't yet quite understand how clients are supposed to recognize the Message set of version 1 and version 2.

VictorDenisov:
It looks like as we read offset, size, crc and version of the first message and see that version is 2 then we should stop parsing message and start parsing batch header after parsing batch header we can start parsing individual messages according to version 2 of the batch format.

VictorDenisov:
Though it's unclear how to reproduce it. I have a broker on my local machine with 1.1.0. version and it doesn't reproduce the issue.

VictorDenisov:
I have a reproduction.

VictorDenisov:
@frranck What value do you have in KAFKA_MESSAGE_MAX_BYTES . I'm troubles getting a stable reproduction of the issue.

frranck:
Hi @VictorDenisov I put  KAFKA_MESSAGE_MAX_BYTES = 10000000
https://github.com/diadata-org/diadata/blob/master/deployments/docker-compose.kafka.yml#L41

VictorDenisov:
@frranck can you try this branch: https://github.com/VictorDenisov/kafka-go/tree/new_v2_message_reader

VictorDenisov:
@achille-roussel according to fetch response v2 doc: https://kafka.apache.org/protocol#The_Messages_Fetch I don't see any MessageSetSize in its header.
However according to: https://github.com/segmentio/kafka-go/blob/master/read.go#L282
[...]

PR #146, opened 2018-11-20
V2 message reader

VictorDenisov:
Fix for https://github.com/segmentio/kafka-go/issues/128

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -108,7 +108,7 @@ func (batch *Batch) Read(b []byte) (int, error) {
 	batch.mutex.Lock()
 	offset := batch.offset
 
-	_, _, err := batch.readMessage(
+	_, _, _, err := batch.readMessage(
 		func(r *bufio.Reader, size int, nbytes int) (int, error) {
 			if nbytes < 0 {
 				return size, nil
@@ -150,7 +150,7 @@ func (batch *Batch) ReadMessage() (Message, error) {
 	msg := Message{}
 	batch.mutex.Lock()
 
-	offset, timestamp, err := batch.readMessage(
+	offset, timestamp, headers, err := batch.readMessage(
 		func(r *bufio.Reader, size int, nbytes int) (remain int, err error) {
 			msg.Key, remain, err = readNewBytes(r, size, nbytes)
 			return
@@ -166,6 +166,7 @@ func (batch *Batch) ReadMessage() (Message, error) {
 	msg.Partition = batch.partition
 	msg.Offset = offset
 	msg.Time = timestampToTime(timestamp)
+	msg.Headers = headers
 
 	return msg, err
 }
@@ -173,12 +174,12 @@ func (batch *Batch) ReadMessage() (Message, error) {
 func (batch *Batch) readMessage(
 	key func(*bufio.Reader, int, int) (int, error),
 	val func(*bufio.Reader, int, int) (int, error),
-) (offset int64, timestamp int64, err error) {
+) (offset int64, timestamp int64, headers []Header, err error) {
 	if err = batch.err; err != nil {
 		return
 	}
 
-	offset, timestamp, err = batch.msgs.readMessage(batch.offset, key, val)
+	offset, timestamp, headers, err = batch.msgs.readMessage(batch.offset, key, val)
 	switch err {
 	case nil:
 		batch.offset = offset + 1
--- a/batch.go
+++ b/batch.go
@@ -151,7 +151,6 @@ func (batch *Batch) ReadMessage() (Message, error) {
 	batch.mutex.Lock()
 
 	var offset, timestamp int64
-	offset = 0
 	var headers []Header
 	var err error
 
--- a/batch.go
+++ b/batch.go
@@ -150,16 +150,23 @@ func (batch *Batch) ReadMessage() (Message, error) {
 	msg := Message{}
 	batch.mutex.Lock()
 
-	offset, timestamp, headers, err := batch.readMessage(
-		func(r *bufio.Reader, size int, nbytes int) (remain int, err error) {
-			msg.Key, remain, err = readNewBytes(r, size, nbytes)
-			return
-		},
-		func(r *bufio.Reader, size int, nbytes int) (remain int, err error) {
-			msg.Value, remain, err = readNewBytes(r, size, nbytes)
-			return
-		},
-	)
+	var offset, timestamp int64
+	offset = 0
+	var headers []Header
+	var err error
+
+	for offset < batch.conn.offset {
+		offset, timestamp, headers, err = batch.readMessage(
+			func(r *bufio.Reader, size int, nbytes int) (remain int, err error) {
+				msg.Key, remain, err = readNewBytes(r, size, nbytes)
+				return
+			},
+			func(r *bufio.Reader, size int, nbytes int) (remain int, err error) {
+				msg.Value, remain, err = readNewBytes(r, size, nbytes)
+				return
+			},
+		)
+	}
 
 	batch.mutex.Unlock()
 	msg.Topic = batch.topic
--- a/batch.go
+++ b/batch.go
@@ -220,16 +220,21 @@ func (batch *Batch) readMessage(
 			// This design decision was made to maximize the chances of keeping
 			// the connection open, the trade off being to lose precision on the
 			// read deadline management.
-			if !batch.deadline.IsZero() && time.Now().After(batch.deadline) {
-				err = RequestTimedOut
-			} else {
-				err = io.EOF
-			}
+			err = checkTimeoutErr(batch.deadline)
 			batch.err = err
 		}
 	default:
 		batch.err = err
 	}
 
+	return
+}
+
+func checkTimeoutErr(deadline time.Time) (err error) {
+	if !deadline.IsZero() && time.Now().After(deadline) {
+		err = RequestTimedOut
+	} else {
+		err = io.EOF
+	}
 	return
 }
--- a/batch.go
+++ b/batch.go
@@ -20,7 +20,7 @@ type Batch struct {
 	mutex         sync.Mutex
 	conn          *Conn
 	lock          *sync.Mutex
-	msgs          *messageSetReader
+	msgs          *messageSetReaderV1
 	deadline      time.Time
 	throttle      time.Duration
 	topic         string
--- a/batch.go
+++ b/batch.go
@@ -167,7 +167,7 @@ func (batch *Batch) ReadMessage() (Message, error) {
 	)
 	for batch.conn != nil && offset < batch.conn.offset {
 		if err != nil {
-			goto end
+			break
 		}
 		offset, timestamp, headers, err = batch.readMessage(
 			func(r *bufio.Reader, size int, nbytes int) (remain int, err error) {
@@ -181,7 +181,6 @@ func (batch *Batch) ReadMessage() (Message, error) {
 		)
 	}
 
-end:
 	batch.mutex.Unlock()
 	msg.Topic = batch.topic
 	msg.Partition = batch.partition
--- a/batch.go
+++ b/batch.go
@@ -20,7 +20,7 @@ type Batch struct {
 	mutex         sync.Mutex
 	conn          *Conn
 	lock          *sync.Mutex
-	msgs          messageSetReader
+	msgs          *messageSetReader
 	deadline      time.Time
 	throttle      time.Duration
 	topic         string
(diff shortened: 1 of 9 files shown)


message.go
```diff
@@ -105,7 +106,16 @@ func (s messageSet) writeTo(w *bufio.Writer) {
...
 
-type messageSetReader struct {
+type messageSetReader interface {
+	readMessage(min int64,
+		key func(*bufio.Reader, int, int) (int, error),
+		val func(*bufio.Reader, int, int) (int, error),
+	) (offset int64, timestamp int64, err error)
+	remaining() (remain int)
+	discard() (err error)
+}
```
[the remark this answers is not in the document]
VictorDenisov: @achille-roussel I don't see how my messageSetReader interfaces is different than previous implementation of messageSetReader. It was a structure allocated every time the batch was read. Can you elaborate?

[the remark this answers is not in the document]
VictorDenisov: I read your article: https://segment.com/blog/allocation-efficiency-in-high-performance-go-services/
It's sad. I'll rework it.

read.go
```diff
@@ -42,6 +42,28 @@ func readInt64(r *bufio.Reader, sz int, v *int64) (int, error) {
 	return peekRead(r, sz, 8, func(b []byte) { *v = makeInt64(b) })
 }
 
+func readVarInt(r *bufio.Reader, sz int, v *int64) (int, error) {
```
stevevls: I think that you can rewrite this function to make it more concise, reduce the number of loops, and also make it leverage/read more like kafka-go.  Something like:

```
func readVarInt(r *bufio.Reader, sz int, v *int64) (remain int, err error) {
	l := 0
	for done := false; !done && err == nil; {
		remain, err = peekRead(r, sz, 1, func(b []byte) {
			done = b[0]&0x80 == 0
			*v |= int64(b[0]&0x7f) << uint(l*7)
		})
		l++
	}
	return
}
```

read_test.go
```diff
@@ -7,6 +7,38 @@ import (
 	"testing"
 )
 
+func TestReadVarInt(t *testing.T) {
+	//buf := []byte{127, 10, 10, 0}
+	testCases := map[int64][]byte{
+		128:   []byte{128, 1, 10, 0},
```
stevevls: I would love to see a couple of base cases such as `0` and `1` just for sanity's sake.  I also think that since the interplay between `sz` and `remain` is so important, it would be better not to throw away the return value and to verify, making sure we have some test cases where not all bytes are consumed.

message.go
```diff
@@ -251,3 +287,200 @@ func extractOffset(base int64, msgSet []byte) (offset int64, err error) {
...
+	firstTimestamp       int64
+	maxTimestamp         int64
+	producerId           int64
+	producerEpoch        int16
+	firstSequence        int32
+}
+
+type compression int8
+
+const (
```
stevevls: I don't see these used anywhere.  Do we need them?  The compression codes should already be registered in `compression.go`

VictorDenisov:
I addressed the comments. It would be great to incorporate it into CI against a real v2 kafka broker, but my reproduction is very unstable and I don't see how to make kafka reliably produce v2 messages. Maybe the submitter of #128 can help us with it.

achille-roussel:
@stevevls didn't you have a way to force generating v2 messages when the values are very large of something like this?

@VictorDenisov If we want this PR to be complete I guess we could also support producing v2 messages. For example, if the message header isn't empty we have to generate a record batch (since the old message set didn't support headers), we could modify the produce requests to use messages v2 in this case, and test the full write + read cycle of messages v2 by producing messages with a non-empty header list.

Let me know what you think.

achille-roussel:
If you want to take a crack at it, this is where it would happen: https://github.com/segmentio/kafka-go/blob/master/write.go#L182-L279

stevevls:
Yeah...it seems that producing a large number of messages in one produce request will cause the broker to convert the v1 MessageSet to a v2 RecordSet.  This code fails for me:

```
func TestMessagesV2(t *testing.T) {
	topic := "test"
	msgs := make([]kafka.Message, 100)
	for i := range msgs {
		value := fmt.Sprintf("Hello World %d!", i)
		msgs[i] = kafka.Message{Value: []byte(value)}
	}

		Brokers: []string{"127.0.0.1:9092"},
		Topic:   topic,
	})
	defer w.Close()

	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	if err := w.WriteMessages(ctx, msgs...); err != nil {
		t.Errorf("failed to produce messages: %+v", err)
	}

	r := kafka.NewReader(kafka.ReaderConfig{
		Brokers:   []string{"127.0.0.1:9092"},
		Topic:     topic,
		Partition: 0,
		MaxWait:   10 * time.Millisecond,
		MinBytes:  1,
		MaxBytes:  1024,
	})
	defer r.Close()

	ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()
	_, err := r.ReadMessage(ctx)
	if err != nil {
		t.Fatal(err)
	}
}
```

You can verify that it's hitting the default case in `readMessageHeader` by either enabling logging or dropping a breakpoint.

VictorDenisov:
If I'm using a later version of kafka not the one in the docker compose then it works. With the docker compose from the project kafka seems to produce a malformed response. I'm investigating.

VictorDenisov:
This is the debug output:
```
V2 test

2018/11/27 22:01:55 Constructing message set reader V2
2018/11/27 22:01:55 Compression: 0
2018/11/27 22:01:55 BatchAttributes: 0
2018/11/27 22:01:55 Message count: 100, remain: 1031
   >> Last value from the header^^^^
2018/11/27 22:01:55 remain: 1031
2018/11/27 22:01:55 message length: 40, remain: 1030
2018/11/27 22:01:55 message attrs: 0, remain: 1029
2018/11/27 22:01:55 timestamp delta: 0
2018/11/27 22:01:55 offset delta: 0
2018/11/27 22:01:55 key len: 1
2018/11/27 22:01:55 Read key: 
2018/11/27 22:01:55 key: 28
2018/11/27 22:01:55 value len: 72
2018/11/27 22:01:55 Read value: ello World 0!(Hello World 1!(Hello World 2!(Hello Worl
```
The distance between two consecutive Hello World is too small to have a complete message v2 header.

VictorDenisov:
It looks like it happens because the size of the batch that was sent is smaller than MaxBytes in the reader and kafka broker doesn't really split message batches into smaller chunks. So far I guess it's producer's responsibility to send batches that can be consumed by readers.

VictorDenisov:
@achille-roussel I also made some experiments with my code that made me look into this problem in the first place. It looks like it's not really this patch that fixed my code. It was the fact that I increased the maximum size of the batch and kafka started sending properly formatted messages. So, it's bitter to admin, but I don't really know if this patch works at all. I'll be looking into possible reproductions of v2 message format.

VictorDenisov:
Here, I would like to give an update on what is going on. I implemented writing messages in v2 format if the messages contain headers. You can see it in my repository's branch v2_message_writer. It looks like even if I write messages in v2 format I will get them in v1 format if I'm using fetch request v2. I presume that kafka is trying not to break old clients and doesn't return v2 messages in apis that existed before 0.11 version. I'll be implementing another set of functions that will allow us to decide which fetch request version to use when working.

It also tells me that if anybody sees v2 message format in their responses it's because kafka sending malformed requests for some reason, not because it actually sends v2 messages. It will only send v2 messages if you use one of the latest fetch request versions.

VictorDenisov:
@achille-roussel I posted two more PRs that should be precursors to final implementation of this PR: https://github.com/segmentio/kafka-go/pull/160 https://github.com/segmentio/kafka-go/pull/163

silence44:
Is there any progress on this PR? 

Currently when I'm using `ReadMessage()` I'm receiving messages with empty headers, but they exist.

VictorDenisov:
@silence44 You may wan to try this branch. The work is mostly done. I would like to add a couple of tests before calling a PR.

achille-roussel:
@VictorDenisov how far do you think this PR is from being complete? Do you need any help to push it through the finish line?

VictorDenisov:
@achille-roussel I think I could use some help in order to accelerate it. My main predicament right now is that setting offset is not compatible with v2 record batches(it keeps reading messages in the batch instead of resetting it). I'm trying to figure out how it works with v1 messages. This problem causes  compression tests fail.

VictorDenisov:
@achille-roussel There is another way to accelerate the PR. I can add an option to the reader that allows the user to decide whether he wants to use V5 fetch request or V2 fetch request. By default it would use V2 fetch request. Right now in this PR I'm using V5 fetch request by default, what breaks the tests.

VictorDenisov:
@achille-roussel My two last issues are with testConnReadEmptyWithDeadline from conn_test.go and TestMixedCompressedMessages from compression_test.go. I'm looking into them, but maybe you can provide any advice. The one with empty read should produce timeout, but produces "no bytes to read".

stevevls:
@VictorDenisov I wrote the compression tests, so maybe I can be of help there.  I installed an error logger on the reader, ran your tests, and I got this:

```
the kafka reader got an unknown error reading partition 0 of kafka-go-3025c5551f570a97 at offset 56: read tcp [::1]:56424->[::1]:9092: i/o timeout
```

There are 60 messages produced to the test topic, and I was able to `kafkcat` them all, so I believe the problem is on the reader side.  When I've seen that error before, it's when Kafka has sent all the bytes, but our code thinks that more are coming.  #143 is an example of that.

Interestingly enough, I ran the test 5 times, and it always complained about offset 56.  That can probably help you narrow it down to a condition where there are fewer bytes available than requested.

stevevls:
Actually, this was looking so similar to #143 that I debugged a bit further.  I found that it was hanging on `readNewBytes` which does indeed have a latent bug.  I have opened #213 to address the issue and added tests.  When I cherry-picked this change to your branch, the compression test passed.

VictorDenisov:
Thank you, @stevevls . In this case only one test left to fix. I'm looking into it.

VictorDenisov:
@achille-roussel It looks like here https://github.com/segmentio/kafka-go/pull/146/files#diff-a2ddca1600f34abb91244424f8431c2cR719 in case of a timeout kafka produces enough garbage to read responseHeaderV2 and then fail in reading a message from the batch here: https://github.com/segmentio/kafka-go/blob/master/batch.go#L202.
In case of V5 response we don't get enough bytes to read full V5 header and receive errShortRead as a result. I'm just inclined to copy the code from readMessage in batch to ReadBatch in conn.go.
I presume you have a story behind this custom behavior with timeout. You probably were overcoming certain limitations of kafka and I would like to know this story behind it.

VictorDenisov:
@achille-roussel The PR is ready for review.

batch.go
```diff
@@ -160,25 +165,42 @@ func (batch *Batch) ReadMessage() (Message, error) {
 			return
 		},
 	)
+	for batch.conn != nil && offset < batch.conn.offset {
+		if err != nil {
+			goto end
```
achille-roussel:
```suggestion
			break
```
It seems like we're not skipping any code, we just want to break out?

Another option would be to check `for err == nil && ... {` directly in the loop.

read.go
```diff
@@ -345,6 +368,97 @@ func readFetchResponseHeader(r *bufio.Reader, size int) (throttle int32, waterma
...
+		FirstOffset int64
+	}
+	var p struct {
+		Partition           int32
+		ErrorCode           int16
+		HighwaterMarkOffset int64
+		LastStableOffset    int64
+		LogStartOffset      int64
+	}
+	var MessageSetSize int32
```
achille-roussel:
```suggestion
	var messageSetSize int32
```

batch.go
```diff
@@ -150,7 +150,12 @@ func (batch *Batch) ReadMessage() (Message, error) {
 	msg := Message{}
 	batch.mutex.Lock()
 
-	offset, timestamp, err := batch.readMessage(
+	var offset, timestamp int64
+	offset = 0
```
achille-roussel: `offset` is already zero here, remove this line maybe?

achille-roussel:
This is looking good! Thanks a lot @VictorDenisov for this massive contribution! 🚀

stevevls:
Echoing @achille-roussel, thanks so much for this contribution!  This is huge. 😄

VictorDenisov:
Thanks, guys, for being responsive maintainers!!!

PR #149, opened 2018-11-25
fix kafka.reader.fetch.count metric

achille-roussel:
This PR attempts to provide a smooth transition towards fixing the typo in the `kafka.reader.fetch.count` metric.

The idea is to produce both `kafka.reader.fetch.count` and `kafak.reader.fetch.count` for a couple of versions, and eventually remove the deprecated field (the one with the typo).

Please take a look and let me know if you have any concerns about it.


yields:
lgtm!

achille-roussel:
Thanks for reviewing @yields !

PR #153, opened 2018-11-29
Retrieve the current controller from any broker

fvosberg:
Hi,

at first, thank you very much for your great library. It's inspiring and I would like to contribute. I have little experience with Kafka though.

I have the requirement to create a topic in Kafka explicitly. So I've found the `*Conn.CreateTopics(...TopicConf) error```. I've started using this function and it worked like a charm until I've added some more broker to my Cluster. After startup, it wasn't deterministic which broker would have gotten elected to be the controller and NotController errors started to occur. 

After some research I've found the MetaData Request, which has the information I needed. So I've implemented a method to retrieve the ControllerURL from the current connection. I'm not sure how the API should look like or the implementation. This is just a first version without a unit test, because I'm not very experienced in Kafka and I've developed it with an integration test locally. 

So this ends up in me having theses questions:

- Is this the way to go, if I want to implement the topic creation explicitly with segmentio/kafka-go?
- Are you interested in having a function implemented, which returns the controller URL?
- How should the API should look like?
- Should I change the implementation?

After theses questions are answered I would be more than happy to implement it (with one or more tests).

Thanks in advance,
Fredi

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -138,6 +138,30 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 	return c
 }
 
+// GetController requests kafka for the current controller and returns its URL
+func (c *Conn) GetController() (controller string, err error) {
+	err = c.readOperation(
+		func(deadline time.Time, id int32) error {
+			return c.writeRequest(metadataRequest, v1, id, topicMetadataRequestV1([]string{}))
+		},
+		func(deadline time.Time, size int) error {
+			var res metadataResponseV1
+
+			if err := c.readResponse(size, &res); err != nil {
+				return err
+			}
+			for _, broker := range res.Brokers {
+				if broker.NodeID == res.ControllerID {
+					controller = fmt.Sprintf("%s:%d", broker.Host, broker.Port)
+					break
+				}
+			}
+			return nil
+		},
+	)
+	return controller, err
+}
+
 // DeleteTopics deletes the specified topics.
 func (c *Conn) DeleteTopics(topics ...string) error {
 	_, err := c.deleteTopics(deleteTopicsRequestV1{


conn.go
```diff
@@ -138,6 +138,30 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 	return c
 }
 
+// GetController requests kafka for the current controller and returns its URL
+func (c *Conn) GetController() (controller string, err error) {
```
achille-roussel: How about just calling this `Controller` or may `ReadController` to match the pattern of other read operations?

Could the method return a `Broker` type? That would provide stronger type checking by the compiler than a string.

Let me know what you think!

fvosberg: Thanks for your input. I would totally agree with naming it Controller. I can do that. I've thought about the return type and there is no Broker Type, yet. I can implement that for sure, but I didn't want to make any big proposals to the API. Another idea, which came to my mind was something like this:

```
// ControllerConn returns a connection to the controller of a cluster
func (c *Conn) ControllerConn() (controller *Conn, err error)
```

The downside of this is, that we would have to *reuse* the dial parameters. Maybe adding a dialer as parameter and if it's nill reusing the existing one? 

I'm open to any suggestions as long as I'm able to connect to the controller afterwards ;)

fvosberg:
Hi there,

I hope that everybody had nice holidays. Is there anything I can do to reduce some impediments here and speed it up a little? 

Thanks
fredi

conn.go
achille-roussel: Actually there is a `kafka.Broker` type https://godoc.org/github.com/segmentio/kafka-go#Broker

Obtaining a connection to the controller is an interesting idea, however it seems like it could be quite inefficient if all we need is the address. Maybe we can add another method on the `kafka.Dialer` type to do this, like `DialController`?

achille-roussel:
Thanks for following up, I left a response to your question, let me know what you think!

achille-roussel:
@fvosberg do you still have interest in finishing up this change?

PR #158, opened 2018-12-12
implement missing error codes

vrischmann:
While starting work on zstd I got a `76` error code from the broker. Looking at the code I realized there's a bunch of missing error codes for Kafka up to 2.1.

I got all of them from the 2.1 branch of the [Kafka client](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/protocol/Errors.java#L243-L290).

Sidenote, I'm wondering why errror code 55 is called `BrokerAuthorizationFailed` because in the [Java code](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/protocol/Errors.java#L239-L242) it's called `OperationNotPermitted` and the description given doesn't mention anything about authorizations.


achille-roussel:
I remember copying the error names from some sources but I can't recall which ones. Happy to introduce an alias to BrokerAuthorizationFailed if it helps.

PR #160, opened 2018-12-17
Add ApiVersions request

VictorDenisov:
I would like to merge this as a separate PR for easier review. I will need this change is implementing v2 message reader: https://github.com/segmentio/kafka-go/pull/146
I'll be using this request right after connecting to kafka cluster in order to determine which fetch requests I'm allowed to send. 
So far the plan is to implement fetch request v5. Fetch request v5 seems to have been introduced in kafka 0.11.0.0. And presumably this is the first version where server returns message sets in v2 format.
So, if fetch request v5 is available I'll be sending v5 otherwise v2.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -71,6 +71,7 @@ type Conn struct {
 
 	// number of replica acks required when publishing to a partition
 	requiredAcks int32
+	apiVersions  []ApiVersion
 }
 
 // ConnConfig is a configuration object used to create new instances of Conn.
@@ -135,6 +136,11 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 		}},
 	}).size()
 	c.fetchMaxBytes = math.MaxInt32 - c.fetchMinSize
+	var err error
+	c.apiVersions, err = c.ApiVersions()
+	if err != nil {
+		c.apiVersions = nil
+	}
 	return c
 }
 
@@ -183,6 +189,7 @@ func (c *Conn) findCoordinator(request findCoordinatorRequestV0) (findCoordinato
 	err := c.readOperation(
 		func(deadline time.Time, id int32) error {
 			return c.writeRequest(groupCoordinatorRequest, v0, id, request)
+
 		},
 		func(deadline time.Time, size int) error {
 			return expectZeroSize(func() (remain int, err error) {
--- a/conn.go
+++ b/conn.go
@@ -1012,6 +1012,66 @@ func (c *Conn) requestHeader(apiKey apiKey, apiVersion apiVersion, correlationID
 	}
 }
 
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -1012,6 +1019,66 @@ func (c *Conn) requestHeader(apiKey apiKey, apiVersion apiVersion, correlationID
...
+		if size, err = readInt16(&c.rbuf, size, &r[i].MinVersion); err != nil {
+			return nil, err
+		}
+		if size, err = readInt16(&c.rbuf, size, &r[i].MaxVersion); err != nil {
+			return nil, err
+		}
+	}
+
+	if errorCode != 0 {
+		return r, Error(errorCode)
```
achille-roussel: Should we return a nil `[]ApiVersion` here? Or is there value in exposing a partial result?

conn.go
```diff
@@ -135,6 +136,11 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 		}},
 	}).size()
 	c.fetchMaxBytes = math.MaxInt32 - c.fetchMinSize
+	var err error
+	c.apiVersions, err = c.ApiVersions()
+	if err != nil {
+		c.apiVersions = nil
```
achille-roussel: Should we put a default value here?

conn.go
```diff
@@ -71,6 +71,7 @@ type Conn struct {
 
 	// number of replica acks required when publishing to a partition
 	requiredAcks int32
+	apiVersions  []ApiVersion
```
achille-roussel: Is there a guarantee on the ordering of the values here? I'm wondering how the lookups will work.

conn.go
VictorDenisov: My approach was, - I'm giving you whatever I have just use it at your own discretion since there was an error. I can return nil here if you think it makes more sense.

conn.go
VictorDenisov: I guess I can put the value that lists all the versions that the current implementation of kafka-go works with.

conn.go
achille-roussel: Up to you, I don't have strong feelings about it.

conn.go
VictorDenisov: I can check for ordering. But I just realized that for our purposes it, probably, makes more sense to put it in a map: map[api key]ApiVersion

conn.go
achille-roussel: Seems like it could help making good use of this field 👍

conn.go
achille-roussel: A map makes sense to me 👍

achille-roussel:
Looking good, thanks for adding this!

PR #163, opened 2018-12-19
Message writer for record batches

VictorDenisov:


write.go
```diff
@@ -232,12 +242,173 @@ func writeProduceRequestV2(w *bufio.Writer, codec CompressionCodec, correlationI
...
+	writeInt64(bufWriter, baseOffset)
+
+	remainderBuf := &bytes.Buffer{}
+	remainderBuf.Grow(int(size - 12)) // 12 = batch length + base offset sizes
+	remainderWriter := bufio.NewWriter(remainderBuf)
+
+	writeInt32(remainderWriter, -1) // partition leader epoch
+	writeInt8(remainderWriter, 2)   // magic byte
+
+	//writeInt16(remainderWriter, 0) // TODO write crc
```
achille-roussel: Is this expected to remain commented out?

write.go
```diff
@@ -232,12 +242,173 @@ func writeProduceRequestV2(w *bufio.Writer, codec CompressionCodec, correlationI
...
+			baseTime = msg.Time
+		}
+	}
+	return
+}
+
+const MaxUint = ^uint64(0)
+const MinUint = 0
+const MaxInt = int64(MaxUint >> 1)
+const MinInt = -MaxInt - 1
```
achille-roussel: Do we have to export those constants? 

Also why not use `math.MaxUint64`, `math.MaxInt64`, etc... ?

write.go
```diff
@@ -286,3 +457,31 @@ func msgSize(key, value []byte) int32 {
 		sizeofBytes(key) +
 		sizeofBytes(value)
 }
+
+// Messages with magic >2 are called records. This method writes messages using message format 2.
+func writeRecord(w *bufio.Writer, attributes int8, baseTime time.Time, baseOffset int64, msg Message) {
+	buf := &bytes.Buffer{}
```
achille-roussel: This intermediary buffer is a bit worrying performance-wise, but we can follow up with an optimization to compute the size later.

write.go
VictorDenisov: It should be removed. I calculate crc below.

write.go
VictorDenisov: Indeed. I should use standard values.

write.go
VictorDenisov: :+1: I'll send a separate PR for it.

VictorDenisov:
@achille-roussel I addressed the comments.

write.go
```diff
@@ -232,12 +253,184 @@ func writeProduceRequestV2(w *bufio.Writer, codec CompressionCodec, correlationI
...
+			1 + // attributes
+			8 + // timestamp
+			sizeofBytes(msg.Key) +
+			sizeofBytes(msg.Value)
 	}
 
-	return w.Flush()
+	buf := &bytes.Buffer{}
+	buf.Grow(int(size))
+	bufWriter := bufio.NewWriter(buf)
```
achille-roussel: This is likely going to introduce a significant performance regression on producers which use kafka-go. This code is introducing two new temporary buffers on code which we have on hot paths of our data pipelines (I'd assume other companies do as well). Those buffers will result in two extra copies of the message set, and increase the memory footprint of kafka producers. We need to introduce this change without taking the risk to suffer performance regressions.

VictorDenisov: Ok. I'll try to do something about it.

VictorDenisov:
@achille-roussel , I got rid of as many extra buffers as possible. The only remaining buffer is crc buffer. bufio.Writer doesn't allow me to jump back and forth the output buffer. In order to calculate crc I would need to leave space for crc value and then write the data, calculate crc, go back and update crc value(as it's actually done in c implementation of kafka client). i thought it would require a bigger refactoring, that would be better off in a separate PR if we decide to go down this route.

VictorDenisov:
@achille-roussel don't merge just yet. It looks like this change breaks the tests. I need to look into it.

PR #168, opened 2019-01-02
Use the unMasked attributes for codec resolution

ShaneSaww:
This is working for codecs registered as  0 and 1 as 0 and 1 in binary are 0 and 01. But for Snappy (2) this turns into a 10 and never finds the correct decoder codec.

### DIFF
--- a/message.go
+++ b/message.go
@@ -143,7 +143,7 @@ func (r *messageSetReader) readMessage(min int64,
 		code := attributes & compressionCodecMask
 		if code != 0 {
 			var codec CompressionCodec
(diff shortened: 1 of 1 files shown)


achille-roussel:
Although I'm not sure I follow the logic you describe, snappy is 2 (10) like you said, but the compression codec mask is 0x03 (11) which also fits in.

My understanding is the code just breaks for any compression codec if other bits are set in the attributes, but maybe I'm missing something?

ShaneSaww:
@achille-roussel It looks like it could break if other bits are set as well. 
https://play.golang.org/p/lo7N7SgZl4O

Issue #116, opened 2018-09-11
WriteMessages should return offset


achille-roussel:
The Produce API response returns the offset at which a batch of messages was written, but this is not an information we currently expose to the program, even tho it is very valuable.

I'm thinking of introducing a method with one of those signatures, and I'd like to get some feedback on which one you think may be the best approach:
```go
func (*Conn) WriteMessagesAt(...Message) (nbytes int, offset int64, err error)
```
```go
func (*Conn) WriteFull(...Message) (nbytes int, offset int64, time time.Time, throttle time.Duration, err error)
```
```go
func (*Conn) Produce(req *ProduceRequest) (res *ProduceResponse, err error)
```
I want to avoid breaking backward compatibility even if we're pre-1.0.

Happy to here your thoughts.

rjenkins:
I prefer `func (*Conn) Produce(req *ProduceRequest) (res *ProduceResponse, err error)`, also fwiw this is similar to existing Producer APIs for example https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html#send-org.apache.kafka.clients.producer.ProducerRecord-

silence44:
Is this feature is in progress right now? 
I want to switch from sarama to segmetio but this missing feature stops me. Is there another way to get write offset in current library version?

rjenkins:
@silence44 I think we could get this in pretty quick, let me see if we can get a PR up for next week.

scarbo87:
I could take on this task.
If I understand correctly, all the necessary data is written in the method `produceResponsePartitionV2.readFrom`. Is it so?
Do you want to make a new method 
```go
func (*Conn) Produce(req *ProduceRequest) (res *ProduceResponse, err error)
``` 
analogue of the 
```go
func (c *Conn) WriteCompressedMessages(codec CompressionCodec, msgs ...Message) (int, error)
```
?

In the current code, it is faster to make this variant 
```go
func (c *Conn) WriteCompressedMessagesAt(codec CompressionCodec, msgs ...Message) (nbytes int, offset int64, time time.Time, throttle time.Duration, err error)
```

rjenkins:
@scarbo87 after some discussions with @achille-roussel we decided to introduce a new client rather modify the existing ones. I'm out until Jan 4th, so current ETA on this for me is week of Jan 7th. If you need it sooner and want to take a pass at it, PRs are always welcome :+1:

scarbo87:
[...]

PR #171, opened 2019-01-03
issue-116: added method WriteCompressedMessagesAt

scarbo87:
#116: added method WriteCompressedMessagesAt, that returns offset, partition, timestamp

### DIFF
--- a/compression_test.go
+++ b/compression_test.go
@@ -1,4 +1,4 @@
-package kafka_test
+package kafka
 
 import (
 	"context"
@@ -77,7 +77,7 @@ func testCompressedMessages(t *testing.T, codec kafka.CompressionCodec) {
 	t.Run("produce/consume with"+codecToStr(codec.Code()), func(t *testing.T) {
 		t.Parallel()
 
-		topic := kafka.CreateTopic(t, 1)
+		topic := CreateTopic(t, 1)
 		w := kafka.NewWriter(kafka.WriterConfig{
 			Brokers:          []string{"127.0.0.1:9092"},
 			Topic:            topic,
@@ -144,7 +144,7 @@ func testCompressedMessages(t *testing.T, codec kafka.CompressionCodec) {
 func TestMixedCompressedMessages(t *testing.T) {
 	t.Parallel()
 
-	topic := kafka.CreateTopic(t, 1)
+	topic := CreateTopic(t, 1)
 
 	offset := 0
 	var values []string
--- a/compression_test.go
+++ b/compression_test.go
@@ -1,4 +1,4 @@
-package kafka
+package kafka_test
 
 import (
 	"context"
@@ -77,7 +77,7 @@ func testCompressedMessages(t *testing.T, codec kafka.CompressionCodec) {
 	t.Run("produce/consume with"+codecToStr(codec.Code()), func(t *testing.T) {
 		t.Parallel()
 
-		topic := CreateTopic(t, 1)
+		topic := kafka.CreateTopic(t, 1)
 		w := kafka.NewWriter(kafka.WriterConfig{
 			Brokers:          []string{"127.0.0.1:9092"},
 			Topic:            topic,
@@ -144,7 +144,7 @@ func testCompressedMessages(t *testing.T, codec kafka.CompressionCodec) {
 func TestMixedCompressedMessages(t *testing.T) {
 	t.Parallel()
 
-	topic := CreateTopic(t, 1)
+	topic := kafka.CreateTopic(t, 1)
 
 	offset := 0
 	var values []string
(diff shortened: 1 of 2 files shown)


scarbo87:
> What do you think about exposing the timestamp as a time.Time value?

@achille-roussel , this is not native for kafka and if I am not mistaken, the time.Time allocates additional memory: https://kafka.apache.org/protocol#The_Messages_Produce
```Produce Response (Version: 2) => [responses] throttle_time_ms 
  responses => topic [partition_responses] 
    topic => STRING
    partition_responses => partition error_code base_offset log_append_time 
      partition => INT32
      error_code => INT16
      base_offset => INT64
      log_append_time => INT64
  throttle_time_ms => INT32```

**Comment by achille-roussel** [comment]
One goal we had in mind with this project was to provide an API which integrates well with Go programs, so even if the underlying representation on the Kafka protocol uses an integer timestamp we can expose the value as a `time.Time` to the program (and avoid showing low-level implementation details).

`time.Time` is a bit larger than `int64` but it will be stack-allocated so there is no performance concern here.

Note that I'm not saying we need to write a `time.Time` in the wire protocol, just do the conversion internally between `int64` and `time.Time`.

Using `time.Time` means the compiler does stronger type checking, there is no risk of misusing a `time.Time` value as an integer for example, or get the ordering wrong and confuse it with the offset that the method returns.

Let me know if you have follow up questions.

File: conn.go [hunk 1]
```diff
@@ -846,6 +862,12 @@ func (c *Conn) WriteCompressedMessages(codec CompressionCodec, msgs ...Message)
 					if err == nil && p.ErrorCode != 0 {
 						err = Error(p.ErrorCode)
 					}
+					if err == nil {
+						partition = p.Partition
+						offset = p.Offset
+						appendTime = time.Unix(p.Timestamp, 0)
```
achille-roussel: The Kafka protocol documentation seems to indicate that the timestamp is expressed as a number of milliseconds since the unix epoch, but the code assumes it is seconds, or am I missing something?

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ProduceAPI

conn.go
```diff
@@ -789,31 +789,47 @@ func (c *Conn) WriteMessages(msgs ...Message) (int, error) {
...
+func (c *Conn) WriteCompressedMessagesAt(codec CompressionCodec, msgs ...Message) (nbytes int, partition int32, offset int64, appendTime time.Time, err error) {
+	return c.writeCompressedMessages(codec, msgs...)
+}
+
+func (c *Conn) writeCompressedMessages(codec CompressionCodec, msgs ...Message) (nbytes int, partition int32, offset int64, appendTime time.Time, err error) {
+
+	nbytes = 0
+	partition = -1
+	offset = -1
+	appendTime = time.Unix(-1, 0)
```
achille-roussel: Those special values will modify the behavior of the `WriteMessages` and `WriteCompressedMessages` methods, how about just keeping the zero values as defaults?

compression_test.go
```diff
@@ -1,4 +1,4 @@
-package kafka_test
+package kafka
```
achille-roussel: What was the reason this package needed to be changed?

scarbo87: I`m sorry,  I had an error in local imports.

conn.go
scarbo87: Why change behavior? They are not used: `nbytes, _, _, _, err = c.writeCompressedMessages(codec, msgs...)`

Ok, I will make zero values as defaults.

conn.go
scarbo87: Interesting, you are probably right.
I read this documentation here: https://kafka.apache.org/protocol#The_Messages_Produce
Not sayd about milliseconds )

conn.go
```diff
@@ -789,31 +789,47 @@ func (c *Conn) WriteMessages(msgs ...Message) (int, error) {
...
+func (c *Conn) WriteCompressedMessagesAt(codec CompressionCodec, msgs ...Message) (nbytes int, partition int32, offset int64, appendTime time.Time, err error) {
+	return c.writeCompressedMessages(codec, msgs...)
+}
+
+func (c *Conn) writeCompressedMessages(codec CompressionCodec, msgs ...Message) (nbytes int, partition int32, offset int64, appendTime time.Time, err error) {
+
+	nbytes = 0
+	partition = 0
+	offset = 0
+	appendTime = time.Unix(0, 0)
```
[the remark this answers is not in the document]
scarbo87: Of course. I was mistaken, thinking that the default `time.Time` will be `time.Now`.
Default `time.Time` equals `0001-01-01 00:00:00 +0000 UTC`

achille-roussel:
Thanks for your contribution!

scarbo87:
Thanks for the review.

PR #182, opened 2019-01-23
Adding error case for UnknownTopicOrPartition into Read loop

ShaneSaww:
This error can happen when a partition is moved brokers, and can be recovered from. It should be treated similar as a Leadership move which is handled in the case below.


ShaneSaww:
Closing in favor of https://github.com/segmentio/kafka-go/pull/184

PR #184, opened 2019-01-25
Adding in extra errors for the read loop and adding a partition watcher. 

ShaneSaww:
1) Adding error case for UnknownTopicOrPartition into the Read loop.
This error can happen when a partition is moved brokers, and can be recovered from. It should be treated similar as a Leadership move which is handled in the case below.

2) Adding in a func in the handshake to watch for partition changes from the brokers. This allows kafka-go to pick up new partitions without having to restart. It does this by polling the brokers on an interval then causing a rebalance via a new call to `handshake`. 

This includes:
https://github.com/segmentio/kafka-go/pull/182


ShaneSaww:
Any update on this? I have another PR about ready to send up, but it's built on top of this! 
So I'm wondering the how the review is going.

ShaneSaww:
I added some code to fix https://github.com/segmentio/kafka-go/issues/188 into this PR. 
I'll close and re-open this PR without that. Sorry for the noise.

Issue #43, opened 2017-12-24
Support Kafka record headers


Pryz:
Since 0.11.x, Kafka support record headers. See : https://issues.apache.org/jira/browse/KAFKA-4208 / https://cwiki.apache.org/confluence/display/KAFKA/A+Case+for+Kafka+Headers.

There is a lot of use cases where headers could be useful !

Pryz:
I will try to take a look to it the coming week :)

townsendbt:
Hi,
I'm wondering if there is any progress on this issue? I need to use Message Headers. I'll use the confluent go client if this isn't on the radar for release.

Thanks,
Brian.

Pryz:
Unfortunately no progress since we don't have the need for it anymore. Feel free to send a PR if you have time to work on this :)

scarbo87:
I tried to implement #185 , but failed.
Messages are sent to Kafka, but there are no headers.
Have any ideas?

achille-roussel:
Making progress on this front with https://github.com/segmentio/kafka-go/pull/163

PR #185, opened 2019-01-26
issue-43: [WIP] headers added

scarbo87:
I tried to implement #43 , but failed.
Messages are sent to Kafka, but there are no headers.
Have any ideas?

### DIFF
--- a/crc32.go
+++ b/crc32.go
@@ -7,7 +7,7 @@ import (
 	"sync"
 )
 
-func crc32OfMessage(magicByte int8, attributes int8, timestamp int64, key []byte, value []byte) uint32 {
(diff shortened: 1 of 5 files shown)


crc32.go
```diff
@@ -16,6 +16,12 @@ func crc32OfMessage(magicByte int8, attributes int8, timestamp int64, key []byte
 	}
 	b.writeBytes(key)
 	b.writeBytes(value)
+
+	b.writeArray(len(headers), func(i int) {
+		b.writeBytes(headers[i].Key)
```
scarbo87: I tried the option when the key is a string: https://kafka.apache.org/documentation/#record
But it also does not work

scarbo87:
thanks for the info, this is exactly what i need

PR #186, opened 2019-01-28
Test against multiple Kafka versions

Pryz:
`kafka-go` is getting more and more contributors using multiple versions of Kafka. To ensure proper compatibility with the supported versions, we need to test against all of them.

This PR creates a CI job for Kafka 0.11.0.1, 1.1.1 and 2.1.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  build:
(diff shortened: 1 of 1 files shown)


Pryz:
Yes I was pretty surprise as well but there is not that many protocol changes between 1.x and 2.x.

PR #187, opened 2019-01-28
Run tests for kafka 0.10

stevevls:
Added a utility method to check broker version so that tests for
newer functionality can be skipped.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,33 @@
 version: 2
 jobs:
+  kafka-010:
+    working_directory: /go/src/github.com/segmentio/kafka-go
+    environment:
+      KAFKA_VERSION: "0.10.1"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1,7 +1,33 @@
 version: 2
 jobs:
+  kafka-010:
+    working_directory: /go/src/github.com/segmentio/kafka-go
+    environment:
+      KAFKA_VERSION: "0.10.1"
```
Pryz: should be 0.10.1.0 ?

stevevls: Do we need to go all the way out to the 4th number?  I thought protocol changes couldn't happen at that release level.

Pryz: You're correct ! :)

Issue #164, opened 2018-12-27
The leaveGroup is not called on close


ppatierno:
I have a simple consumer application which uses the blocking `ReadMessage` in an infinite for loop for getting messages.
In order to close the consumer gracefully I am using a context with "cancel" ...

```
ctx, cancel := context.WithCancel(context.Background())
```

There is a go routine which is in charge to get signals (SIGING, SIGKILL, ...) and calling the above `cancel` function in order to have the main routine exiting from the blocking `ReadMessage`.

After that I call the `close` on the reader.
What I see from the Kafka broker log but even using Wireshark is that the leave group message is not sent by the reader when the application calls `close`.
Is this pattern the right way? Or cancelling the context corrupts something so that it's normal that the `close` doesn't call the `leaveGroup`? What's the right way to do it? 

From the doc it seems I am on the right way because is says "... and exposes an API that supports asynchronous cancellations and timeouts using Go contexts"

I also tried to call the `close` just after receiving a message (inside the for loop) in order to avoid using canceling the context but it's the same, no `leaveGroup` is called.

ppatierno:
I don't know the library in details (I should dig into the code) but if "reader is closed" means that the consumer has closed the connection to the brokers, then it will start a partitions rebalancing in the consumer group. The clean way should be having the consumer sending the "leave group" while it's still connected.

WillAbides:
I've recently run into this issue after moving from confluent-kafka-go.

This code in `reader.Close()` confused me:  https://github.com/segmentio/kafka-go/blob/e442c3981eae27aa4da0fa4d7d31d17d0d9721de/reader.go#L1157-L1168

From what I can tell, `r.coordinator()` on line 1164 will always return an error because `r.stctx` will have been canceled by the `r.stop()` on line 1158.  Since `r.coordinator()` is always returning an error, there is no condition where `r.leaveGroup()` is executed.  So the whole block starting on line 1161 is essentially a noop (except for unsuccessfully attempting to dial the leader).

That block could be moved above `r.cancel()` so it works as it was apparently intended, or be removed entirely so it doesn't confuse people like me into thinking that leaveGroup is called on Close.

If @achille-roussel is correct, and it isn't always desirable to leave the group on Close, then there should still be a way to call `leaveGroup()`.  Either by exposing it or by adding `LeaveConsumerGroupOnClose` to ReaderConfig.

PR #189, opened 2019-01-29
Ensure reader leaves consumer group on graceful exit

stevevls:
Issue #164

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -815,6 +815,14 @@ func (r *Reader) handshake() error {
 
 	rg.Wait()
 
+	select {
+	case <-r.stctx.Done():
+		// this reader is closing...leave the consumer group.
+		_ = r.leaveGroup(conn)
+	default:
+		// another consumer has left the group
+	}
+
 	return nil
 }
 
@@ -1157,26 +1165,6 @@ func (r *Reader) Close() error {
 	r.cancel()
 	r.stop()
 	r.join.Wait()
-
-	if r.useConsumerGroup() {
-		// gracefully attempt to leave the consumer group on close
-		if generationID, membershipID := r.membership(); generationID > 0 && membershipID != "" {
-
-			// note that we do not call r.coordinator() because that passes
-			// r.stctx go DialContext.  we just cancelled that context a couple
-			// lines above by calling r.stop().
-			r.mutex.Lock()
-			address := r.address
-			r.mutex.Unlock()
-
-			// DialContext will wrap the background context to add appropriate
-			// timeout/deadline configuration.
-			conn, err := r.config.Dialer.DialContext(context.Background(), "tcp", address)
-			if err == nil {
-				_ = r.leaveGroup(conn)
-			}
-		}
-	}
 
 	<-r.done
 
--- a/reader.go
+++ b/reader.go
@@ -1161,7 +1161,18 @@ func (r *Reader) Close() error {
 	if r.useConsumerGroup() {
 		// gracefully attempt to leave the consumer group on close
 		if generationID, membershipID := r.membership(); generationID > 0 && membershipID != "" {
-			if conn, err := r.coordinator(); err == nil {
+
+			// note that we do not call r.coordinator() because that passes
+			// r.stctx go DialContext.  we just cancelled that context a couple
+			// lines above by calling r.stop().
+			r.mutex.Lock()
+			address := r.address
+			r.mutex.Unlock()
+
+			// DialContext will wrap the background context to add appropriate
+			// timeout/deadline configuration.
+			conn, err := r.config.Dialer.DialContext(context.Background(), "tcp", address)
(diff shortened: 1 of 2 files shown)


reader.go
```diff
@@ -1161,7 +1161,18 @@ func (r *Reader) Close() error {
...
+			// note that we do not call r.coordinator() because that passes
+			// r.stctx go DialContext.  we just cancelled that context a couple
+			// lines above by calling r.stop().
+			r.mutex.Lock()
+			address := r.address
+			r.mutex.Unlock()
+
+			// DialContext will wrap the background context to add appropriate
+			// timeout/deadline configuration.
+			conn, err := r.config.Dialer.DialContext(context.Background(), "tcp", address)
```
stevevls: @achille-roussel can you 2x check that this is a correct assumption?

achille-roussel: That's assuming there is a timeout configured on the dialer, but if the program sets none it's definitely looking for trouble so I think we're good there!

WillAbides:
I was working on a PR as well.  For the fix, I just moved the whole `if r.useConsumerGroup()` block to the top of the `Close()` method.  I don't know if that method or this one will work better.

Can I suggest adding the test below?  It passes for the changes in this PR.

```go
func TestCloseLeavesGroup(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
	defer cancel()
	topic := makeTopic()
	createTopic(t, topic, 1)
	r := NewReader(ReaderConfig{
		Brokers:  []string{"localhost:9092"},
		Topic:    topic,
		GroupID:  makeGroupID(),
		MinBytes: 1,
		MaxBytes: 10e6,
		MaxWait:  100 * time.Millisecond,
	})
	prepareReader(t, ctx, r)
	groupID := r.Config().GroupID

	// wait for generationID > 0 so we know our reader has joined the group
	membershipTimer := time.After(1 * time.Second)
	for {
		done := false
		select {
		case <-membershipTimer:
			t.Fatalf("our reader never joind its group")
		default:
			generationID, _ := r.membership()
			if generationID > 0 {
				done = true
			}
		}
		if done {
			break
		}
	}

	err := r.Close()
	if err != nil {
		t.Fatalf("unexpected error closing reader: %s", err.Error())
	}

	conn, err := Dial("tcp", "localhost:9092")
	if err != nil {
		t.Fatalf("error dialing: %v", err)
	}
	resp, err := conn.describeGroups(describeGroupsRequestV1{
		GroupIDs: []string{groupID},
	})
	if err != nil {
		t.Fatalf("error from describeGroups %v", err)
	}
	if len(resp.Groups) != 1 {
		t.Fatalf("expected 1 group. got: %d", len(resp.Groups))
	}
	if len(resp.Groups[0].Members) != 0 {
		t.Fatalf("expected group membership size of %d, but got %d", 0, len(resp.Groups[0].Members))
	}
}
```

stevevls:
@achille-roussel I agree re: the connection.  Let me dig a little further.  We may be able to add a closing check in `handshake` and leave the group then.  We def. don't want to just call `leaveGroup` because `handshake` is called once per generation...that could result in all consumers leaving the group when one does.

@WillAbides I don't think that moving the code up is a good idea because we should wait for the reader go routines to exit before leaving the group.  The unit test looks great, though.  Feel free to commit it to this branch!

stevevls:
Just pushed code that moves leaving the group into the `handshake` function.  It checks to see if the reader's context is done and leaves the group if so.  The reader's context isn't used by the call graph when invoking `leaveGroup` function, so we don't run into the original issue.  The update passes the unit test that @WillAbides provided.  I didn't check in the test, though, because I prefer that he gets recognition for contributing it.

WillAbides:
I don't have rights to add to this branch directly, but I created #190 to add the test to this branch.

WillAbides:
That test passed before 2cead73, but it's failing now.

stevevls:
@WillAbides I think there's a timing condition in the test.  Adding a short sleep after the `membershipTimer` receive and before the `r.Close()` makes the test pass.  I think the issue is that we don't have a connection with which we can leave the group until after calling `rebalance` on line 794.  However, I'm seeing that the call to `rebalance` is error-ing out if the reader is closed before it finishes.  I may be able to refactor this so that we get a handle to the connection earlier.

stevevls:
@WillAbides Give this latest commit a try.  I refactored it so that all of the consumer group operations share a single `Conn`.  By establishing the connection first, we can then put the `defer` block to leave the group if any of the CG operations fail.  The test you provided was passing reliably for me against a local kafka 0.11.0 cluster.

WillAbides:
@stevevls It's also passing for me using the kafka provided by docker-compose.

stevevls:
Cool.  So if this looks okay to @achille-roussel, we can merge it in.

PR #190, opened 2019-01-29
Add test for Reader leaving its consumer group on Close()

WillAbides:
This adds a test that will pass when a Reader with a group membership successfully leaves its group on `r.Close()`


stevevls:
Thanks for the contribution!

PR #191, opened 2019-01-30
Fix for Dialer.LookupPartition

scarbo87:
There is no break in forever loop, so we will never receive error “ UnknownTopicOrPartition” on line 181. My code fixes case when there is no partitions found.


scarbo87:
Of course

scarbo87:
I was wrong, if you use `context.WithTimeout`, then will exit this function `LookupPartition` by timeout.

But, if you set up a Kafka with an option `auto.create.topics.enable=false`, then the error `t.TopicErrorCode=3 (UNKNOWN_TOPIC_OR_PARTITION)` on this [line](https://github.com/segmentio/kafka-go/blob/master/conn.go#L758) will not be processed.
On this [line](https://github.com/segmentio/kafka-go/blob/master/dialer.go#L160) partitions and err will equlas nil, this is wrong.

achille-roussel:
@scarbo87 should we always return `UnknownTopicOrPartition` if `len(partitions) == 0` ?

scarbo87:
No, this is not a good solution.

Currently I found two problems in the code.

1. We need to process sleep boolean result on line XXX, otherwise we ignore context and have memory leak.
2. If we connect to kafka and topic not exists, then on line XXX currently we will not return error, because c.topic will be always empty string.

I will take care of this. Let me know if you agree.

scarbo87:
@achille-roussel, this fix is more correct

PR #196, opened 2019-02-06
Added coverage and timeout options to test build step

stevevls:
This will give us diagnostic info the next time the tests fail due to a deadlock condition (e.g. https://circleci.com/gh/segmentio/kafka-go/745).


stevevls:
Works for me.  I picked 5 minutes pretty randomly.  We'll need to remember to bump up the numbers if/when we add more tests.  At some point, we may also want to introduce a Makefile so that we don't have to touch so many places in order to change test params.  But I think that's something we can do another time.

PR #204, opened 2019-02-12
Adding partition watcher, so Kafka-go picks up new partitions. 

ShaneSaww:
Adding in a func in the handshake to watch for partition changes from the brokers. This allows kafka-go to pick up new partitions without having to restart. It does this by polling the brokers on an interval then causing a rebalance via a new call to `handshake`.

This is from breaking apart:
https://github.com/segmentio/kafka-go/pull/184

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -788,10 +788,10 @@ func (r *Reader) commitLoop(conn *Conn) func(stop <-chan struct{}) {
 }
 
 // partitionWatcher queries kafka and watches for partition changes, triggering a rebalance if changes are found.
-// Similar to heartbeat it's okay to return on error here as if you are unable to ask a broker for baisc metadata
+// Similar to heartbeat it's okay to return on error here as if you are unable to ask a broker for basic metadata
 // you're in a bad spot and should rebalance. Commonly you will see an error here if there is a problem with
 // the connection to the coordinator and a rebalance will establish a new connection to the coordinator.
-func (r *Reader) partitionWatcher(conn *Conn) func(stop <-chan struct{}) {
+func (r *Reader) partitionWatcher(conn partitionReader) func(stop <-chan struct{}) {
 	return func(stop <-chan struct{}) {
 		ticker := time.NewTicker(r.config.PartitionWatchInterval)
 		defer ticker.Stop()
@@ -802,7 +802,7 @@ func (r *Reader) partitionWatcher(conn *Conn) func(stop <-chan struct{}) {
 			})
 			return
 		}
-		oParts := make([]int, len(ops))
+		oParts := len(ops)
 		for {
 			select {
 			case <-stop:
@@ -815,7 +815,7 @@ func (r *Reader) partitionWatcher(conn *Conn) func(stop <-chan struct{}) {
 					})
 					return
 				}
-				if len(ops) != len(oParts) {
+				if len(ops) != oParts {
 					r.withErrorLogger(func(l *log.Logger) {
 						l.Printf("Partition changes found, reblancing group: %v.", r.config.GroupID)
 					})
@@ -960,8 +960,12 @@ type ReaderConfig struct {
 	//
 	// Default: 5s
 	//
-	// Only used when GroupID is set
+	// Only used when GroupID is set and WatchPartitionChanges is set.
 	PartitionWatchInterval time.Duration
+
+	// WatchForPartitionChanges is used to inform kafka-go that a consumer group should be
+	// polling the brokers and rebalancing if any partition changes happen to the topic.
+	WatchPartitionChanges bool
 
 	// SessionTimeout optionally sets the length of time that may pass without a heartbeat
 	// before the coordinator considers the consumer dead and initiates a rebalance.
@@ -995,10 +999,6 @@ type ReaderConfig struct {
 	// ErrorLogger is the logger used to report errors. If nil, the reader falls
 	// back to using Logger instead.
 	ErrorLogger *log.Logger
-
-	// WatchForPartitionChanges is used to inform kafka-go that a consumer group should be
-	// polling the brokers and rebalancing if any partition changes happen to the topic.
-	WatchPartitionChanges bool
 }
 
 // ReaderStats is a data structure returned by a call to Reader.Stats that exposes
--- a/reader.go
+++ b/reader.go
@@ -788,9 +788,12 @@ func (r *Reader) commitLoop(conn *Conn) func(stop <-chan struct{}) {
 }
 
 // partitionWatcher queries kafka and watches for partition changes, triggering a rebalance if changes are found.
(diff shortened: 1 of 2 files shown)


reader.go
```diff
@@ -783,6 +787,55 @@ func (r *Reader) commitLoop(conn *Conn) func(stop <-chan struct{}) {
...
+		ops, err := conn.ReadPartitions(r.config.Topic)
+		if err != nil {
+			r.withErrorLogger(func(l *log.Logger) {
+				l.Printf("Problem getting partitions during startup, %v\n, Returning and setting up handshake", err)
+			})
+			return
+		}
+		// It's possible that the list of partitions returned are not in any order
+		// so put them into a map to compare.
+		oParts := make(map[int]struct{})
```
stevevls: I don't think you need a map here since we believe that Kafka won't return a partition range with holes.  You can use a slice instead to avoid a heap allocation:
```suggestion
		oParts := make([]int, len(ops))
		for i := range oParts {
			oParts[i] = i
		}
```

reader.go
```diff
@@ -783,6 +787,55 @@ func (r *Reader) commitLoop(conn *Conn) func(stop <-chan struct{}) {
 	}
 }
 
+// partitionWatcher queries kafka and watches for partition changes, triggering a rebalance if changes are found.
```
stevevls: If you can leave a little bit of clarifying documentation as to why it's okay to return from errors here, that would be great.  On an initial reading, I had thought that returning in the face of an error might not be appropriate because it could introduce consumer group rebalance churn due to tangential errors (e.g. failure to list partitions isn't necessarily a good reason to rebalance the group...it would be a more cautious approach to let the group continue and retry in the background).  However, since we're share the `Conn` with the other CG go routines, it means that we're interacting the CG coordinator...and if we lose that connection, then the other go routines in the `runGroup` will also be exiting and we'll have a rebalance on our hands.  Do you see where I'm going with that?  If you can find a way to say it more concisely, that'd be great. 😄

reader.go
```diff
@@ -783,6 +787,55 @@ func (r *Reader) commitLoop(conn *Conn) func(stop <-chan struct{}) {
...
+					})
+					return
+				}
+				if len(ops) != len(oParts) {
+					r.withErrorLogger(func(l *log.Logger) {
+						l.Printf("Partition changes found, reblancing group: %v.", r.config.GroupID)
+					})
+					return
+				}
+				for _, p := range ops {
```
stevevls: Do we need this check here?  Seems to me like the length check above should be sufficient.  I'm having a hard time imagining a case where this check would cause us to return where the length check would not.

reader.go
ShaneSaww: I can do that, I was using the map to ensure that it's sorted as well, because I didn't want to do the array each time.

reader.go
ShaneSaww: I agree. Just checking the length should work.

reader.go
ShaneSaww: If I'm only checking the length tho, the array should be fine.

ShaneSaww:
@stesegment PR updated with the feedback, I'll start working on a unit test next.

reader.go
```diff
@@ -788,9 +788,12 @@ func (r *Reader) commitLoop(conn *Conn) func(stop <-chan struct{}) {
 }
 
 // partitionWatcher queries kafka and watches for partition changes, triggering a rebalance if changes are found.
+// Similar to heartbeat it's okay to return on error here as if you are unable to ask a broker for baisc metadata
```
stevevls: Nit: baisc -> basic

reader.go
```diff
@@ -799,12 +802,7 @@ func (r *Reader) partitionWatcher(conn *Conn) func(stop <-chan struct{}) {
 			})
 			return
 		}
-		// It's possible that the list of partitions returned are not in any order
-		// so put them into a map to compare.
-		oParts := make(map[int]struct{})
-		for _, p := range ops {
-			oParts[p.ID] = struct{}{}
-		}
+		oParts := make([]int, len(ops))
```
stevevls: After the latest round of changes, I think we can go even one farther and just keep a count.  What do you think?

reader.go
```diff
@@ -908,6 +954,15 @@ type ReaderConfig struct {
 	// Only used when GroupID is set
 	CommitInterval time.Duration
 
+	// PartitionWatchInterval indicates how often a reader checks for partition changes.
+	// If a reader sees a partition change (such as a partition add) it will rebalance the group
+	// picking up new partitions.
+	//
+	// Default: 5s
+	//
+	// Only used when GroupID is set
```
stevevls: Let's also call out that `WatchPartitionChanges` needs to be set.

reader.go
ShaneSaww: Yeah, I think that would work. 
Let me get that unit test working and I can see how it looks.

ShaneSaww:
@stevevls whoops! Sorry other Steve! 

I added in a unit test into the consumer group section. I'm happy to move as it, it might be a little strange where it's at now.

reader_test.go
```diff
@@ -1047,6 +1053,54 @@ func testReaderConsumerGroupReadContentAcrossPartitions(t *testing.T, ctx contex
...
 }
 
+// Build a struct to implement the ReadPartitions interface.
+type MockConnWatcher struct {
+	count int
+	partitions [][]Partition
+}
+func (m *MockConnWatcher) ReadPartitions(topics ...string) (partitions []Partition, err error) {
+	partitions = m.partitions[m.count]
+	m.count++
```
stevevls: Unfortunately, this is the type of project where there are lots of time-based unit tests.  Our approach is to make sure that things eventually end up in a consistent state, and usually to deal with CI noisiness, add a lot of extra time.  One thing we could do here to make the test more resilient is to cap `m.count` at `len(m.partitions)-1`.  That way, the test will be idempotent even if the watcher comes in a couple more times.

reader_test.go
```diff
@@ -1047,6 +1053,54 @@ func testReaderConsumerGroupReadContentAcrossPartitions(t *testing.T, ctx contex
...
+			},
+		},
+	}
+
+	rg := &runGroup{}
+	rg = rg.WithContext(ctx)
+	r.config.PartitionWatchInterval = watchTime
+	rg.Go(r.partitionWatcher(conn))
+	rg.Wait()
+	if time.Now().Sub(now).Seconds() > r.config.PartitionWatchInterval.Seconds() * 2 {
```
stevevls: If you make the change above, you can add in additional padding here.  Since there's a lot of parallel tests going on, it's usually best to add additional time.

reader_test.go
ShaneSaww: That makes sense! Never thought about that. 
Changes pushed

stevevls:
Looks great.  Thanks for your contribution!

PR #205, opened 2019-02-13
Downgraded delete topics request/response to v0

stevevls:
This change makes the DeleteTopics function work with Kafka 0.10.0


stevevls:
Before the test was disabled for Kafka 0.10 (e.g. https://github.com/segmentio/kafka-go/pull/205/files#diff-f9df1a974544ddcc7ccada9d10f705a4L244).  The tests pass now, so we should be good.

PR #208, opened 2019-02-15
Adding MaxMessageBytes into the ProducerConfig

ShaneSaww:
This adds in the ability to send a batch of messages based on Max byte count. 
The idea from this is taken from the Java client settings of `max.request.size` which also is where the default comes from. This feels a  is a bit more natural/kafka life as topics will also a have max message size.


writer.go
```diff
@@ -132,11 +139,12 @@ type WriterStats struct {
...
-	WriteTime DurationStats `metric:"kafka.writer.write.seconds"`
-	WaitTime  DurationStats `metric:"kafka.writer.wait.seconds"`
-	Retries   SummaryStats  `metric:"kafka.writer.retries.count"`
-	BatchSize SummaryStats  `metric:"kafka.writer.batch.size"`
+	DialTime       DurationStats `metric:"kafka.writer.dial.seconds"`
+	WriteTime      DurationStats `metric:"kafka.writer.write.seconds"`
+	WaitTime       DurationStats `metric:"kafka.writer.wait.seconds"`
+	Retries        SummaryStats  `metric:"kafka.writer.retries.count"`
+	BatchSize      SummaryStats  `metric:"kafka.writer.batch.size"`
+	BatchSizeBytes SummaryStats  `metric:"kafka.writer.batch.size.bytes"`
```
achille-roussel:
```suggestion
	BatchBytes SummaryStats  `metric:"kafka.writer.batch.bytes"`
```
This would better follow the naming scheme of other metrics.

writer.go
```diff
@@ -73,6 +74,12 @@ type WriterConfig struct {
 	// The default is to use a target batch size of 100 messages.
 	BatchSize int
 
+	// Limit the maximum size of a request in bytes before being sent to
+	// a partition.
+	//
+	// The default is to use a kafka default value of 1048576.
+	MaxMessageBytes int
```
achille-roussel: Should this be called `BatchBytes` to fit the naming scheme of `BatchSize` and `BatchTimeout`?

writer.go
```diff
@@ -575,7 +599,9 @@ func (w *writer) run() {
 	var done bool
 	var batch = make([]Message, 0, w.batchSize)
 	var resch = make([](chan<- error), 0, w.batchSize)
+	var lstMsg *writerMessage
```
achille-roussel: Does this stand for `lastMsg`? Probably worth renaming because I initially read this as `listMsg`, which made it hard to understand what was happening.

Also can we not make it a pointer type? `writerMessage` is passed by value everywhere, it seems weird to have it be a pointer here.

ShaneSaww: Yeah, `lastMsg`. I'll update the name.
I'll see what I can about removing the pointer.

ShaneSaww:
@achille-roussel If an individual message is sent that's too large it will throw an error.  Inside of WriteMessage(https://github.com/segmentio/kafka-go/pull/208/files#diff-dc03dbc53655fac7a4c8d16b1b4144c1R297) 
This is a little strange as you could have a partial batch go through, but I think it's okay.

writer.go
```diff
@@ -275,6 +289,13 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 		}
 
 		for _, msg := range msgs {
+			if int(msg.message().size()) > w.config.MaxMessageBytes {
+				err := errors.New(fmt.Sprintf("The message is %d bytes "+
```
achille-roussel: Any reasons not to use `fmt.Errorf` here?

ShaneSaww: I don't think so, I'll update it.

ShaneSaww:
@achille-roussel For errors I think the latter case (don't return an error, just skip large messages) works fine.
The only thing I'm worried about is letting the caller know which, which means we could increment an error counter and log. 
Do you have any thoughts on returning an error if there is only a single record passed in?

writer.go
```diff
@@ -585,15 +611,30 @@ func (w *writer) run() {
...
+			resch = append(resch, lstMsg.res)
+			batchSizeBytes += int(lstMsg.msg.message().size())
+			lstMsg = nil
+		}
 		select {
 		case wm, ok := <-w.msgs:
 			if !ok {
 				done, mustFlush = true, true
 			} else {
+				if int(wm.msg.message().size())+batchSizeBytes > w.maxMessageBytes {
```
stevevls: This is out of scope, but do you happen to know how `maxMessageBytes` relates to compression?  Since this calculation is happening prior to compression, I'm wondering whether there's an opportunity to get better compression ratios by continuing to add to the batch until the *compressed* payload is nearing `maxMessageBytes`.

writer_test.go
```diff
@@ -30,6 +31,18 @@ func TestWriter(t *testing.T) {
 			scenario: "running out of max attempts should return an error",
 			function: testWriterMaxAttemptsErr,
 		},
+		{
+			scenario: "writing a message larger then the max bytes should return an error",
+			function: testWriterMaxBytes,
+		},
+		{
+			scenario: "writing a batch of message based on batch byte size",
+			function: testWrtierBatchBytes,
```
stevevls: Nit: typo

writer_test.go
```diff
@@ -30,6 +31,18 @@ func TestWriter(t *testing.T) {
...
+		},
+		{
+			scenario: "writing a batch of messages",
+			function: testWrtierBatchSize,
```
stevevls: Nit: typo
```suggestion
			function: testWriterBatchSize,
```

writer_test.go
```diff
@@ -151,6 +164,37 @@ func testWriterMaxAttemptsErr(t *testing.T) {
 	}
 }
 
+func testWriterMaxBytes(t *testing.T) {
```
stevevls: Let's use a context with a timeout instead of `context.Background`.  That will defend against cases where the `WriteMessages` call may hang.

Also, it would be great if you could `createTopic` instead of using constant topic names to ensure that we always start from a clean slate.  That's more of import for folks testing locally than for CI.

writer_test.go
```diff
@@ -189,3 +233,101 @@ func readPartition(topic string, partition int, offset int64) (msgs []Message, e
 		msgs = append(msgs, msg)
 	}
 }
+
+func testWrtierBatchBytes(t *testing.T) {
```
stevevls: I feel like we're missing a condition here.  We're ensuring that the two messages are sent in a batch, but I'd also like to see the case where they're written in two separate requests (e.g. if `MaxMessageBytes` were set to `47` so that we cover the `lastMsg != nil` condition).

writer.go
ShaneSaww: Yeah, this is a great question. 
All the batching is done before compression so it doens't related. So we are erring on the side of being conservative with the batch size.  
There is an opportunity here to get better compression, but we would need to do some re-working on where do the compression as currently it's done in the `write` and done to the entire batch. 

I think this is something to fix, but I would want to get this PR done first before getting after that.  

related: https://issues.apache.org/jira/browse/KAFKA-4169

stevevls: Cool.  Thanks for digging that Kafka ticket up.  I agree that it's a much bigger change that's out of scope for this PR.

writer_test.go
ShaneSaww: Great ideas!

writer_test.go
ShaneSaww: Case added!

ShaneSaww:
@achille-roussel Just pushed up a change that skips messages inside of WriteMessages. I had to add an extra var to ensure that we don't wait on those skipped messages to pass something back on the `res` chan. It feels a little strange.

stevevls:
I think the code reads okay, but if I'm understanding correctly, the contract of this function just got scary because now we can drop messages without any indication to the caller that it happened.

My thoughts are that the `Writer` is a higher-level abstraction, so there's no reason why one call to `Writer. WriteMessages` must correspond to exactly one produce request.  I'd prefer that we break up the messages slice into chunks of `maxMessageBytes` or less and call produce multiple times.  This is in contrast to the low-level code in `Conn.WriteMessages`, which *must* map to exactly one produce request.  What do you think @achille-roussel?

ShaneSaww:
@stevevls I'm not sure that I'm following. 
Are you talking about moving the logic for checking how large a batch is into the `WriteMessages` func?

>. I'd prefer that we break up the messages slice into chunks of maxMessageBytes or less and call produce multiple times
isn't that what's happening in the writer `run` func now?

stevevls:
@ShaneSaww Sorry...I was reading the function incorrectly.  Please disregard!  I still am concerned that the large message gets dropped w/out returning an error, though.  Any chance we could surface it to the caller so that they can decide how to proceed?

ShaneSaww:
@stevevls Yeah. I'm happy to have it return an error instead of just log. 
The question that arises there is what happens if you hand a batch of messages into the Write and only 1 is too large.

stevevls:
Talked with @achille-roussel, and here's what we're thinking.  It's possible that a client has the expectation that the messages are published in order to a partition--we actually have one such producer in house.  Taking that into account, the safest behavior would be to stop producing with an error when we hit a message that's too large.  This feels like the safest course of action, plus it won't change the current behavior of kafka-go.  In general, we think that a producer creating messages that are too large is a buggy producer and shouldn't be left unchecked.

That said, we also should make it possible for a sophisticated user to recover and continue.  A special error type would come in handy here.  It could carry with it the message (or messages) that are too large and any remaining unpublished messages.  The caller could then look for this error and react accordingly.  What do you think?

ShaneSaww:
@stevevls That makes sense, and it makes the WriteMessages func have less changes overall. 
I like the idea of a custom error type, but I'm not sure on how we would return the messages that are too large, without changing the return of `WriteMessages` or you do something strange with the Error interface.

stevevls:
My initial thought was to have a custom error type that looks something like:
```
type MessageTooLargeError struct {
	Message   Message
	Remaining []Message
}

func (e MessageTooLargeError) Error() string {
	return MessageSizeTooLarge.Error()
}
```
The downside, though, is that it wouldn't be comparable with the existing `MessageSizeTooLarge`, so we would be changing behavior of the API in a subtle and non b/w compatible way.  @achille-roussel What do you think?

achille-roussel:
The custom error type sounds great to me 👍 I think aiming for non-breaking changes, especially for error handling is the right approach.

stevevls:
@ShaneSaww This PR has been inactive for a bit.  Did you still want to try to land it?

ShaneSaww:
Yeah, I do. I'll try to push something up soon.

ShaneSaww:
@achille-roussel @stevevls PR updated to include the new ErrorType.

writer.go
```diff
@@ -274,7 +287,15 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 			return io.ErrClosedPipe
 		}
 
-		for _, msg := range msgs {
+		for i, msg := range msgs {
+			if int(msg.message().size()) > w.config.BatchBytes {
+				err := MessageTooLargeError{
+					Message: msg,
+					Remaining: msgs[i:],
```
stevevls:
```suggestion
					Remaining: msgs[i+1:],
```
I would expect to see `Remaining` contain the messages in the slice _after_ the message that's too large.  That way, I could feed them back into write.  WDYT?

writer_test.go
```diff
@@ -151,6 +168,38 @@ func testWriterMaxAttemptsErr(t *testing.T) {
...
+	}
+
+	if err := w.WriteMessages(context.Background(), Message{
+		Value: []byte("Hello World!"),
+	}); err == nil {
+		t.Error("expected error")
+		return
+	} else if err != nil {
+		switch err.(type) {
+		case MessageTooLargeError:
```
stevevls: I would love to see assertions for the `Message` and `Remaining` fields of the error.

writer.go
ShaneSaww: 👍 I agree. I'll make that change.

ShaneSaww:
@stevevls Tests and writer error updated.

stevevls:
Thanks @ShaneSaww.  Merging this! 😄

PR #209, opened 2019-02-15
Retrieve controller

rjenkins:
This PR is in reference to the feedback on the inactive PR https://github.com/segmentio/kafka-go/pull/153. Here we're changing the name of the func to `Controller` and returning a `Broker` as requested.


stevevls:
Any chance you could slap a unit test on that?

PR #213, opened 2019-02-21
Fix bug in readNewBytes so it doesn't attempt to read beyond sz

stevevls:


stevevls:
Test failures are our favorite racy reader stats test.  Unrelated to the PR. 😿

PR #214, opened 2019-02-21
Workaround for racy testReaderStats test

stevevls:
This makes me sad, but it will make me far less sad in the long term than seeing builds fail all the time. 😹


stevevls:
@achille-roussel I love the "keep running it until it passes" idea. 😹

PR #221, opened 2019-02-28
Set offset by timestamp.

rjenkins:
This is a fix to complete the inactive PR https://github.com/segmentio/kafka-go/pull/122/. Rather than using the coordinator we connect to the partition leader. Additionally there was a deadlock in the `SetOffsetWithTimestamp` func when attempting to call `r.SetOffset`, this has also been resolved.


reader.go
```diff
@@ -1483,6 +1483,37 @@ func (r *Reader) SetOffset(offset int64) error {
 	return err
 }
 
+// SetOffsetWithTimestamp changes the offset from which the next batch of messages will be
+// read given the timestamp t.
+//
+// The method fails if the unable to connect partition leader, or unable to read the offset
+// given the ts, or if the reader has been closed.
+func (r *Reader) SetOffsetWithTimestamp(ctx context.Context, t time.Time) error {
```
achille-roussel:
```suggestion
func (r *Reader) SetOffsetAt(ctx context.Context, t time.Time) error {
```
I'm not particularly attached to this suggestion, I feel like the current name is a bit verbose and redundant tho.

reader.go
```diff
@@ -1483,6 +1483,37 @@ func (r *Reader) SetOffset(offset int64) error {
...
+
+	for _, broker := range r.config.Brokers {
+		conn, err := r.config.Dialer.DialLeader(ctx, "tcp", broker, r.config.Topic, r.config.Partition)
+		if err != nil {
+			continue
+		}
+
+		deadline, _ := ctx.Deadline()
+		conn.SetDeadline(deadline)
+		offset, err := conn.ReadOffset(t)
```
achille-roussel: We're missing a `conn.Close()` after this line, aren't we?

rjenkins: good catch

PR #223, opened 2019-03-01
Added support for SASL PLAIN and SCRAM mechanisms

stevevls:
This PR picks up the work on the inactive PR #134 in order to drive it to conclusion.


sasl/scram/scram.go
```diff
@@ -0,0 +1,83 @@
...
+	"crypto/sha512"
+	"hash"
+
+	"github.com/pkg/errors"
+	"github.com/xdg/scram"
+)
+
+// HashFunction determines the hash function used by SCRAM to protect the user's
+// credentials.
+type HashFunction int
```
achille-roussel: Any reasons not to make this an interface with implementations for the SHA256 and SHA512 hash functions instead of an enum type?

stevevls: I did have reasons...the most important of which was to make sure we didn't export anything 3rd party libraries (e.g. in case we swap out scram providers later).  But I think that can also be done with an interface.  I just pushed up what I would see as the interface version...let me know which you prefer!

sasl/scram/scram.go
```diff
@@ -0,0 +1,80 @@
...
+	client *scram.Client
+	convo  *scram.ClientConversation
+}
+
+// Mechanism returns a new sasl.Mechanism that will use SCRAM with the provided
+// Algorithm to securely transmit the provided credentials to Kafka.
+//
+// SCRAM-SHA-256 and SCRAM-SHA-512 were added to Kafka in 0.10.2.0.  These
+// mechanisms will not work with older versions.
+func Mechanism(algo Algorithm, username, password string) (*mechanism, error) {
```
achille-roussel: Exposing an unexported type in the return value here is a bit weird, was this intended?

Issue #216, opened 2019-02-24
Transactions


VictorDenisov:
I would like to work on transactions for kafka-go: https://www.confluent.io/blog/transactions-apache-kafka/
Are there any objections?

Tasks:

- [x] Add isolation level config to reader (READ_COMMITTED, READ_UNCOMMITTED);
- [x] Introduce produce request V3 which will allow us to specify transactional id;
- [x] Handle control messages; #261 
- [x] Implement find transaction coordinator request; #275
- [x] Implement init producer id request; #275
- [ ] Implement idempotent writer
- [ ] Implement begin and end transaction.

achille-roussel:
The plan sounds solid to me 👍

Andy320:
@VictorDenisov @achille-roussel 
Thanks for your hard working about this major feature. And may I know the ETA？Looking forward to it in our project.

VictorDenisov:
It's difficult to say. Of course I'm just like you are interested in delivering it as soon as possible.

VictorDenisov:
@achille-roussel I have a question about this function: https://github.com/segmentio/kafka-go/blob/50823e6c2ed38b7ed6b896105636c1e0c65bb3d3/batch.go#L105
and this function
https://github.com/segmentio/kafka-go/blob/50823e6c2ed38b7ed6b896105636c1e0c65bb3d3/batch.go#L193

Was there any specific reason to implement `readMessage` with key and value readers as arguments? I would like to refactor it to return full message instead of reading it using `key` function.

I'm asking because transactions introduce control messages and control messages should not surface to the client app. The key of control messages specifies the type of the control message - aborted or committed transaction. It means I need to read the key of a control message within https://github.com/segmentio/kafka-go/blob/50823e6c2ed38b7ed6b896105636c1e0c65bb3d3/message.go#L440.
If this message is an abort message then I need to abandon the whole chunk of messages between the previous control message and the current control message. It means I need to read the messages, store them somewhere and once I understand what message I want to return to the client I need to invoke the `key`  function for the `key` of the message.

Long story short - it would make more sense to me to return full message from readMessage function.
Do you have any comment?

sneko:
@VictorDenisov great initiative! 👍 

I saw you closed some of your PRs, is it still something you're working on?

Thank you,

VictorDenisov:
[...]

PR #224, opened 2019-03-02
Isolation level

VictorDenisov:
This PR is for configuring isolation level in batch reading. It's also one of the steps in this story: #216.
There are two variants to configure isolation level either through configuring batch reading, as it's done here, or by configuring connection.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -675,10 +675,10 @@ func (c *Conn) ReadBatch(minBytes, maxBytes int) *Batch {
 		MinBytes: minBytes,
 		MaxBytes: maxBytes,
 	}
-	return c.ReadBatchWithConfig(cfg)
+	return c.ReadBatchWith(cfg)
 }
 
-func (c *Conn) ReadBatchWithConfig(cfg BatchReadConfig) *Batch {
+func (c *Conn) ReadBatchWith(cfg BatchReadConfig) *Batch {
 
 	var adjustedDeadline time.Time
 	var maxFetch = int(c.fetchMaxBytes)
--- a/conn.go
+++ b/conn.go
@@ -671,11 +671,10 @@ func (c *Conn) ReadMessage(maxBytes int) (Message, error) {
 // gives the minimum and maximum number of bytes that it wants to receive from
 // the kafka server.
 func (c *Conn) ReadBatch(minBytes, maxBytes int) *Batch {
-	cfg := ReadBatchConfig{
+	return c.ReadBatchWith(ReadBatchConfig{
 		MinBytes: minBytes,
 		MaxBytes: maxBytes,
-	}
-	return c.ReadBatchWith(cfg)
+	})
 }
 
 func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -82,6 +82,19 @@ type ConnConfig struct {
...
+type BatchReadConfig struct {
+	MinBytes       int
+	MaxBytes       int
+	IsolationLevel IsolationLevel
+}
+
+type IsolationLevel int8
+
+const (
+	READ_UNCOMMITTED IsolationLevel = 0
```
achille-roussel:
```suggestion
	ReadUncommitted IsolationLevel = 0
```

conn.go
```diff
@@ -82,6 +82,19 @@ type ConnConfig struct {
...
+	READ_COMMITTED   IsolationLevel = 1
```
achille-roussel:
```suggestion
	ReadCommitted   IsolationLevel = 1
```
Let's follow Go naming guidelines ;)

conn.go
```diff
@@ -658,17 +671,26 @@ func (c *Conn) ReadMessage(maxBytes int) (Message, error) {
...
 // the kafka server.
 func (c *Conn) ReadBatch(minBytes, maxBytes int) *Batch {
+	cfg := BatchReadConfig{
+		MinBytes: minBytes,
+		MaxBytes: maxBytes,
+	}
+	return c.ReadBatchWithConfig(cfg)
+}
+
+func (c *Conn) ReadBatchWithConfig(cfg BatchReadConfig) *Batch {
```
achille-roussel:
```suggestion
func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
```
- We have `NewConnWith` already, so let's match that naming scheme and call the method `ReadBatchWith`?
- To follow these naming schemes, how about also calling the configuration type `ReadBatchConfig`?

conn.go
```diff
@@ -658,17 +671,26 @@ func (c *Conn) ReadMessage(maxBytes int) (Message, error) {
 // gives the minimum and maximum number of bytes that it wants to receive from
 // the kafka server.
 func (c *Conn) ReadBatch(minBytes, maxBytes int) *Batch {
+	cfg := BatchReadConfig{
```
achille-roussel: This temporary variable doesn't seem useful, how about just inlining the declaration in argument to the method call?

VictorDenisov:
Done

achille-roussel:
> Could we add documentation for the new types and method that this PR adds?

VictorDenisov:
Ah, of course. I will.

VictorDenisov:
@achille-roussel I added some docs without documenting most obvious(in my opinion) things. Feel free to let me know if you want more docs.

write.go
```diff
@@ -196,7 +196,7 @@ func writeFetchRequestV5(w *bufio.Writer, correlationID int32, clientID, topic s
 	writeInt32(w, milliseconds(maxWait))
 	writeInt32(w, int32(minBytes))
 	writeInt32(w, int32(maxBytes))
-	writeInt8(w, int8(0)) // isolation level 0 - read uncommitted
+	writeInt8(w, isolationLevel) // isolation level 0 - read uncommitted
```
achille-roussel:
```suggestion
	writeInt8(w, isolationLevel)
```

Issue #200, opened 2019-02-11
Consumer group larger than partition count causes constant reconnection attempts 


cosmopetrich:
Running a group of kafka-go consumers that contains more consumers than there are partitions in a topic results in the surplus consumers continuously attempting to rejoin the group with no delay. 

The library will log these three messages 20+ times per second.

```
INFO[0015] rebalancing consumer group, example
INFO[0015] joinGroup succeeded for response, example.  generationID=51, memberID=example@local (github.com/segmentio/kafka-go)-72f08ec6-b36e-48cd-94b9-a33e32fb82e6
INFO[0015] rebalance failed for consumer group, example: received empty assignments for group, example as member example@local (github.com/segmentio/kafka-go)-72f08ec6-b36e-48cd-94b9-a33e32fb82e6 for generation 51
```

Running unnecessary extra consumers isn't usually desirable but it's something that can happen. For example, a Kubernetes event might result in additional replicas of a service being run for some length of time. If nothing else the large amount of connections and log entries generated by kafka-go can cause problems for connection-tracking firewalls and aggregated logging systems.

I'm quite new to Kafka. Is having a consumer with no allocations a problem? It seems like it could safely stay idle until the next rebalance. If it is very undesirable then would some kind of delay/backoff in the `run()` for loop be worthwhile? https://github.com/segmentio/kafka-go/blob/62cc6a8f0b9af4a2199adf28be3dd365d3e9e8f2/reader.go#L822-L833

abuchanan-nr:
@stevevls we're going to need this very soon, so I'm going to take a look. Just a heads up in case you already have a bunch of work done.

stevevls:
@abuchanan-nr Thanks for the heads up!  I haven't dug in on this bug yet, so please feel free to send in a PR.  Let me know if I can help.

abuchanan-nr:
@stevevls I think this is the source of the behavior: https://github.com/segmentio/kafka-go/blob/master/reader.go#L437

Question is, what to do instead. In the case where you have too many consumers (this bug), this should be an info log, I think. Is there another case? I can't think of one yet.

I can PR that simple change, but wanted to get some careful thought from others first.

pawanrawal:
This also happens when the topic doesn't exist and auto-creation of topics is disabled on the server. This loop: https://github.com/segmentio/kafka-go/blob/master/reader.go#L868 should try the reconnect with a delay in such cases.

abuchanan-nr:
[...]

PR #229, opened 2019-03-12
Don't rebalance on too many consumers. Wait for topic to exist.

abuchanan-nr:
Fix for #200 

I'll try to write tests for this. Putting this up since there's some discussion in #200 on the best approach.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -240,9 +240,6 @@ type partitionReader interface {
 // assignTopicPartitions uses the selected GroupBalancer to assign members to
 // their various partitions
 func (r *Reader) assignTopicPartitions(conn partitionReader, group joinGroupResponseV1) (GroupMemberAssignments, error) {
-	const backoffDelayMin = 100 * time.Millisecond
-	const backoffDelayMax = 1 * time.Second
-
 	r.withLogger(func(l *log.Logger) {
 		l.Println("selected as leader for group,", r.config.GroupID)
 	})
@@ -258,23 +255,9 @@ func (r *Reader) assignTopicPartitions(conn partitionReader, group joinGroupResp
 	}
 
 	topics := extractTopics(members)
-	var partitions []Partition
-	for attempt := 0; true; attempt++ {
-		if attempt != 0 {
-			if !sleep(r.stctx, backoff(attempt, backoffDelayMin, backoffDelayMax)) {
-				return nil, r.stctx.Err()
-			}
-		}
-		partitions, err = conn.ReadPartitions(topics...)
-		if err == UnknownTopicOrPartition {
-			r.withErrorLogger(func(l *log.Logger) {
-				l.Printf("error reading partitions, retrying: %v", err)
-			})
-			continue
-		}
-		if err != nil {
-			return nil, fmt.Errorf("unable to read partitions: %v", err)
-		}
+	partitions, err := conn.ReadPartitions(topics...)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read partitions: %v", err)
 	}
 
 	r.withLogger(func(l *log.Logger) {
@@ -469,33 +452,47 @@ func (r *Reader) syncGroup(conn *Conn, memberAssignments GroupMemberAssignments)
 }
 
 func (r *Reader) rebalance(conn *Conn) (map[string][]int32, error) {
-	const backoffDelayMin = 100 * time.Millisecond
-	const backoffDelayMax = 1 * time.Second
 
 	r.withLogger(func(l *log.Logger) {
 		l.Printf("rebalancing consumer group, %v", r.config.GroupID)
 	})
 
+	members, err := r.joinGroup(conn)
+	if err != nil {
+		return nil, err
+	}
+
+	assignments, err := r.syncGroup(conn, members)
+	if err != nil {
+		return nil, err
+	}
+
+	return assignments, nil
+}
+
+func (r *Reader) waitForTopic(conn *Conn) error {
+	const backoffDelayMin = 1 * time.Second
+	const backoffDelayMax = 30 * time.Second
+
 	for attempt := 0; true; attempt++ {
 		if attempt != 0 {
 			if !sleep(r.stctx, backoff(attempt, backoffDelayMin, backoffDelayMax)) {
-				return nil, r.stctx.Err()
+				return r.stctx.Err()
 			}
 		}
-
-		members, err := r.joinGroup(conn)
-		if err != nil {
-			return nil, err
+		_, err := conn.ReadPartitions(r.config.Topic)
+		if err == UnknownTopicOrPartition {
+			r.withErrorLogger(func(l *log.Logger) {
+				l.Printf("unknown topic or partition %q, retrying", r.config.Topic)
+			})
+			continue
+		} else if err != nil {
+			return err
+		} else {
+			break
 		}
-
-		assignments, err := r.syncGroup(conn, members)
-		if err != nil {
-			return nil, err
-		}
-
-		return assignments, nil
 	}
-	return nil, fmt.Errorf("failed all attempts to rebalance consumer group %v", r.config.GroupID)
+	return nil
 }
 
 func (r *Reader) unsubscribe() error {
@@ -860,6 +857,9 @@ func (r *Reader) handshake() error {
 		}
 		_ = conn.Close()
 	}()
+
+	// wait for topic to exist
+	r.waitForTopic(conn)
 
 	// rebalance and fetch assignments
 	assignments, err := r.rebalance(conn)
--- a/reader.go
+++ b/reader.go
@@ -158,6 +158,9 @@ func (r *Reader) refreshCoordinator() (err error) {
 
 		address, err := r.lookupCoordinator()
 		if err != nil {
+			r.withErrorLogger(func(l *log.Logger) {
+				l.Printf("coordinator lookup failed for group %v: %v", r.config.GroupID, err)
+			})
 			continue
 		}
 
@@ -237,6 +240,9 @@ type partitionReader interface {
 // assignTopicPartitions uses the selected GroupBalancer to assign members to
 // their various partitions
 func (r *Reader) assignTopicPartitions(conn partitionReader, group joinGroupResponseV1) (GroupMemberAssignments, error) {
+	const backoffDelayMin = 100 * time.Millisecond
+	const backoffDelayMax = 1 * time.Second
+
 	r.withLogger(func(l *log.Logger) {
 		l.Println("selected as leader for group,", r.config.GroupID)
 	})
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -827,6 +858,9 @@ func (r *Reader) handshake() error {
 		_ = conn.Close()
 	}()
 
+	// wait for topic to exist
```
stevevls: As mentioned in #200, I don't think this is a good place to insert the wait as it will cause the rebalance to time out if the topic doesn't exist.  I think that as you test it, you should be able to produce that condition fairly easily, but LMK if you find otherwise!

abuchanan-nr: Hrm, I'm confused. This is being called before rebalance, and therefore before joinGroup and syncGroup. Also, joinGroup is the only place I see config.RebalanceTimeout being used. There's no rebalance/join to timeout at this point, unless I'm missing something.

stevevls: This is a little bit subtle.  In a nutshell, there are three types of participants in consumer groups: the coordinator, the leader (which is a special consumer), and the consumers.  The coordinator is a Kafka broker whose responsibility is to manage group membership and coordinate rebalances.

When a rebalance occurs:
1. All the consumers phone in to the coordinator to join the group: https://github.com/segmentio/kafka-go/blob/d19f52f08673027cb346c63eb317b8e644b101af/reader.go#L303
1. Via the join group response, one of the consumers discovers that it has been appointed the leader and will be responsible for generating assignments: https://github.com/segmentio/kafka-go/blob/d19f52f08673027cb346c63eb317b8e644b101af/reader.go#L337
1. Upon successful join, all consumers *including the leader* call sync group. https://github.com/segmentio/kafka-go/blob/d19f52f08673027cb346c63eb317b8e644b101af/reader.go#L410  The leader will produce assignments and include them in the sync group request.  The other consumers leave that part of the request empty. https://github.com/segmentio/kafka-go/blob/d19f52f08673027cb346c63eb317b8e644b101af/reader.go#L369
1. The coordinator takes the leader's assignments and returns them to the other consumers via the sync group response.

So if you put in a blocking call before generating assignments, your leader won't ever send the sync group request with the assignments.  If that happens, the coordinator will eventually time out the sync, evict the leader from the group, and then run another rebalance.  It's not immediately obvious from reading the code that all this is happening, but you can read up on the protocol here: https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Client-side+Assignment+Proposal.

As with the case of more consumers than partitions, it's not a protocol level error to request a non-existent topic...it's a perfectly valid state.  It's kind of odd to me that the CG protocol doesn't provide a way to flag that as an error, but that's not for us mere mortals to debate. 😉   Anyway, that's why I was recommending to return empty assignments and then leverage the topic watcher feature that's already built in to trigger the rebalance when the topic appears.  Does that make sense?

ShaneSaww: @stevevls What about moving the wait for topic logic earlier? 
Maybe inside of the run func? 
Like here https://github.com/segmentio/kafka-go/blob/master/reader.go#L867

From my understanding there isn't any reason to move into the handshake logic if the topic doesn't exist. By moving it higher we can also reduce the amount of times we can block and reduce the overall amount of network calls made.

stevevls: @ShaneSaww That's an interesting idea, though I wonder how it would handle the case where you have a large group of consumers all waiting for a topic to exist.  The thing is, their wait times won't be synchronized, so you will have an initial flurry of rebalances as the consumers discover the existence of the topic and join the group.

Contrast that to where the group has already gone through the handshake.  In that case, only a single rebalance would be triggered.  Furthermore, having gone through handshake, all the consumers will be in sync and heartebeating, so the first member to detect the existence of the topic will force the rebalance.

What do you think?

abuchanan-nr: I believe the broker will wait a few seconds for new consumers to show up before rebalancing, controlled by `group.initial.rebalance.delay.ms` which defaults to 3 seconds. So if the retry loop is short enough, you should only get one or two rebalances, right?

stevevls: I think the number of rebalances would depend on a couple of things.  First would be the setting for `group.initial.rebalance.delay.ms`, second would be the sleep time between polling to check the topic, and third would be the number of consumers in the group.  Depending on those three inputs, you easily have more than 2 rebalances...especially if the number of consumers is large and there's a big difference between the rebalance delay and the sleep on the topic loop.  Those timeouts are configurable (or should be if they're not), so we should be careful about making assumptions about their values.

The more we talk this out, the more convinced I am that we the only way to _guarantee_ a single rebalance is to succeed the initial consumer group rebalance when the topic doesn't exist and then trigger a rebalance when it comes into existence.

As a side note, kafka-go doesn't currently support the `Reader` using consumer groups to pull messages from multiple topics.  However, the Kafka protocol allows for it.  If we were to implement that feature later, I could imagine that if we ended up in case where a group asks to consume two topics and only one exists, we wouldn't want to block consuming that one until the other one comes into existence.  I think that's another reason it would be preferable to create a consumer group for the non-existent topic.

I realize that all said and done, it's a weird case to be trying to consume a topic that doesn't exist, and this is a lot of thought going into what's an edge case.  But I think that it's best to handle edge cases as gracefully as possible, so I'd propose we go with what I suggested.  If you have strong objections, then let's keep talking through them!

ShaneSaww: @stevevls I would deal with that by having the wait time between calls for checking the topic to be fixed (and with what @abuchanan-nr  said ideally less then `group.initial.rebalance.delay.ms`).  That way the rebalance jitter _should_ be similar to just starting all the consumers without the topic checking. 
I don't think having them all ask for the same topic info is too hard on the brokers.  

The other way makes sense to me as well, it feels a little strange as the coordinator will technically get an error when trying to  get the topic info to build the assignments.

ShaneSaww: @stevevls  I didn't even think about the double topic case. That's an interesting one, and I totally see the validity in moving the check into the partitionWatcher in that case. 

Add adding all that config + trying to get the numbers correct to minimize rebalances makes the API quite confusing.

stevevls: @ShaneSaww @abuchanan-nr I started to get curious about how other libraries handle the case of unknown topics, so I looked at Sarama and librdkafka.

Sarama handles it like kafka-go currently does--if the topic doesn't exist, the CG leader throws an error and the rebalance goes into a failure loop.  Not terribly desirable.  

AFAICT, librdkafka ignores unknown topics at assignment time: https://github.com/edenhill/librdkafka/blob/5140cd9d965406a144de067b775bc9ca5255aea1/src/rdkafka_cgrp.c#L924.  But it does succeed the rebalance.  I would appreciate another set of eyes on that code b/c I'm not a C++ or librdkafka expert. 😄 Either way, I think we should mimic whatever librdkafka is doing.

On another note, I think we've gotten into the weeds a bit with handling the unknown topic.  One thing we can consider is merging in the fix for number consumers > number of partitions and then solve the unknown topic case in a separate PR.  Up to you!

ShaneSaww: @stevevls I can look into the java code as well.

reader.go
```diff
@@ -436,7 +439,9 @@ func (r *Reader) syncGroup(conn *Conn, memberAssignments GroupMemberAssignments)
 
 	if len(assignments.Topics) == 0 {
 		generation, memberID := r.membership()
-		return nil, fmt.Errorf("received empty assignments for group, %v as member %s for generation %d", r.config.GroupID, memberID, generation)
+		r.withLogger(func(l *log.Logger) {
```
WillAbides: If I'm reading this correctly, this is the only change required to address #200.

Could we make this a ReaderOption `AllowEmptyAssignments`? That would allow the option of preserving existing behavior for the non-existent topic case while allowing us to move forward with allowing more consumers than topics.

abuchanan-nr: There are a few downsides to consider there:
- It would be public, and therefore hard to get rid of.
- I'd argue it should default to true, to avoid a risky, subtle, and possibly common situation. 
- Too much config is a bad thing, IMO.

stevevls: I prefer not to have a configuration option.  Currently, the code is incorrect because there are valid use cases and configurations that _should_ work but don't.  The downside of allowing empty assignments is that perhaps someone accidentally spins up too many consumers and some number of them sit idle.  Preventing things like that feels like it's outside the scope of this library.

stevevls:
@abuchanan-nr I'm going to re-float the idea of splitting this into two PRs. 😄  What do you think about scoping this one down to just the problem of num consumers > num partitions?

reader.go
abuchanan-nr: Thanks for all the great discussion.

I think it's important to move forward here. There are some important cases mentioned in #200 that could be very common. 

We can revisit to optimize rebalances for unknown topics in future work. I agree it's important, but might take awhile.

We can either use this PR as-is, or I could move the `waitForTopic` into `run` where it would have its own connection that is not associated with a group coordinator.

Thoughts?

abuchanan-nr:
> I'm going to re-float the idea of splitting this into two PRs.

So this PR would remove the `waitForTopic` stuff?

stevevls:
Yeah...deferring the `waitForTopic` functionality to another PR is an option.  Everyone seems in agreement on allowing the assignments to be empty, and that fixes one immediate problem.  I think there's still some discussion to be had on the correct approach for `waitForTopic`, so splitting the PR would allow us to fix one issue while we make progress the another.

reader.go
stevevls: @ShaneSaww Did you get a chance to see how the java client works?  I'm still of the opinion that we should take the librdkafka approach where a non-existent topic simply results in empty assignments, but I think it's most important to behave like other official libraries in case there are places where folks need to inter-operate between languages.

abuchanan-nr:
Works for me. 

The only thing holding me up is figuring out how to write a test for this. Only idea so far: sleep long enough for rebalances to occur, then check ReaderStats. But, it would need to sleep for a relatively long time (~30 seconds).

reader.go
WillAbides: Those are good arguments for not making it an option.  Consider my suggestion withdrawn.

stevevls:
Yeah...that's a tough one...in part because we're trying to prove a negative (the rebalance *doesn't* churn).  I looked through the code to see if there was any internal state of the `Reader` that we might be able to poke at, but nothing jumped out.  Another thing that came to mind, though, was you could create a topic with a single partition, spin up a bunch of reader (say 10?), then publish messages at a fixed rate and ensure the same reader received all of them.  Sounds complicated, though, and also time-dependent.  The rebalance idea may be a simpler!

abuchanan-nr:
Created https://github.com/segmentio/kafka-go/pull/242 for the more minimal fix in order to avoid nuking any comments going in this thread.

Feel free to close this PR if you want.

reader.go
ShaneSaww: Yeah, from what I can tell with the java client it does something like 
- start everything as normal
- wait for metadata refresh 
- if metdata refresh shows a topic that you care about rebalance

stevevls:
Closing this, but I opened #251 to make sure we eventually follow up.  Thanks!

PR #232, opened 2019-03-14
writer: make the force flush timer accurate

twmb:
Prior, every writer to a partition had its own timer that ticked at
1/10th the BatchTimeout interval. Each tick, the code would check
whether BatchTimeout has elapsed since the prior flush and, if so,
trigger a flush.

This change makes the timer more accurate so that it only needs to tick
once every BatchTimeout interval, and only needs to begin a timeout
on the first message for a batch.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -568,14 +568,15 @@ func (w *writer) withErrorLogger(do func(*log.Logger)) {
 func (w *writer) run() {
 	defer w.join.Done()
 
-	ticker := time.NewTicker(w.batchTimeout / 10)
-	defer ticker.Stop()
+	batchTimer := time.NewTimer(0)
+	<-batchTimer.C
+	batchTimerRunning := false
+	defer batchTimer.Stop()
 
 	var conn *Conn
 	var done bool
 	var batch = make([]Message, 0, w.batchSize)
 	var resch = make([](chan<- error), 0, w.batchSize)
-	var lastFlushAt = time.Now()
 
 	defer func() {
 		if conn != nil {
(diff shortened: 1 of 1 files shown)


writer.go
```diff
@@ -595,13 +596,24 @@ func (w *writer) run() {
...
+			mustFlush = true
+			forceBatchRunning = false
 		}
 
 		if mustFlush {
-			lastFlushAt = time.Now()
+			if forceBatchRunning {
+				stopped := forceBatch.Stop()
+				forceBatchRunning = false
+				if !stopped {
```
stevevls: I would prefer to see this line moved up (and possibly even collapsed into `if stopped := forceBatch.Stop(); !stopped`) so that we put all the logic for timer management together and then toggle the `forceBatchRunning` boolean.  I think it's easier to reason about if we don't mix the two.

twmb: Field moved up & all uses of `forceBatch` have been changed to `batchTimer`.

achille-roussel:
The reason I initially built it this way was to avoid recreating timers on every message (by calling Reset), and avoid the complexity of optimizing that by tracking whether the timer needs to be reset or not (so exactly what this PR does 😉).

I'm not against the change tho if it helps to have more accurate flush intervals 👍

twmb:
Heh, well... doing 10x fewer time operations ultimately should be more efficient.

Both ticker and timer do the same select/default case to notify a consumer of a tick/timery expiry.

Each tick sifts a timer down the timer heap to reset it. Timer's just don't get automatically sifted back down.

Timer's Stop removes the timer from the heap and does not re-push back to heap. Reset only removes from the heap after a quick check that it's still present, and then does push back to the heap.

Long story short, careful resets should result in fewer heap operations.

stevevls:
Thanks @twmb.  I had to give myself a crash course on the go timer implementation to make sure I fully understood the implications. :grin:

PR #234, opened 2019-03-15
Added compression support for RecordBatch

stevevls:


stevevls:
This just increased in scope a bit. 😁

write.go
```diff
@@ -295,19 +279,85 @@ func writeProduceRequestV2(w *bufio.Writer, codec CompressionCodec, correlationI
...
+
+	// topic array
+	writeArrayLen(w, 1)
+	writeString(w, topic)
+
+	// partition array
+	writeArrayLen(w, 1)
+	writeInt32(w, partition)
 
+	writeInt32(w, size)
```
achille-roussel: I'm growing more and more fond of the idea of having a sub-package for all protocol structs so we can try to better control this growing complexity of inlined serializers... work for another day!

stevevls: Totally.  It's only going to get worse now that we have API versions. 😄

PR #235, opened 2019-03-16
Add transactional id to producers

VictorDenisov:


VictorDenisov:
Second step for: https://github.com/segmentio/kafka-go/issues/216

conn.go
```diff
@@ -73,13 +73,20 @@ type Conn struct {
...
 // ConnConfig is a configuration object used to create new instances of Conn.
 type ConnConfig struct {
 	ClientID  string
 	Topic     string
 	Partition int
+
+	// The transactional id to use for transactional delivery. Idempotent
+	// deliver should be enabled if transactional id is configured.
+	// For more details look at transactional.id description here: http://kafka.apache.org/documentation.html#producerconfigs
+	TransactionalId *string
```
achille-roussel: This pointer-to-string model is not used anywhere else in kafka-go, it seems a bit out of place to introduce this.

Could we use the zero-value as a way to express that there is no transactional ID? Or is an empty string valid in this case?

VictorDenisov: From the description of transactional.id I infer that empty string is a valid value: http://kafka.apache.org/documentation.html#producerconfigs
TransactionalId is a nullable string: https://kafka.apache.org/protocol#protocol_types
We need to be able to distinguish between empty value and no value. It doesn't need to be a pointer, but then we need to decide how to communicate no value. We can probably use something like Optional(Maybe), but then it will be even more against the current design and not idiomatic to go.

achille-roussel: Screen Shot 2019-03-17 at 12 38 27 PM

The forth column describes "valid values", seems like even if it's represented as a nullable string on the wire it cannot be an empty string?

VictorDenisov: Indeed. I didn't pay attention to this column. I'll use empty string as default value.

VictorDenisov:
@achille-roussel done

conn.go
```diff
@@ -73,13 +73,21 @@ type Conn struct {
...
 type ConnConfig struct {
 	ClientID  string
 	Topic     string
 	Partition int
+
+	// The transactional id to use for transactional delivery. Idempotent
+	// deliver should be enabled if transactional id is configured.
+	// For more details look at transactional.id description here: http://kafka.apache.org/documentation.html#producerconfigs
+	// Empty string means that this connection can't be transactional.
+	TransactionalId string
```
achille-roussel:
```suggestion
	TransactionalID string
```
This would match the spelling of `ClientID` and follow common Go naming conventions.

conn.go
```diff
@@ -120,6 +128,13 @@ func NewConn(conn net.Conn, topic string, partition int) *Conn {
 	})
 }
 
+func transactionalIdToNullableString(transactionalId string) (result *string) {
```
achille-roussel: This seems more like a generic function to convert a `string` to a `*string`, how about a simpler name like
```suggestion
func nullableString(transactionalId string) (result *string) {
```
?

This is an implementation detail so feel free to keep it this way if you think it's more meaningful.

VictorDenisov: This name was my initial intention. I was concerned that it had preconditions not reflected in the name. I think `emptyToNullable` should suit it.

VictorDenisov:
done

conn.go
```diff
@@ -120,6 +128,13 @@ func NewConn(conn net.Conn, topic string, partition int) *Conn {
 	})
 }
 
+func emptyToNullable(transactionalID string) (result *string) {
+	if transactionalID != "" {
+		*result = transactionalID
```
achille-roussel: Isn't this dereferencing a `nil` pointer?
```suggestion
		result = &transactionalID
```

VictorDenisov: :man_facepalming: fixed

PR #240, opened 2019-03-19
Add isolation level config to reader

VictorDenisov:


message.go
```diff
@@ -372,15 +372,15 @@ func (h *messageSetHeaderV2) compression() int8 {
...
 }
 
 func (h *messageSetHeaderV2) transactionType() transactionType {
-	return transactionType(h.batchAttributes & 16)
+	return transactionType(h.batchAttributes & (1 << 4) >> 4)
 }
 
 func (h *messageSetHeaderV2) controlType() controlType {
-	return controlType(h.batchAttributes & 32)
+	return controlType(h.batchAttributes & (1 << 5) >> 5)
```
achille-roussel: While this change makes it more explicit that we're looking at extracting one bit of the value, it also introduces an implicit dependency on the order of operations to understand what is happening. Could we add parens to make it more explicit? For example:
```suggestion
	return controlType((h.batchAttributes & (1 << 5)) >> 5)
```

VictorDenisov:
Fixed by amending the latest commit.

PR #242, opened 2019-03-22
reader: allow empty partition assignments when joining a group

abuchanan-nr:
Follow up to #229


stevevls:
Looks great.  Thanks!

PR #244, opened 2019-03-29
Add new method ConsumerGroupOffsets to Reader

rjenkins:
This is a reviewer first PR and can be read commit by commit. In summary we expose a new method on Reader called `ConsumerGroupOffsets` that returns `map[int]int64, error`. The map contains a mapping of partition to the last committed offset for the Readers consumer group ID.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -100,6 +100,8 @@ type Reader struct {
 	// reader stats are all made of atomic values, no need for synchronization.
 	// Use a pointer to ensure 64-bit alignment of the values.
 	stats *readerStats
+	// A cache of our current assignments
+	assignments map[string][]int32
 }
 
 // useConsumerGroup indicates whether the Reader is part of a consumer group.
@@ -463,6 +465,9 @@ func (r *Reader) rebalance(conn *Conn) (map[string][]int32, error) {
 	if err != nil {
 		return nil, err
 	}
+	r.mutex.Lock()
+	r.assignments = assignments
+	r.mutex.Unlock()
 
 	return assignments, nil
 }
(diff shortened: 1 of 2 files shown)


reader.go
```diff
@@ -1422,6 +1427,29 @@ func (r *Reader) ReadLag(ctx context.Context) (lag int64, err error) {
 	return
 }
 
+func (r *Reader) ConsumerGroupOffsets() (map[int]int64, error) {
+	if !r.useConsumerGroup() {
+		return nil, fmt.Errorf("only supported for consumer groups")
+	}
+
+	// establish a connection to the coordinator.  this connection will be
```
stevevls: This is nitpicky, but I'd prefer to see the connection to the coordinator below where you actually use it.

rjenkins: @stevevls can do, just to be clear though there is a small trade-off there, that is if the `r.coordinator()` errors we would needlessly lock the reader to make the copy (which won't get used). Total minor nit as well, but just to be clear. If you're still +1 I'll move it down.

stevevls: Yeah, I dig that.  My hope would be that no one is looping on this so quickly as to cause contention on the lock. 🤞 If they are, then that should be the least of their worries.  I'll leave it up to you since my suggestion was for readability reasons.

reader.go
```diff
@@ -1422,6 +1427,29 @@ func (r *Reader) ReadLag(ctx context.Context) (lag int64, err error) {
 	return
 }
 
+func (r *Reader) ConsumerGroupOffsets() (map[int]int64, error) {
```
achille-roussel:
```suggestion
func (r *Reader) GroupOffsets() (map[int]int64, error) {
```
We don't use the term `ConsumerGroup` in the API currently, but we do have `GroupID` in the reader configuration, this could keep the naming a bit more consistent.

Issue #243, opened 2019-03-27
Reconnect with Backoff Config


VictorDenisov:
**Describe the solution you'd like**
Java clients have infinite reconnect policy even if kafka brokers refuse connections completely. Java clients keep trying to reconnect with configurable backoff.ms timeout.  I suggest to implement similar behavior in this client as it's naturally expected when people transition from other kafka clients.

**Supporting documentation**
https://kafka.apache.org/documentation/#consumerapi

VictorDenisov:
The story consists of two items:
- [ ] allow infinite attempts in Writer. -1 in MaxAttempts should be infinite number of attempts.
- [ ] add retry logic to Reader.

VictorDenisov:
I posted a PR for the second item: https://github.com/segmentio/kafka-go/pull/264
This issue can also be resolved by retrying manually:
```
for ;; {
  m, err = reader.ReadMessage()
  if err == nil {
    break;
  }
}

PR #245, opened 2019-03-29
Use -1 as a value for infinite number of writer attempts

VictorDenisov:


VictorDenisov:
Fix for writer from ticket: https://github.com/segmentio/kafka-go/issues/243

VictorDenisov:
@achille-roussel Could you please take a look at this PR. This fix is rather critical for me. Thanks in advance.

VictorDenisov:
@stevevls maybe you can comment?

writer.go
```diff
@@ -58,6 +58,7 @@ type WriterConfig struct {
 	Balancer Balancer
 
 	// Limit on how many attempts will be made to deliver a message.
+	// -1 means that we can make unlimited number of attempts.
```
stevevls: Could we make this a constant value instead?

VictorDenisov:
Indeed MaxInt retries should be enough for any practical situation. Sorry for the disturbance. I'll close the PR.

Issue #241, opened 2019-03-22
NewReader and NewWriter shouldn't panic


abuchanan-nr:
NewReader panics instead of returning an error: https://github.com/segmentio/kafka-go/blob/master/reader.go#L1051

As does NewWriter: https://github.com/segmentio/kafka-go/blob/master/writer.go#L178

Some people might want to log errors differently, others might not be creating readers/writers in main, and a panic might take their whole service down. A more idiomatic Go API would be to return an error, allowing the caller to decide what to do when NewReader fails.

PR #246, opened 2019-03-31
Panics have changed with Errors for NewWriter And NewReader

olimpias:
Errs are handled more gracefully.
Unit tests are added for NewWriter and NewReader.

Ref Task: #241

### DIFF
--- a/compression_test.go
+++ b/compression_test.go
@@ -79,7 +79,7 @@ func testCompressedMessages(t *testing.T, codec kafka.CompressionCodec) {
 		t.Parallel()
 
 		topic := kafka.CreateTopic(t, 1)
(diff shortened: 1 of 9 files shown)


achille-roussel:
First, thanks for the contribution!

I don't think we can make this change as proposed tho, because we would be breaking pretty much all programs using this package.

What do you think about taking a different approach, and introducing a mechanism for validating the configuration values that are passed to reader/writer, for example:

```go
func (c *ReaderConfig) Validate() error {
    ...
}
```

Then the program can explicitly call `Validate` to detect malformed configuration before it creates the reader, and we can leave the signature of `NewReader`/`NewWriter` unchanged.

What do you think?

olimpias:
Sure, it makes sense for backward compatibility.

PR #249, opened 2019-04-03
Fix full buffer errors.

wjywbs:
bufio.Reader.Peek() returns ErrBufferFull if the peek size is larger than the buffer size. Grow the provided slice up to cap.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -120,8 +120,11 @@ func (batch *Batch) Read(b []byte) (int, error) {
 				return size, nil
 			}
 			n = nbytes // return value
+			if nbytes > cap(b) {
+				nbytes = cap(b)
+			}
 			if nbytes > len(b) {
-				nbytes = len(b)
+				b = b[:nbytes]
 			}
 			nbytes, err := io.ReadFull(r, b[:nbytes])
 			if err != nil {
--- a/message.go
+++ b/message.go
@@ -463,6 +463,7 @@ func (r *messageSetReaderV2) readMessage(min int64,
 				err = errShortRead
 				return
 			}
+			r.reader = bufio.NewReaderSize(r.reader, batchRemain)
 			var b []byte
 			if b, err = r.reader.Peek(batchRemain); err != nil {
 				return
--- a/message.go
+++ b/message.go
@@ -463,7 +463,6 @@ func (r *messageSetReaderV2) readMessage(min int64,
 				err = errShortRead
 				return
 			}
-			r.reader = bufio.NewReaderSize(r.reader, batchRemain)
 			var b []byte
 			if b, err = r.reader.Peek(batchRemain); err != nil {
 				return


stevevls:
Thanks for the PR!  Could you give a little more info about the problem you're working around?  The documentation on this function states that 
> // The method fails with io.ErrShortBuffer if the buffer passed as argument is
> // too small to hold the message value.

So the code is currently acting correctly, and I would worry about breaking backwards compatibility if we changed the behavior.  I think understanding your use case better will help us figure out what to do.

wjywbs:
I'm using compression where each record size is large. r.reader's buffer is only 4096 bytes, and the record size (batchRemain) is more than 20000 bytes. So r.reader.Peek(batchRemain) will fail with ErrBufferFull. Then the code will retry the offset again which causes infinite loop on the record, for a simple reader use case.

https://github.com/segmentio/kafka-go/blob/master/batch.go#L137

For the batch.go change, I'm using `b := make([]byte, 0, 10e3); n, err := batch.Read(b)` instead of `b := make([]byte, 10e3)`. The change will grow the slice up to cap to fill the data, and I use `b = b[:n]` to grow the slice after batch.Read().

batch.go
```diff
@@ -120,8 +120,11 @@ func (batch *Batch) Read(b []byte) (int, error) {
 				return size, nil
 			}
 			n = nbytes // return value
+			if nbytes > cap(b) {
```
stevevls: I think it will read more clearly if we move this check down into the if block for the `len` check and to re-slice only in the case where `len < cap`

message.go
```diff
@@ -463,6 +463,7 @@ func (r *messageSetReaderV2) readMessage(min int64,
 				err = errShortRead
 				return
 			}
+			r.reader = bufio.NewReaderSize(r.reader, batchRemain)
```
stevevls: This could have unanticipated side effects for allocations...will need to think on this for a bit.

stevevls:
Thanks for the follow-ups!  I think we can merge the change to take advantage of slice capacity with some minor changes.  Will have to think a little bit more on what the correct behavior is in the case where you're reading the compressed message.

message.go
wjywbs: The document mentions that "If the argument io.Reader is already a Reader with large enough size, it returns the underlying Reader." I think this shouldn't be an issue if the record sizes are within a range.

https://golang.org/pkg/bufio/#NewReaderSize

stevevls: Have a look at #250.  I believe it solves the same underlying issue, but it avoids an allocation and a copy in the case where the buffer needs to grow.  Sorry to be so slow...some of this code was recently contributed, so I'm not intimately familiar with it yet. 😅

wjywbs: Thanks for your fix. I tested and it works. The turn around time is already very short.

batch.go
wjywbs: Do you mean:

```
if nbytes > len(b) {
  if nbytes > cap(b) {
    nbytes = cap(b)
  }
  if nbytes > len(b) {
    b = b[:nbytes]
  }
}
```

wjywbs: Do you mean:

```
if nbytes > len(b) {
  if nbytes > cap(b) {
    nbytes = cap(b)
  }
  if nbytes > len(b) {
    b = b[:nbytes]
  }
}
```

stevevls: Thanks for checking back.  Now that I see this code, I prefer what you originally had.  Apologies!

message.go
stevevls: Great!  I'll merge it down then.

stevevls:
Thanks for the contribution!

PR #253, opened 2019-04-05
Add zstd compression

Pryz:
This PR adds the zstd compression algorithm available since Kafka 2.1.0 with the required internal fetch and produce request versions.

Note about this PR : in this PR we are using github.com/DataDog/ztsd which is a Go wrapper around the C implementation of ztsd. Due to that using zstd will require cgo which is quite unfortunate.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -70,9 +70,10 @@ type Conn struct {
 	correlationID int32
 
 	// number of replica acks required when publishing to a partition
-	requiredAcks int32
-	apiVersions  map[apiKey]ApiVersion
-	fetchVersion apiVersion
+	requiredAcks   int32
+	apiVersions    map[apiKey]ApiVersion
+	fetchVersion   apiVersion
+	produceVersion apiVersion
 
 	transactionalID *string
 }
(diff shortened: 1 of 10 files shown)


zstd/zstd.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Code implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Code() int8 {
+	return Code
+}
+
+// Encode implements the kafka.CompressionCodec interface.
+func (c CompressionCodec) Encode(src []byte) ([]byte, error) {
+	return zstd.Compress(nil, src)
```
Pryz: I forget the use the compression level here. Will fix that tomorrow.

read.go
```diff
@@ -459,6 +459,110 @@ func readFetchResponseHeaderV5(r *bufio.Reader, size int) (throttle int32, water
 
 }
 
+func readFetchResponseHeaderV10(r *bufio.Reader, size int) (throttle int32, watermark int64, remain int, err error) {
```
stevevls: I worry a bit about the amount of copy/paste happening here.  Not sure that I have an answer now for how to fix it, but I'd worry about bugfixes making it to all the right places.

Pryz: Yes, I'm also pretty worried about this getting hard to maintain as well. Especially if we keep adding more and more versions.

conn.go
```diff
@@ -725,6 +733,19 @@ func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
 		deadline = adjustDeadlineForRTT(deadline, now, defaultRTT)
 		adjustedDeadline = deadline
 		switch c.fetchVersion {
```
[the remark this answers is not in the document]
Pryz: indeed. I forgot to change https://github.com/segmentio/kafka-go/blob/53ce538e2da8f858dc1958d5ed1bfd99ee5ee8cd/conn.go#L191. Interesting that the tests are passing tho :D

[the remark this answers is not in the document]
Pryz: fixed !

read.go
Pryz: We could find a way to factorize this code but I'm worried about adding too much complexity since the Kafka protocol can add fields in the nested structures.

PR #256, opened 2019-04-09
Added clarifying docs on Writer batching behavior

### LINKED ISSUE #92: Document the working of QueueCapacity/BatchSize/BatchTimeout
I've been using the `Writer` with a deadlined `Context` and I'm unsure how to configure it to allow batching while not blocking for the whole deadline duration.

If I set a `BatchSize` to 1, all is well but the load on the brokers and the network traffic is considerable. It's greatly reduced by setting `BatchSize` to another number - like 2, however `WriteMessages` blocks for the duration of the deadline (300ms here)

capture d ecran 2018-04-19 a 12 56 30

I believe my understanding of the documentation is wrong?

```
Unless the writer was configured to write messages asynchronously, the method blocks until all messages have been written, or until the maximum number of attempts was reached.
```

There's about 1.3k `WriteMessages` per second in this test, with `BatchSize` set to 2 and `RequiredAcks` set to 1.

stevevls:
Fixes #92

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -260,6 +260,15 @@ func NewWriter(config WriterConfig) *Writer {
 // blocks until all messages have been written, or until the maximum number of
 // attempts was reached.
 //
+// When sending synchronously and the writer's batch size is configured to be
+// greater than 1, this method blocks until either a full batch can be assembled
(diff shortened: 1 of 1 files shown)


stevevls:
Oh yeah...very confidence inspiring. 😉

PR #259, opened 2019-04-16
Add ReadTopics

Pryz:
This PR adds a `ReadTopics` function which retrieves all the topics available on the current connection.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -27,6 +27,12 @@ type Broker struct {
 	Rack string
 }
 
+// Topic carries the metadata associated with a kafka topic.
+type Topic struct {
+	Topic      string
+	Partitions []Partition
+}
+
 // Partition carries the metadata associated with a kafka partition.
 type Partition struct {
 	Topic    string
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -953,6 +959,68 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
...
+				for i, id := range ids {
+					b[i] = brokers[id]
+				}
+				return b
+			}
+
+			topics = make([]Topic, 0, len(res.Topics))
+			for _, t := range res.Topics {
+				if t.TopicErrorCode != 0 && (c.topic == "" || t.TopicName == c.topic) {
+					// We only report errors if they happened for the topic of
```
stevevls: This error handling seems a little weird to me.  When would we have this case and why would we want to return the empty partitions slice?  It feels like we're losing information and hiding things from the caller.

conn_test.go
```diff
@@ -981,6 +985,38 @@ func testBrokers(t *testing.T, conn *Conn) {
 	}
 }
 
+func testReadTopics(t *testing.T, conn *Conn) {
+	time.Sleep(2 * time.Second)
```
stevevls: What is this sleep playing at?

metadata.go
```diff
@@ -9,7 +9,11 @@ func (r topicMetadataRequestV1) size() int32 {
 }
 
 func (r topicMetadataRequestV1) writeTo(w *bufio.Writer) {
-	writeStringArray(w, []string(r))
+	if len([]string(r)) > 0 {
```
stevevls: Heh heh...in some places, we get by without this check when serializing arrays b/c kafka usually don't differentiate between null and empty.

conn_test.go
Pryz: gonna remove that. Tests were failing locally because of my local kafka state.

metadata.go
Pryz: interesting. gonna test without but it was failing for me.

stevevls: Oh, I believe that it was failing.  Just saying that Kafka is inconsistent on this front!

achille-roussel:
@Pryz is this still needed?

stevevls:
Should we close this PR?

furkanyilmazx:
I need this feature.

Pryz:
@furkanyilmazx : you can get the same result with https://godoc.org/github.com/segmentio/kafka-go#Conn.ReadPartitions.

furkanyilmazx:
> @furkanyilmazx : you can get the same result with https://godoc.org/github.com/segmentio/kafka-go#Conn.ReadPartitions.
> 
> @stevevls : yes but I will need to fix the ReadPartitions so it works when we don't pass any topic first. Will try to tackle that this week.

I think, 'ReadPartitions' doesn't suitable method to read topics. And for now, either it doesn't work or i couldn't deal with it  :)

achille-roussel:
@furkanyilmazx can you describe the issue you were seeing with `ReadPartitions`? Maybe share a code sample that we can use to track down what's wrong?

achille-roussel:
I think the issue may have been fixed by https://github.com/segmentio/kafka-go/pull/383

I'll close, but feel free to re-open if you're still having issues with `Conn.ReadPartitions`.

PR #260, opened 2019-04-19
handle empty batch message set

abuchanan-nr:
We have some topics that are low traffic. On topics like this, we are getting a lot of error messages, such as `the kafka reader got an unknown error reading partition 17 of some_topic_name at offset 577667: unexpected EOF`

This error originates here: https://github.com/segmentio/kafka-go/blob/master/conn.go#L806

When there are no messages to be read, kafka returns an empty response body, and `highWaterMark == currentOffset`, however `newMessageSetReader` treats this as a bad response.

This change modifies `messageSetReader` to allow it to be empty, and check for this case.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -803,7 +803,11 @@ func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
 
 	var msgs *messageSetReader
 	if err == nil {
-		msgs, err = newMessageSetReader(&c.rbuf, remain)
+		if highWaterMark == offset {
+			msgs = &messageSetReader{empty: true}
+		} else {
+			msgs, err = newMessageSetReader(&c.rbuf, remain)
+		}
 	}
 	if err == errShortRead {
 		err = checkTimeoutErr(adjustedDeadline)
--- a/message.go
+++ b/message.go
@@ -108,6 +108,7 @@ func (s messageSet) writeTo(w *bufio.Writer) {
 }
 
 type messageSetReader struct {
+	empty   bool
 	version int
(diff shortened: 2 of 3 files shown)


abuchanan-nr:
>  I wonder if instead you could make the Batch encapsulate the behavior by giving it a nil messageSetReader and checking for that in readMessage

We explored various options, including returning nil. In my opinion, returning (nil, nil) is more likely to result in an unchecked nil panic down the road, which is why I ruled it out. The concept of having an empty message set seems natural, but I admit the implementation is very awkward because of the existing version handling in messageSetReader. A better solution would be to replace messageSetReader with an interface, but I wanted to keep this minimal.

Anywho, happy to change it if you're unconvinced :)

Unit test on the way soon.

stevevls:
Sounds reasonable!  I think either way reads a bit awkward, but let's run with the code as-is to avoid the potential panic case you mentioned.

The `messageSetReader` as interface idea is actually something that we discussed but decided against as a memory optimization: https://github.com/segmentio/kafka-go/pull/146#discussion_r236064809.

stevevls:
@abuchanan-nr Wanted to check in since this PR has been inactive for a bit.  We're just waiting on a unit test to exercise the empty batch code, yeah?

abuchanan-nr:
Ya, sorry I got distracted. I'll finish this up this weekend or Monday.

abuchanan-nr:
Added a test for an empty message set reader.

stevevls:
Excellent...thanks for the contribution!

Issue #251, opened 2019-04-04
Consumer Groups gracefully handle non-existent topics


stevevls:
**Describe the solution you'd like**
Currently, the consumer group rebalance will error out if a topic does not exist.  Official Kafka clients (librdkafka, java) do not throw an error in this case.  Instead, they hand back empty assignments to all the consumers.  If the topic comes into existence later, then a rebalance is initiated.

It would be great if kafka-go had the same behavior.  See background and discussion in #200 and #229

PR #262, opened 2019-04-26
Don't fail consumer group assignment if topic does not exist

stevevls:
Official Kafka clients (librdkafka, java) do not throw an error in
this case. Instead, they hand back empty assignments to all the
consumers. The partition watch functionality can be used to trigger
a rebalance if the topic comes into existence later.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -257,7 +257,7 @@ func (r *Reader) assignTopicPartitions(conn partitionReader, group joinGroupResp
 	// it's not a failure if the topic doesn't exist yet.  it results in no
 	// assignments for the topic.  this matches the behavior of the official
 	// clients: java, python, and librdkafka.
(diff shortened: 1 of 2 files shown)


stevevls:
Fixes #251

reader.go
```diff
@@ -253,7 +253,12 @@ func (r *Reader) assignTopicPartitions(conn partitionReader, group joinGroupResp
 
 	topics := extractTopics(members)
 	partitions, err := conn.ReadPartitions(topics...)
-	if err != nil {
+
+	// it's not a failure if the topic doesn't exist yet.  it results in no
+	// assignments for the topic.  this matches the behavior of the official
+	// clients: java, python, and librdkafka.
+	// ana topic watcher can trigger a rebalance when the topic comes into being.
```
Pryz: `ana topic` -> typo ? :)

PR #268, opened 2019-05-08
expose batch error

achille-roussel:
This PR adds `func (*Batch) Err() error`, see the documentation for detailed explanations about the use cases that this new API solves for.


achille-roussel:
Thanks for the reviews!

Issue #181, opened 2019-01-22
Question: best way to get offsets for partition?


abuchanan-nr:
I need to write tests which check that the offsets for partitions are what I expect. At first I just called `Conn.ReadOffsets`, but I get an error `Unknown Topic Or Partition: ...`.

What is the best way to get a topic/partition-specific connection in order to read offsets? The Dial API has a lot of variety: some take a net.Conn, some a Partition, some a topic string + partition int, etc.

Sorry, I think I'm probably missing something obvious.

abuchanan-nr:
I should clarify, I'm testing the behavior of consumer groups, so the offsets should be group + partition specific.

achille-roussel:
This is not a feature we currently expose in the public API, but if you're available to contribute a change we'll be happy to merge it!

rjenkins:
hiya @abuchanan-nr just opened https://github.com/segmentio/kafka-go/pull/244 to address this issue. Could you have a look at the changes and the tests and see if it addresses your request?

abuchanan-nr:
@rjenkins My request was a little different. I'd like to read the state of consumer groups, without  creating/joining a group. That would likely end up on the `Conn` type.

rjenkins:
@abuchanan-nr ok I'm going to keep this change because I think it'll be useful as well on the readers, I'll talk to @achille-roussel and see what he thinks about adding it to the `Conn` type as well.

rjenkins:
@abuchanan-nr, spoke with @achille-roussel and we're still sorting out where this should live. We have ideas for some other utility funcs as well and we might want to consolidate these. I'll have an update by the end of the week.

abuchanan-nr:
Fantastic, thanks @rjenkins!

rjenkins:
@abuchanan-nr we've decided we're going to expose a new api/interface for these types of operations, planning on getting something up for review this weekend.

rjenkins:
Finally had some time to revisit this. https://github.com/segmentio/kafka-go/pull/272, we will most likely close https://github.com/segmentio/kafka-go/pull/244 once 272 lands.

rjenkins:
Resolved in https://github.com/segmentio/kafka-go/pull/272

PR #272, opened 2019-05-09
Introduce Client API

rjenkins:
Client is a new API. It is an experimental API and it is expected to grow over time. We intend the API to be higher level than `Conn` but provide more lower level functions than `Reader` or `Writer`. The first provided feature is `ConsumerOffsets` in support of issue https://github.com/segmentio/kafka-go/issues/181. Both `NewClient` and `NewClientWith` require a bootstrap list of brokers. The `*ClientWith` func allows for a user defined Dialer to be passed. The non public funcs `connect`, `lookupCoordinator` and `coordinator` are similar to funcs in Reader. I wasn't sure what the projects preferences are here (DRY vs. some level of duplication is OK for separation of concerns). so I've just included new versions of these funcs on the Client. If we decide it would be better to move those to another type that both Reader and Client can compose we can refactor to do so.


abuchanan-nr:
I suggest following the existing pattern of passing `ClientConfig` (or `ClusterClientConfig`) to the constructor, which both the reader and writer do. Over time, config for client code tends to grow.

Also, is it worth planning out a `Client` type? `ClusterClient` feels verbose.

Looking forward to having this though!

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+		dialer:  DefaultDialer,
+	}
+}
+
+// NewClusterClient creates and returns a *ClusterClient with a []string of bootstrap
+// brokers for connecting to the cluster and a user supplied Dialer
+func NewClusterClientWith(brokers []string, d *Dialer) *ClusterClient {
+	return &ClusterClient{
+		brokers: brokers,
+		dialer:  d,
```
achille-roussel: If we change it to take a config type, let's check if the dialer is `nil` and automatically assume `DefaultDialer` in that case?

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+		brokers: brokers,
+		dialer:  DefaultDialer,
+	}
+}
+
+// NewClusterClient creates and returns a *ClusterClient with a []string of bootstrap
+// brokers for connecting to the cluster and a user supplied Dialer
+func NewClusterClientWith(brokers []string, d *Dialer) *ClusterClient {
+	return &ClusterClient{
+		brokers: brokers,
```
achille-roussel: Let's make a copy of the list of brokers so we don't risk sharing memory that may get reused and mess up the internal configuration of the client.

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+		dialer:  d,
+	}
+}
+
+// ConsumerOffsets returns a map[int]int64 of partition to committed offset for a consumer group id and topic
+func (cc *ClusterClient) ConsumerOffsets(ctx context.Context, groupId string, topic string) (map[int]int64, error) {
```
achille-roussel: Should it be `topic string, groupId string` ? 😅  or should we have an API that helps programs not mistakenly invert the arguments maybe?

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	offsetsByPartition := map[int]int64{}
+	for _, pr := range offsets.Responses[0].PartitionResponses {
+		offset := pr.Offset
+		if offset < 0 {
+			// No offset stored
+			offset = FirstOffset
```
achille-roussel: Could we link to the protocol documentation and explain why we can't just use `offset` when it's negative? (reading this I'm assuming it has an undefined value, or at least it's not FirstOffset).

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+			},
+		},
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	offsetsByPartition := map[int]int64{}
+	for _, pr := range offsets.Responses[0].PartitionResponses {
```
achille-roussel: Is it possible to get no responses and no errors?

clusterclient.go
rjenkins: I'm actually not sure about that, this implementation is based off the one in Reader and that's what it currently does.

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+
+// ClusterClient provides a high-level API for miscellaneous (i.e. non consume or produce) Kafka cluster operations
+type ClusterClient struct {
+	brokers []string
+	dialer  *Dialer
+}
+
+// NewClusterClient creates and returns a *ClusterClient with a []string of bootstrap
+// brokers for connecting to the cluster and the DefaultDialer
+func NewClusterClient(brokers []string) *ClusterClient {
```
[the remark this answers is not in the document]
rjenkins: OK, addressed in latest

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+	return &ClusterClient{
+		brokers: brokers,
```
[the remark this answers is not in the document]
rjenkins: Addressed in latest

clusterclient.go
rjenkins: Done

clusterclient.go
```diff
@@ -0,0 +1,120 @@
...
+		dialer:  DefaultDialer,
+	}
+}
+
+// NewClusterClient creates and returns a *ClusterClient with a []string of bootstrap
+// brokers for connecting to the cluster and a user supplied Dialer
+func NewClusterClientWith(brokers []string, d *Dialer) *ClusterClient {
```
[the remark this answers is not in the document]
rjenkins: Done

clusterclient.go
rjenkins: Added comments

clusterclient.go
rjenkins: Addressed in latest

clusterclient.go
rjenkins: Done

clusterclient.go
rjenkins: done

PR #276, opened 2019-05-20
Add ReadBackoffMin/Max config to Reader (#269)

### LINKED ISSUE #269: Configure retry.backoff.ms in Reader
I have a need to configure the `backoffDelayMin` (and potentially `backoffDelayMax`) property that decides how tight the Readers read loop runs. This property is currently defaulted to [100 ms in the reader](https://github.com/segmentio/kafka-go/blob/master/reader.go#L1705-L1706), and backs off up to one second. Waiting even 30ms on my project would be too slow!

Issue #243 is similar but refers to the `reconnect.backoff.ms` property - I believe this issue is more similar to the [`retry.backoff.ms`](https://kafka.apache.org/documentation/#consumerapi) property.

I understand that the importance in the documentation is "low". But given that I will be altering my own fork anyway, I would be happy to PR it in here too if appropriate.

If so, should both `backoffDelayMin` and `backoffDelayMax` be configurable? Or just `delayMin`

ostenbom:
Fixes #269

Introduces configuration parameters for bounds on how often to poll for new messages in the reader.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -62,6 +62,11 @@ const (
 	// defaultPartitionWatchTime contains the amount of time the kafka-go will wait to
 	// query the brokers looking for partition changes.
 	defaultPartitionWatchTime = 5 * time.Second
+
+	// defaultReadBackoffMax/Min sets the boundaries for how long the reader wait before
+	// polling for new messages
+	defaultReadBackoffMin = 100 * time.Millisecond
+	defaultReadBackoffMax = 1 * time.Second
 )
 
 // Reader provides a high-level API for consuming messages from kafka.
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -1039,6 +1056,14 @@ func (config *ReaderConfig) Validate() error {
 		return errors.New(fmt.Sprintf("minimum batch size greater than the maximum (min = %d, max = %d)", config.MinBytes, config.MaxBytes))
 	}
 
+	if config.ReadBackoffMax < 0 || config.ReadBackoffMax < config.ReadBackoffMin {
+		return errors.New(fmt.Sprintf("ReaxBackoffMax out of bounds: %d", config.ReadBackoffMax))
```
stevevls:
```suggestion
		return errors.New(fmt.Sprintf("ReadBackoffMax out of bounds: %d", config.ReadBackoffMax))
```

What do you think of breaking this out into two separate checks and errors?   As an end user, I think that it would be easier to debug the issue where max < min.

I also think that it might be useful to do the max < min check *after* applying default values.  As currently written, I think this would error if I wanted to use the default for `ReadBackoffMax` but specified a non-zero value for `ReadBackoffMin`.

reader.go
```diff
@@ -1039,6 +1056,14 @@ func (config *ReaderConfig) Validate() error {
...
+	}
+
+	if config.ReadBackoffMin < 0 {
+		return errors.New(fmt.Sprintf("ReaxBackoffMin out of bounds: %d", config.ReadBackoffMin))
```
stevevls:
```suggestion
		return errors.New(fmt.Sprintf("ReadBackoffMin out of bounds: %d", config.ReadBackoffMin))
```

ostenbom:
Thanks for the feedback - I have taken it on board in the latest diff.

stevevls:
Great...thanks!

PR #280, opened 2019-05-29
auto split kafkaURL to array

zgfh:
auto split kafkaURL to array


achille-roussel:
Thanks for the contribution!

PR #289, opened 2019-06-07
Pad shorter version with zeroes

VictorDenisov:
Shorter(including empty) kafka versions are perceived as satisfying atLeast requirement.

### DIFF
--- a/testing/version.go
+++ b/testing/version.go
@@ -20,6 +20,11 @@ func (v semver) atLeast(other semver) bool {
 			return true
 		}
 	}
+	for i := len(v); i < len(other); i++ {
+		if 0 < other[i] {
+			return false
+		}
+	}
 	return true
 }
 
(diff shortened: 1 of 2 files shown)


stevevls:
Thanks @VictorDenisov.  I have two questions:
1. Is the intention that if `KAFKA_VERSION` is unspecified, we will only run tests known to work against all versions of Kafka?
1. Could the same be achieved by checking if `len(v) == 0` and returning false?

VictorDenisov:
I could only check the length, but then it would accept 1.2 as at least 1.2.1. This solution fixes both cases, so I picked this solution instead of the one that checks only length.

VictorDenisov:
@stevevls I added two unit tests to illustrate the cases that are fixed by this change. I should have done it right away.

testing/version.go
```diff
@@ -20,6 +20,11 @@ func (v semver) atLeast(other semver) bool {
 			return true
 		}
 	}
+	for i := len(v); i < len(other); i++ {
+		if 0 < other[i] {
```
stevevls:
```suggestion
		if other[i] > 0 {
```

I think the intention reads a little bit more clearly if we flip this condition.

PR #290, opened 2019-06-10
Reduce writer allocations in async mode

sagarkrkv:
We can increase the performance of the Kafka writer by avoiding unnecessary allocations when operating in async mode.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -295,13 +295,12 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 		return nil
 	}
 
+	var err error
 	var res chan error
-	var t0 time.Time
 	if !w.config.Async {
 		res = make(chan error, len(msgs))
-		t0 = time.Now()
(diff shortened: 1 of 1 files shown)


writer.go
```diff
@@ -375,10 +378,10 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 			break
 		}
 	}
-
-	t1 := time.Now()
-	w.stats.writeTime.observeDuration(t1.Sub(t0))
-
+	if !w.config.Async {
+		t1 := time.Now()
+		w.stats.writeTime.observeDuration(t1.Sub(t0))
```
achille-roussel: While this stat is not very meaningful in async mode, it's a bit weird that we just won't be reporting it anymore, some programs out there could be depending on it.

Would it be possible to keep it?

achille-roussel:
Looking good, thanks for your contribution!

PR #292, opened 2019-06-13
output messages to get id's

ldrrp:
issue #150 . not sure how you want it pulled in, but i made these changes for my own code base because i needed it today.


stevevls:
Thanks @ldrrp.  Unfortunately, since this change isn't backwards compatible, we can't accept it at the current time.  We're tracking the requirement in #150, though, so when we do get ready to cut a major release, we'll be sure to include this functionality!

PR #294, opened 2019-06-19
Log EOF errors to standard logger

dylannz-sailthru:
They were being logged to the error logger, which caused error-level
logs to be written frequently when topics had no messages to consume:

```
the kafka reader got an unknown error reading partition 14 of my_topic at offset 80: EOF
```

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1883,6 +1883,14 @@ func (r *reader) run(ctx context.Context, offset int64) {
 				r.sendError(ctx, err)
 				break readLoop
 
+			case io.EOF:
+				r.withLogger(func(log *log.Logger) {
+					log.Printf("the kafka reader got an EOF for partition %d of %s at offset %d: %s", r.partition, r.topic, offset, err)
+				})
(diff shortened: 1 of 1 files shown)


mkorolyov:
Faced a lot of EOF's on producer, which are also logged as errors.
Hi @achille-roussel , what do you think of this?

achille-roussel:
@dylannz-sailthru thanks for your PR.

Reading https://github.com/segmentio/kafka-go/blob/362fb4c149fcc1dd09f6a6d215ba401f24bb58b9/reader.go#L1898 it seems like `io.EOF` should be logged to the error logger already. Did you have troubles getting this working?

dylannz-sailthru:
Yep, that was the problem; we only wanted to log at error level, but were getting lots of erroneous io.EOF errors in the error log. So the goal is to move this message to the standard/debug logger :)

mkorolyov:
@achille-roussel any progress with this one?

PR #299, opened 2019-06-25
Start reader go routines on SetOffset

stevevls:
It was a bit surprising when using relative offsets that the reader wasn't started up immediately since the start/end could be changing between the time of calling `SetOffset` and `FetchMessage`.  I could also only start up the reader in the case where a relative offset is specified, but it seems to me that eager initialization is preferable.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1542,9 +1542,10 @@ func (r *Reader) SetOffset(offset int64) error {
 			log.Printf("setting the offset of the kafka reader for partition %d of %s from %d to %d",
 				r.config.Partition, r.config.Topic, r.offset, offset)
 		})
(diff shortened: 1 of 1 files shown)


achille-roussel:
Correct me if I'm wrong, but positioning on an offset is always racy in kafka right? This would just reduce the time between setting the offset and starting the reader?

stevevls:
Yeah...this is true.  We're working around some differences in kafka-go and sarama for our HW project.  We very well may end up dumping this PR later.

stevevls:
Okay...we moved this logic into our project so as not to pollute the kafka-go API.  Thanks!

PR #302, opened 2019-07-02
return io.ErrNoProgress when no progress is being made

achille-roussel:


conn.go
```diff
@@ -1264,7 +1265,10 @@ func (c *Conn) doRequest(d *connDeadline, write func(time.Time, int32) error) (i
 }
 
 func (c *Conn) waitResponse(d *connDeadline, id int32) (deadline time.Time, size int, lock *sync.Mutex, err error) {
-	for {
+	const limit = 100
```
stevevls: Would be curious as to how you arrived at this number.  I have little context here, but at first glance 100 seems like a lot of loops (especially since the counter can get reset then the response ID changes).  Infinite seems worse, though, so it's an improvement!

achille-roussel: I applied exactly zero scientific process to choose this value.

If the ID changes then progress is being made, which is why it gets reset.

My guess is 100 iterations where the goroutine gets descheduled by calling `runtime.Gosched()` may end up on a wait of ~10ms to ~1s (if the programs is heavily CPU bound and has lots of goroutines), so it should allow for bailing quickly without taking too much risk to get false positives.

Happy to apply more science to this if you have a suggestion on how to determine what a good value would be 👍

stevevls: I think a rough guess is great.  I would love to see some of this commentary pasted in to the code so we remember how we arrived at the magic number. 😄

PR #314, opened 2019-07-17
Extract record batch structure

VictorDenisov:
@achille-roussel @stevevls Would you be willing to merge this refactoring?

As I was working on adding transaction support to batches I realized that constructing batches deep in the invocation stack is very tedious as all their parameters should be passed to V3 and V7 functions.

I extracted record batch structure and now it's constructed one level above the writeProduceRequest functions. This way if we need to add arguments to recordBatches we can do it without touching signatures of writeProduceRequest functions.

### DIFF
--- a/recordbatch.go
+++ b/recordbatch.go
@@ -2,6 +2,7 @@ package kafka
 
 import (
 	"bytes"
+	"time"
 )
 
 const recordBatchHeaderSize int32 = 0 +
@@ -72,6 +73,8 @@ func (r *recordBatch) init() (err error) {
 }
 
 func (r *recordBatch) writeTo(wb *writeBuffer) {
+	wb.writeInt32(r.size)
+
 	baseTime := r.msgs[0].Time
 	lastTime := r.msgs[len(r.msgs)-1].Time
 	if r.compressed != nil {
@@ -86,4 +89,16 @@ func (r *recordBatch) writeTo(wb *writeBuffer) {
 			}
 		})
 	}
+}
+
+func recordSize(msg *Message, timestampDelta time.Duration, offsetDelta int64) int {
+	return 1 + // attributes
+		varIntLen(int64(milliseconds(timestampDelta))) +
+		varIntLen(offsetDelta) +
+		varBytesLen(msg.Key) +
+		varBytesLen(msg.Value) +
+		varArrayLen(len(msg.Headers), func(i int) int {
+			h := &msg.Headers[i]
+			return varStringLen(h.Key) + varBytesLen(h.Value)
+		})
 }
--- a/recordbatch.go
+++ b/recordbatch.go
@@ -63,7 +63,11 @@ type recordBatch struct {
 	size       int32
 }
 
-func (r *recordBatch) init() (err error) {
+func newRecordBatch(codec CompressionCodec, msgs ...Message) (r *recordBatch, err error) {
+	r = &recordBatch{
+		codec: codec,
+		msgs:  msgs,
+	}
 	if r.codec == nil {
 		r.size = recordBatchSize(r.msgs...)
 	} else {
--- a/recordbatch.go
+++ b/recordbatch.go
@@ -1,0 +1,59 @@
+package kafka
+
+import (
+	"bytes"
+)
+
+const recordBatchHeaderSize int32 = 0 +
+	8 + // base offset
+	4 + // batch length
+	4 + // partition leader epoch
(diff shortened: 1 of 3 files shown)


record_batch.go
```diff
@@ -0,0 +1,204 @@
+package kafka
```
achille-roussel: Can we name this file `recordbatch.go` to match the naming style of the rest of the files?

record_batch.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+type recordBatch struct {
+	codec      CompressionCodec
+	attributes int16
+	msgs       []Message
+	compressed []byte
+}
+
+func (r *recordBatch) Init() (err error) {
```
achille-roussel:
```suggestion
func (r *recordBatch) init() (err error) {
```
Unless I'm mistaking, it seems like this method is only used in the package, there's no need for it to be exported.

record_batch.go
```diff
@@ -0,0 +1,204 @@
...
+
+	for _, h := range r.msg.Headers {
+		writeVarInt(w, int64(len(h.Key)))
+		w.Write([]byte(h.Key))
+		writeVarInt(w, int64(len(h.Value)))
+		w.Write(h.Value)
+	}
+}
+
+type recordBatch struct {
```
achille-roussel:
```suggestion
type recordBatchWriter struct {
```
My understanding is this type is only used for writing record batches, not for reading them?

record_batch.go
```diff
@@ -0,0 +1,204 @@
...
+		r.compressed, err = r.codec.Encode(recordBuf.Bytes())
+		if err != nil {
+			return
+		}
+		r.attributes = int16(r.codec.Code())
+	}
+	return nil
+}
+
+func (r recordBatch) size() (size int32) {
```
achille-roussel:
```suggestion
func (r *recordBatch) size() (size int32) {
```
It looks like we're using these values through pointers (https://github.com/segmentio/kafka-go/pull/306), how about keeping things consistent and pass the value by pointer everywhere?

VictorDenisov:
Sure. I'll redo it after the merge of #306

record_batch.go
VictorDenisov: I'm using non pointer receivers when I want to communicate that this is a const method - method that doesn't modify the structure. If you insist I can always use pointer receivers.

achille-roussel: I assumed that this was the intention, however in practice I found that mixing pointer and value receiver on a single type tends to bring more confusion to the readers. I feel like a type should either be immutable, in which case all it’s method should have value receivers, or mutable and have all its methods have pointer receivers. 

Consistency is a powerful way of guiding people’s intuition to follow the intent of the original author, and helps them be more successful with the tools we build.

VictorDenisov: I see your point. Pushed the fix.

achille-roussel:
@VictorDenisov we've made a couple more optimizations that we'll have to integrate here as well:
- https://github.com/segmentio/kafka-go/pull/318
- https://github.com/segmentio/kafka-go/pull/320

VictorDenisov:
@achille-roussel I updated the PR for the latest code.

recordbatch.go
```diff
@@ -0,0 +1,104 @@
...
+	codec      CompressionCodec
+	attributes int16
+	msgs       []Message
+
+	// parameters calculated during init
+	compressed *bytes.Buffer
+	size       int32
+}
+
+func (r *recordBatch) init() (err error) {
```
achille-roussel: It looks like it's never valid to use a `recordBatch` without calling `init`, how about making it a constructor instead?
```go
func newRecordBatch(codec CompressionCodec, msgs []Message) *recordBatch {
    ...
}
```
One challenge is managing the error returned by `compressRecordBatch`, maybe the `recordBatch` type can hold the error value like we do for `writeBuffer`?

VictorDenisov: Yeah, I was choosing between this option and the option that I implemented. My motivation was to not duplicate the fields in the record batch and in the declaration of the method. I don't have any strong attachment to either solution. Would you prefer your variant?

achille-roussel: Since a `recordBatch` value is never fully initialized without calling `init` I think it would make the code harder to use wrong to have the construction contained in a single function 👍

VictorDenisov: Done

achille-roussel:
Thanks for your contribution!

PR #318, opened 2019-07-22
kafka.writeBuffer

achille-roussel:
This PR refactors the write paths to remove the intermediary buffer that we used to compute checksums of record batches. These temporary buffers would end up being really large as they would have to hold the entire batch, depending on the buffering strategy this could end up being multiple MB of memory allocated on each write to kafka.

Memory footprint on writes should be greatly reduced by this change:
```
benchmark               old ns/op     new ns/op     delta
BenchmarkConn/Write     1026737       1004170       -2.20%

benchmark               old MB/s     new MB/s     speedup
BenchmarkConn/Write     9.74         9.96         1.02x

benchmark               old allocs     new allocs     delta
BenchmarkConn/Write     10             7              -30.00%

benchmark               old bytes     new bytes     delta
BenchmarkConn/Write     14738         322           -97.82%
```

Varint encoding is also much faster now that we work on a local buffer:
```
benchmark                old ns/op     new ns/op     delta
BenchmarkWriteVarInt     34.5          9.70          -71.88%
BenchmarkReadVarInt      50.1          47.9          -4.39%

benchmark                old allocs     new allocs     delta
BenchmarkWriteVarInt     0              0              +0.00%
BenchmarkReadVarInt      0              0              +0.00%

benchmark                old bytes     new bytes     delta
BenchmarkWriteVarInt     0             0             +0.00%
BenchmarkReadVarInt      0             0             +0.00%
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -55,6 +55,7 @@ type Conn struct {
 	// write buffer (synchronized on wlock)
 	wlock sync.Mutex
 	wbuf  bufio.Writer
+	wb    writeBuffer
 
 	// deadline management
 	wdeadline connDeadline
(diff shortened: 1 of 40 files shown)


achille-roussel:
Thanks for the reviews!

PR #319, opened 2019-07-23
protect against invalid header counts

achille-roussel:
This PR protects the program against bugs or corrupted data on the wire which could result in crashes caused by allocating very large header slices.

We set a limit to 1024 header entries per message, which in practice seem more than any acceptable use cases, although we could increase it later as needed, I just want to make sure any insane value is not let through and cascades in even worse outcomes.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1040,8 +1040,14 @@ func (c *Conn) writeCompressedMessages(codec CompressionCodec, msgs ...Message) 
 			err = errInvalidWriteTopic
 			return
 		}
+
 		if msg.Partition != 0 {
 			err = errInvalidWritePartition
+			return
+		}
+
+		if len(msg.Headers) > headerCountLimit {
+			err = fmt.Errorf("header count beyond the supported limit: %d > %d", len(msg.Headers), headerCountLimit)
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


stevevls:
If we don't need this PR immediately, I'd like to take some time to dig in and better understand why we get this weird data on the wire when we're running in a mixed mode.  The change seems good in and of itself (always better to be defensive), but since it technically deviates from the protocol, I think we should try to find the root cause and avoid merging this if possible.  I can't think of why anyone would ever want more than 1024 headers, though. 🤷‍♂ WDYT?

achille-roussel:
I think my reasoning is it's better to return an error when we detect corruption or malformed data on the wire than crash the program.

The 1024 limit is arbitrary, we can make it much higher. the panic we got was `panic: runtime error: makeslice: len out of range` which indicates the amount of memory we attempted to allocated was larger than the available memory, or the number was negative.

The kafka protocol uses a varint to describe the number of headers, so technically it is possible to have 2^64-1 headers, but this is not practical, so I think it's left to interpretation what a sane limit should be here (just like http servers setting a limit on the size of the http request headers they accept).

I agree that this should not distract us from tracking down the underlying cause, it's a bandaid to handle the issue more gracefully.

achille-roussel:
I haven't encountered this issue anymore with recent version of kafka-go, so I'm just going to close for now since I'm planning on making more changes to this part of the code in other PRs.

Thanks for the reviews!

PR #323, opened 2019-07-23
Make sasl.Mechanism safe for concurrent use

### LINKED ISSUE #317: Reading message fails with SASL_SSL when using consumer groups
**Describe the bug**
We have a Kafka cluster configured to use `SASL_SSL` and `SCRAM-SHA-256`.  Using a `Dialer` configured with `github.com/segmentio/kafka-go/sasl/scram`, I am able to:
* Create a topic
* Write a message
* Read the message when not using a consumer group

But when I try to change the reader to use a consumer group, the fetch starts failing with an error originating from scram: `server nonce did not extend client nonce`

**Kafka Version**
Kafka 2.1.1

**To Reproduce**
With a server configured as above:
```go
func main() {
	saslmech, err := scram.Mechanism(scram.SHA256, "my-user", "my-pass")
	if err != nil {
		panic(err)
	}

	dialer := &kafka.Dialer{
		Timeout:       10 * time.Second,
		DualStack:     true,
		SASLMechanism: saslmech,
		TLS: &tls.Config{
			InsecureSkipVerify: true,
		},
	}

	rdr := kafka.NewReader(kafka.ReaderConfig{
		Brokers: []string{"my-broker:9094"},
		Topic:   "my-topic",
		Dialer:  dialer,
		GroupID: "my-group",
	})

	_, err := rdr.FetchMessage(context.Background())
	if err != nil {
		panic(err)
	}
}
```

This errors on the `FetchMessage` line with `server nonce did not extend client nonce`; if I remove the `GroupID` from the reader config, it reads the message successfully.

stevevls:
While is a breaking change for the sasl.Mechanism interface, it's
not expected that library clients are implementing said interface
as Kafka only accepts a very specific set of mechanisms.

Fixes #317

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -283,17 +283,16 @@ func (d *Dialer) connect(ctx context.Context, network, address string, connCfg C
 // In case of error, this function *does not* close the connection.  That is the
 // responsibility of the caller.
 func (d *Dialer) authenticateSASL(ctx context.Context, conn *Conn) error {
-	mech, state, err := d.SASLMechanism.Start(ctx)
-	if err != nil {
+	if err := conn.saslHandshake(d.SASLMechanism.Name()); err != nil {
 		return err
 	}
-	err = conn.saslHandshake(mech)
+
+	sess, state, err := d.SASLMechanism.Start(ctx)
 	if err != nil {
 		return err
 	}
 
-	var completed bool
-	for !completed {
+	for completed := false; !completed; {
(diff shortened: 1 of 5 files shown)


stevevls:
@achille-roussel I think we can make a case here for ignoring b/w compatibility similar to the way that we did with the streaming compression change, but LMK if you think otherwise.

PR #324, opened 2019-07-24
revisit wire corruption detection

achille-roussel:
https://github.com/segmentio/kafka-go/pull/302 did not work as well as I expected, and ended up burning CPU without making much progress.

This PR reverts the change and replaces it with a different approach: we have no use cases for multiplexing multiple requests on connections for now, so detect the common case where there is only one goroutine using a connection, and if the correlation IDs don't match we know that we have corrupted input.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1279,21 +1279,25 @@ func (c *Conn) doRequest(d *connDeadline, write func(time.Time, int32) error) (i
 }
 
 func (c *Conn) waitResponse(d *connDeadline, id int32) (deadline time.Time, size int, lock *sync.Mutex, err error) {
-	c.rlock.Lock()
-
 	for {
 		var rsz int32
 		var rid int32
 
+		c.rlock.Lock()
 		deadline = d.setConnReadDeadline(c.conn)
+		rsz, rid, err = c.peekResponseSizeAndID()
 
-		if rsz, rid, err = c.peekResponseSizeAndID(); err != nil {
+		if err != nil {
+			d.unsetConnReadDeadline()
+			c.conn.Close()
+			c.rlock.Unlock()
 			break
 		}
 
 		if id == rid {
 			c.skipResponseSizeAndID()
 			size, lock = int(rsz-4), &c.rlock
+			// Don't unlock the read mutex to yield ownership to the caller.
 			break
 		}
 
@@ -1303,6 +1307,7 @@ func (c *Conn) waitResponse(d *connDeadline, id int32) (deadline time.Time, size
 			// one it expects. This is a sign that the data we are reading on
 			// the wire is corrupted and the connection needs to be closed.
 			err = io.ErrNoProgress
+			c.rlock.Unlock()
 			break
 		}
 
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -1255,67 +1271,60 @@ func (c *Conn) doRequest(d *connDeadline, write func(time.Time, int32) error) (i
...
+			// the wire is corrupted and the connection needs to be closed.
+			err = io.ErrNoProgress
+			break
 		}
 
 		// Optimistically release the read lock if a response has already
 		// been received but the current operation is not the target for it.
 		c.rlock.Unlock()
 		runtime.Gosched()
+		c.rlock.Lock()
```
stevevls: I think it would read a little more clearly if you added a `c.rlock.Lock()` as the first statement in the loop and removed this one and the one before entering the loop.  IIUC, the logic would be the same.

conn.go
```diff
@@ -1255,67 +1271,60 @@ func (c *Conn) doRequest(d *connDeadline, write func(time.Time, int32) error) (i
...
-		} else {
-			attempt = 0
-		}
+	if err != nil {
+		d.unsetConnReadDeadline()
+		c.conn.Close()
+	}
 
-		lastID = rid
+	if lock == nil {
```
stevevls: This could be combined with the above case.  Essentially `(err != nil) == (lock == nil)` since you either break from the loop with an error or because you successfully read.

conn.go
achille-roussel: I tried to refactor the code to apply this suggestion, but I think it's actually making it harder to understand: the lock is not always released, sometimes it's returned to the caller to transfer ownership, so having it locked in the loop makes this case (which is the common case) much harder to reason about, so I think we should keep it as it is for now.

achille-roussel: Actually after re-reading your other comment I think I have something that's not too tricky, let me know what you think about it.

stevevls: Nice...I like that a lot more. 😁

bat22:
Hi guys!
Looks like i've got this issue: no new messages from kafka and CPU burning in /github.com/segmentio/kafka-go.(*Conn).waitResponse method.
Do you have any idea how to reproduce it?

achille-roussel:
Thanks for reaching out. 

Happy to revert this change if it breaks. We’ve been running it in production for weeks now tho, so it may be an issue that arises from a specific setup. 

I did observe an issue with high CPU usage, which I addressed with this PR, but maybe it broke a different case. 

Can you confirm that this code (which we merged yesterday) is what you have running?

Also info like the kafka version you run, or the features of the package that you use could be useful.

bat22:
> Can you confirm that this code (which we merged yesterday) is what you have running?

No, i used v0.2.5 with kafka 2.2.0, only consuming messages.
Reader configuration:
```
dialer := new(kafka.Dialer)
*dialer = *kafka.DefaultDialer
dialer.TLS = l.params.TlsConfig
cfg := kafka.ReaderConfig{
    Dialer:   dialer,
    Brokers:  l.params.Brokers,
    Topic:    topic,
    Partition: pid,
}
```

Yesterday i built my service with kafka-go master branch and ran in parallel with old version and both versions still work well.

The issue happened only once, but all instances was affected. I don't know reason for this issue and ask you steps to reproduce it. Without it information testing can take a long time.

achille-roussel:
v0.2.5 is a bit old at this point, is upgrading to a more recent version of the package an option for you? I have good confidence that the issue you described has been addressed already, as I observed a similar problem a while back and had put fixes in place.

Regarding reproducing it, to the best of my knowledge this was a bug in an older version of kafka-go which was triggered by race conditions, so I was never able to reproduce it consistently.

bat22:
Ok, i will continue to monitor the service with latest kafka-go. If i get any results i will write here.

PR #325, opened 2019-07-24
kafka.(*readBuffer)

achille-roussel:
This PR refactors the kafka-go _read paths_ to use a new `readBuffer` type (similar to what we introduced in #318).

This new model reduces the amount of error handling and byte counting that was proliferating all over the place, making the code more robust and easier to read. It also pave the road towards removing the intermediary buffer used when decompressing record batches, and brings us one step closer to being able to offer a zero-copy API for consumers.

I also took some time to cleanup a lot of tests, remove code duplication, and hopefully fix a couple of bugs on the way by reducing the amount of special assumptions that are made between different layers of the codebase.


achille-roussel:
@stevevls if you want to take a second look at this, I think it should be ready to go. Besides the main change which is the introduction of the `readBuffer` abstraction, the PR addresses a race condition in `kafka.Reader`, and a behavior change I've observed on master, where time lag sometimes get reported incorrectly (maybe the message timestamps are wrong?).

achille-roussel:
Closing in favor of https://github.com/segmentio/kafka-go/pull/438

PR #330, opened 2019-08-02
declare SeekDontCheck as uint32 to avoid integer overflow in 386 arch

lovelykd:
When i build my program with `GOARCH=386`, i got the following error caused by constant `SeekDontCheck`:
```
..\github.com\segmentio\kafka-go\conn.go:613:19: constant 2147483648 overflows int
..\github.com\segmentio\kafka-go\conn.go:614:9: constant -2147483649 overflows int
..\github.com\segmentio\kafka-go\conn.go:758:38: constant 2147483648 overflows int
```
The reason it that its value (1<<31) exceeds the limit of 32-bit integer.

I think it is OK to explicitly declare it as _uint32_ and change the type of `whence` param to _uint32_ as well.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -576,7 +576,7 @@ func (c *Conn) SetWriteDeadline(t time.Time) error {
 // to interpret it.
 //
 // See Seek for more details about the offset and whence values.
-func (c *Conn) Offset() (offset int64, whence int) {
+func (c *Conn) Offset() (offset int64, whence uint32) {
 	c.mutex.Lock()
 	offset = c.offset
 	c.mutex.Unlock()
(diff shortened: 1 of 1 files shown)


achille-roussel:
Thanks for reporting the issue, and excited to learn this package is being used on 32 bits architectures!

Changing the type is likely going to break compilation on some other programs, so I doubt we can do that.

Would it work if we changed the value of this flag from `1<<31` to `1<<30`?

lovelykd:
@achille-roussel Thank you for your quick response.
> Would it work if we changed the value of this flag from 1<<31 to 1<<30?

I've just confirmed that it works.

lovelykd:
@achille-roussel I made a new pull request based on your suggestion. Please check [!331](https://github.com/segmentio/kafka-go/pull/331).

PR #331, opened 2019-08-06
decrease value of seekdontcheck to avoid integer overflow in 386 arch

lovelykd:
The value `1<<31` exceeds the limit of 32-bit integer, which will cause compile failure in 386 arch.
Reducing it to 1<<30 (as suggested by @achille-roussel in [!330](https://github.com/segmentio/kafka-go/pull/330)) solves the error and preserves the flag's functionality.


achille-roussel:
Thanks for the contribution!

PR #334, opened 2019-08-08
 Add CRC32 and Murmur2 balancers

stevevls:


stevevls:
This PR supplants #142 due to inactivity.  Since I was the one giving feedback, I figure I should be the one to implement it. 😁

PR #336, opened 2019-08-09
 Fixed compatibility with the sarama hash partitioner

stevevls:


stevevls:
Since they touch the same lines, this PR also includes the commit for #335.

balancer.go
```diff
@@ -134,31 +134,51 @@ var (
...
 	if msg.Key == nil {
 		return h.rr.Balance(msg, partitions...)
 	}
 
 	hasher := h.Hasher
-	if hasher == nil {
+	shared := hasher != nil
+
+	if shared {
+		h.lock.Lock()
```
achille-roussel: Alternatively, you could `defer h.lock.Unlock()` here to avoid checking `shared` again below.

Issue #339, opened 2019-08-21
Kafka MirrorMaker problem with nil keys


Zavierazo:
**Describe the bug**
When messages are produces with nil key seems message are created with empty value instead of nil.
This causes the mirrormaker to send everything to the same partition. 

**Kafka Version**
2.3.0

**To Reproduce**
Send multiple messages with nil key:
```
kafka.Message{
    Key:   nil,
    Time:  time.Now(),
    Value: json,
}
```
Enable mirror-maker between source kafka cluster to destination kafka cluster:
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=27846330

Example:
`./bin/kafka-mirror-maker.sh --consumer.config consumer.properties --producer.config producer.properties --num.streams 3 --whitelist="topic-name.*"
`

See the results

Source cluster:
```
./kafka/bin/kafka-run-class.sh kafka.tools.GetOffsetShell --topic kafka-go-poc --broker-list 10.222.104.73:9092  
kafka-go-poc:0:152
kafka-go-poc:1:153
kafka-go-poc:2:150
kafka-go-poc:3:152
kafka-go-poc:4:151
kafka-go-poc:5:153
kafka-go-poc:6:153
kafka-go-poc:7:151
kafka-go-poc:8:155
kafka-go-poc:9:153
kafka-go-poc:10:152
kafka-go-poc:11:152
kafka-go-poc:12:153
kafka-go-poc:13:153
kafka-go-poc:14:151
kafka-go-poc:15:148
kafka-go-poc:16:154
kafka-go-poc:17:155
kafka-go-poc:18:153
kafka-go-poc:19:154
kafka-go-poc:20:151
kafka-go-poc:21:153
kafka-go-poc:22:150
kafka-go-poc:23:152
kafka-go-poc:24:152
kafka-go-poc:25:154
kafka-go-poc:26:149
kafka-go-poc:27:151
kafka-go-poc:28:151
kafka-go-poc:29:153
kafka-go-poc:30:149
kafka-go-poc:31:152
kafka-go-poc:32:153
kafka-go-poc:33:148
kafka-go-poc:34:151
```

Destination cluster:
```
./kafka/bin/kafka-run-class.sh kafka.tools.GetOffsetShell --topic kafka-go-poc --broker-list 10.185.36.91:9092
kafka-go-poc:0:0
kafka-go-poc:1:0
kafka-go-poc:2:0
kafka-go-poc:3:0
kafka-go-poc:4:0
[...]

PR #340, opened 2019-08-21
#339 Fix problem with nil keys

Zavierazo:
Fix problem with nil keys. Currently nil keys are writed as empty byte instead of nil

### DIFF
--- a/write.go
+++ b/write.go
@@ -80,8 +80,13 @@ func (wb *writeBuffer) writeBytes(b []byte) {
 }
 
 func (wb *writeBuffer) writeVarBytes(b []byte) {
-	wb.writeVarInt(int64(len(b)))
-	wb.Write(b)
+	if b != nil {
+		wb.writeVarInt(int64(len(b)))
+		wb.Write(b)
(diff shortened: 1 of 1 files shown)


write.go
```diff
@@ -621,7 +621,12 @@ func (wb *writeBuffer) writeRecord(attributes int8, baseTime time.Time, offset i
 	wb.writeVarInt(int64(milliseconds(timestampDelta)))
 	wb.writeVarInt(offsetDelta)
 
-	wb.writeVarBytes(msg.Key)
+	if msg.Key != nil {
+		wb.writeVarBytes(msg.Key)
```
achille-roussel: Should we handle this case in the `writeVarBytes` method instead? It could prevent future issues of the same kind.

Zavierazo: Good idea ;) I will do and test it ;)

Zavierazo: Change done and tested.
This also will fix the tombstone used for kafka with log compaction.

achille-roussel:
Thanks for your contribution!

Issue #345, opened 2019-09-04
broken pipe connections using WriteMessages


joaoferrao:
**Describe the bug**
Sporadic (several times a day) producer related `broken pipe` errors. There are multiple topics, all of them with 9 partitions each. 3x brokers are also running in the cluster.
When the microservice is spun up, a Writer is instantiated and is reused for the lifecycle of that pod for each message that needs to be sent to kafka. Each microservice is typically only producing to one or two topics. 

Example error is the one that follows. 

```
error writing messages to <some-topic> (partition 3): write tcp 123.123.123.123:55555->123.123.123.123:9092: write: broken pipe
```

The errors seem to originate from the the below method on `write.go`.

```go
	t0 := time.Now()
	conn.SetWriteDeadline(time.Now().Add(w.writeTimeout))
	if _, err = conn.WriteCompressedMessages(w.codec, batch...); err != nil {
		w.stats.errors.observe(1)
		w.withErrorLogger(func(logger *log.Logger) {
			logger.Printf("error writing messages to %s (partition %d): %s", w.topic, w.partition, err)
		})
		for i, res := range resch {
			res <- &writerError{msg: batch[i], err: err}
		}
	} else {
		for _, m := range batch {
			w.stats.messages.observe(1)
			w.stats.bytes.observe(int64(len(m.Key) + len(m.Value)))
		}
		for _, res := range resch {
			res <- nil
		}
	}
```

**Kafka Version**
2.2.1 running on kubernetes (bitnami's Helm Chart if it helps)

**To Reproduce**
The code simply uses your kafka.WriteMessages method while looping through a channel:

```go
for record := someChan {
  if err := writer.WriteMessages(ctx, record); err != nil {
   log.Info(err)
}
}
```

Configuration of `dialer` and `writer` are:

	dialer := &kafka.Dialer{
		Timeout:   time.Second * 5,
		DualStack: true,
	}

	config := kafka.WriterConfig{
		Brokers:          "<some-broker-list>",
		Topic:            "<some-topic>",
		Dialer:           dialer,
		Balancer:         &kafka.Hash{},
		QueueCapacity:    100,
		BatchSize: 1,
		MaxAttempts: 20,
		RequiredAcks: -1,
		Async:        false,
	}

**Expected behavior**
Can you propose a way to deal with this error? 
Shouldn't the producer retry to redial and resend the message?

**Additional context**
[...]

PR #348, opened 2019-09-05
Add max idle conn timeout for Writer.

sagarkrkv:
https://github.com/segmentio/kafka-go/issues/345#issuecomment-528451990
> We saw a similar issue on our production boxes, we have a metrics topic where we observed weird behavior, messages are successfully published for 25 mins, then we see broken pipe errors for the next 25 mins, and then the cycle continues.

> Upon further investigation we found that,

> We publish one message per 5 mins, and there are 5 partitions, so the Kafka writer sends 1 message to each partition every 25mins. Under the default settings, the Kafka broker closes a connection if it doesn't receive any new message in the last 10 mins. But the Kafka writer doesn’t know that the connection has been closed, so when it tries to a second message on the same connection, it sees broken pipe errors.

> So for the next 25 mins as the writer cycles through the partitions it sends messages on connections which have been closed, and on next cycle new connections are created and the messages sent on these new connections succeed, and cycle repeats.

> When I looked through Kafka documentation, I found that there is a config that can solve this issue, please note that the default value is set to 9mins, which is 1 min less than broker timeout.

> connections.max.idle.ms | Close idle connections after the number of milliseconds specified by this config. | long | 540000 |  

Add max idle connection time for writer.
https://kafka.apache.org/documentation/#producerconfigs

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -106,7 +106,7 @@ type WriterConfig struct {
 	// Connections that were idle for this duration will not be reused.
 	//
 	// Defaults to 9 minutes.
-	IdleTimeout time.Duration
+	IdleConnTimeout time.Duration
 
 	// Number of acknowledges from partition replicas required before receiving
 	// a response to a produce request (default to -1, which means to wait for
@@ -253,8 +253,8 @@ func NewWriter(config WriterConfig) *Writer {
 	if config.RebalanceInterval == 0 {
 		config.RebalanceInterval = 15 * time.Second
 	}
-	if config.IdleTimeout == 0 {
-		config.IdleTimeout = 9 * time.Minute
+	if config.IdleConnTimeout == 0 {
+		config.IdleConnTimeout = 9 * time.Minute
 	}
 
 	w := &Writer{
@@ -564,7 +564,7 @@ type writer struct {
 	maxMessageBytes int
 	batchTimeout    time.Duration
 	writeTimeout    time.Duration
-	idleTimeout     time.Duration
+	idleConnTimeout time.Duration
 	dialer          *Dialer
 	msgs            chan writerMessage
 	join            sync.WaitGroup
@@ -584,7 +584,7 @@ func newWriter(partition int, config WriterConfig, stats *writerStats) *writer {
 		maxMessageBytes: config.BatchBytes,
 		batchTimeout:    config.BatchTimeout,
 		writeTimeout:    config.WriteTimeout,
-		idleTimeout:     config.IdleTimeout,
+		idleConnTimeout: config.IdleConnTimeout,
 		dialer:          config.Dialer,
 		msgs:            make(chan writerMessage, config.QueueCapacity),
 		stats:           stats,
@@ -634,7 +634,7 @@ func (w *writer) run() {
 	var resch = make([](chan<- error), 0, w.batchSize)
 	var lastMsg writerMessage
 	var batchSizeBytes int
-	var idleDeadline time.Time
+	var idleConnDeadline time.Time
 
 	defer func() {
 		if conn != nil {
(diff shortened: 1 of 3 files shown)


writer.go
```diff
@@ -248,6 +253,9 @@ func NewWriter(config WriterConfig) *Writer {
 	if config.RebalanceInterval == 0 {
 		config.RebalanceInterval = 15 * time.Second
 	}
+	if config.ConnMaxIdleTimeout == 0 {
+		config.ConnMaxIdleTimeout = 540 * time.Second
```
achille-roussel:
```suggestion
		config.ConnMaxIdleTimeout = 9 * time.Minute
```
This would be simpler to read for maintainers?

writer_test.go
```diff
@@ -432,3 +439,64 @@ func testWriterSmallBatchBytes(t *testing.T) {
...
+	defer w.Close()
+	ctx := context.Background()
+	msg := Message{Value: []byte("Hello World")}
+	// initial message when we establish a connection with the broker
+	err := w.WriteMessages(ctx, msg)
+	if err != nil {
+		t.Errorf("Unexpected error %v", err)
+	}
+	// sleep for more than the broker's max idle conn time
+	time.Sleep(2300 * time.Millisecond)
```
achille-roussel: Could we make the wait time under 1s? Slow tests are a pain for development iteration.

sagarkrkv: https://circleci.com/gh/sagarkrkv/kafka-go/tree/addConnIdleTimeout

This test is flaky even at 2s, I don't know if it's possible to set up a reliable test, thoughts?

sagarkrkv: I tried several combinations of timeout/sleep values, but I was unable to set up a good test that can reliably simulate the kafka broker connection close issue.

writer.go
```diff
@@ -687,13 +698,18 @@ func (w *writer) run() {
 			if len(batch) == 0 {
 				continue
 			}
+			if conn != nil && time.Now().After(connIdleDeadline) {
```
achille-roussel: Should this appear right above the `if len(batch) == 0` ? https://github.com/segmentio/kafka-go/blob/150aa3726777853d906d78cc1a764093e7d597d5/writer.go#L698

That way even if there is no activity the connection gets properly closed when it reaches its deadline.

writer.go
```diff
@@ -103,6 +103,11 @@ type WriterConfig struct {
 	// The default is to refresh partitions every 15 seconds.
 	RebalanceInterval time.Duration
 
+	// Connections that were idle for this duration will not be reused.
+	//
+	// Defaults to 540 seconds.
+	ConnMaxIdleTimeout time.Duration
```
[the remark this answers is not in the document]
sagarkrkv: This property is tied to a connection, and a writer will have multiple connections, so I think naming it just IdleTimeout will be a bit ambiguous. But I don't have a strong objection.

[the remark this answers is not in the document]
achille-roussel: That's a good point. How about calling it `IdleConnTimeout` then to match the wording of the `Transport` type in `net/http`?

sagarkrkv:
Thanks @achille-roussel :) Is it possible to cut a new release tag?

joaoferrao:
A new tag would be greatly appreciated!

sagarkrkv:
@achille-roussel do you have a tentative date in mind for the new release or are you waiting for a PR to be merged? Cannot use this fix until a new release has been created.

joaoferrao:
@sagarkrkv did you get any sort of feedback (e.g. offline)? I am also hoping this could be

sagarkrkv:
@joaoferrao I haven't heard anything from @achille-roussel

GeorgeJahad:
@joaoferrao it looks like v0.3.4 has the fix:

 https://github.com/segmentio/kafka-go/releases/tag/v0.3.4

https://github.com/segmentio/kafka-go/blob/v0.3.4/writer.go#L697-L699

PR #353, opened 2019-09-19
Handle end of fetched batches of messages gracefully

### LINKED ISSUE #258: Seeing a lot of EOF errors when reading from partitions
Hi, I see in the error logs a lot of messages like

`reader.go:1837: the kafka reader got an unknown error reading partition 13 of presence at offset 20067939: EOF`

is this expected if there are no new messages to consume? Is the EOF coming from kafka? I don't see anything suspicious in the kafka logs.

Kafka version is 2.11-2.0.1

Thanks, Andreas

stevevls:
Previously, the batch read functions returned io.EOF to signal the
end of the batch.  However, this caused ambiguity because there are
valid io.EOF errors from other sources (e.g. a connection that is
closed by the broker).  Because io.EOF can leave the conection in
an unusable state, we are closing the connection and logging it.
However, this also means that we're closing the connection after
every single fetch request, which is not desirable.

This PR also fixes an accidental line deletion of deadline
management that was introduced in #277.

Fixes #258

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -2,10 +2,15 @@ package kafka
 
 import (
 	"bufio"
+	"errors"
 	"io"
 	"sync"
 	"time"
 )
+
+// EndOfBatch is returned by the Read functions when there are no more messages
+// left in the batch.
+var EndOfBatch = errors.New("no more messages in this batch")
 
 // A Batch is an iterator over a sequence of messages fetched from a kafka
 // server.
@@ -68,7 +73,7 @@ func (batch *Batch) close() (err error) {
 		batch.msgs.discard()
 	}
 
-	if err = batch.err; err == io.EOF {
+	if err = batch.err; err == EndOfBatch {
 		err = nil
 	}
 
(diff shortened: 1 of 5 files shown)


stevevls:
That could work, though it could potentially introduce bugs as well since it will change the way errors are propagated.  It's probably safer, though.  Worst case is that other folks currently have code like ours that closes the connection on `io.EOF` b/c otherwise their code wouldn't work.

What worries me more is the number of places we'd need to change to capture the real `io.EOF` and swap it for `io.ErrClosedPipe`.  If we missed just one, it could manifest in some weird behavior if we're treating `io.EOF` as the continue normally case.

stevevls:
Found some prior art where we're already translating `io.EOF` to `io.ErrUnexpectedEOF`, so I'll roll with that.  Since we already have a bunch of cases where we're doing this wrapping, it limits the surface (and my concerns).  I'll leave some docs lying around stressing the importance of doing this wrapping.

PR #361, opened 2019-10-15
Added RackAffinityGroupBalancer strategy

stevevls:
This strategy attempts to optimize round trip transfer time between
brokers and consumers in addition to minimizing inter-zone data
transfer costs in cloud environments.  Currently only supports AWS.

### DIFF
--- a/groupbalancer.go
+++ b/groupbalancer.go
@@ -1,8 +1,10 @@
 package kafka
 
 import (
+	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"sort"
 	"strings"
 	"time"
@@ -144,7 +146,8 @@ func (r RoundRobinGroupBalancer) AssignGroups(members []GroupMember, topicPartit
 // performance benefits by minimizing round trip latency between the consumer
 // and the broker.  In environments where network traffic across racks incurs
 // charges (such as cross AZ data transfer in AWS), this strategy is also a cost
-// optimization measure by keeping network traffic local.
+// optimization measure because it keeps network traffic within the local rack
+// where possible.
 //
 // The primary objective is to spread partitions evenly across consumers with a
 // secondary focus on maximizing the number of partitions where the leader and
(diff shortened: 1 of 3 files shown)


groupbalancer.go
```diff
@@ -133,6 +139,180 @@ func (r RoundRobinGroupBalancer) AssignGroups(members []GroupMember, topicPartit
 	return groupAssignments
 }
 
+// RackAffinityGroupBalancer makes a best effort to pair up consumers with
+// partitions whose leader is in the same rack.  This strategy can have
+// performance benefits by minimizing round trip latency between the consumer
+// and the broker.  In environments where network traffic across racks incurs
+// charges (such as cross AZ data transfer in AWS), this strategy is also a cost
+// optimization measure by keeping network traffic local.
```
Pryz: nit: local -> local to the zone ?

groupbalancer.go
```diff
@@ -185,3 +365,56 @@ func findGroupBalancer(protocolName string, balancers []GroupBalancer) (GroupBal
...
+		}
+	}
+	return "somewhere"
+}
+
+// awsAvailabilityZone queries the metadata endpoint to discover the
+// availability zone where this code is running.  we avoid calling this function
+// unless we know we're in AWS.  Otherwise, in other environments, we would need
+// to wait for the request to 169.254.169.254 to timeout before proceeding.
+func awsAvailabilityZone() string {
```
Pryz: FWIW : While on aws this may not work for teams running on ecs and blocking the ec2 metadata.

groupbalancer.go
stevevls: Better now?

groupbalancer.go
stevevls: Good call.  I also found that there's a better way to do this on ECS, so I pushed up some new code.  Now it will try to use the local ECS metadata and will fail back to trying to EC2 if that doesn't work out.

Pryz: LGTM !

groupbalancer.go
```diff
@@ -185,3 +368,93 @@ func findGroupBalancer(protocolName string, balancers []GroupBalancer) (GroupBal
 	}
 	return nil, false
 }
+
+// findRack is the default rack resolver strategy.  It currently only supports
+//  * ECS with the task metadata endpoint enabled (returns the container
+//    instance's availability zone)
+//  * Linux EC2 (returns the instances' availability zone)
```
erikdw: super minor nit:  `instances'` -> `instance's`?

stevevls: thanks as always for proofreading. 😄

groupbalancer.go
```diff
@@ -133,6 +141,181 @@ func (r RoundRobinGroupBalancer) AssignGroups(members []GroupMember, topicPartit
...
+			memberAssignments[topic] = parts
+		}
+	}
+	return assignments
+}
+
+func (r *RackAffinityGroupBalancer) UserData() ([]byte, error) {
+	var rack string
+	if r.RackResolver == nil {
+		rack = findRack()
```
riking: Consider making `findRack` an exported function such as `RackResolverAWS` so that client code can explicitly call it from their own resolver.

stevevls: Thanks for the feedback!  On further reflection, I'm actually planning to pull out the AWS-specific code.  If I left it in, then I think the suggestion to export `RackResolverAWS` is a good one.  However, that got me to thinking that, from the perspective of exporting a clean and focused public API, AWS code has no place in kafka-go.  That got me thinking even more, and I think that it's weird for the code to use the AWS logic quietly and _by default_.

That said, I can provide this code as an example and make it available to the community that way.  Maybe someday someone will create an open source, multi-cloud library that can be plugged in as a rack resolver.  Until then, I think it's better to keep the cloud provider code out of kafka-go. 😄

Issue #113, opened 2018-08-04
How can I list topics?


hiveminded:
Hello, good work with this library. My question is self-described, I want to list all the available topics and their partitions, or all partitions -> topics, like the `Client.Topics()` of sarama package. Thank you.

achille-roussel:
I'm going to close it due to inactivity, but we'll still welcome a contribution to add this feature 👍

PR #371, opened 2019-10-27
Function for obtaining a list of topics #113

letum:


letum:
If everything is ok, please create new release after accept.

Pryz:
Seems to be a duplicate of https://github.com/segmentio/kafka-go/pull/259

letum:
> Seems to be a duplicate of #259

Yes, it is another implementation. We can leave only your changes for `topicMetadataRequestV1.writeTo` and use `ReadPartitions()`, but at the moment the function returns empty list of Partitions if I call it without arguments.

letum:
@Pryz please any feedback.

PR #372, opened 2019-11-01
Update `Error` consts and `Error.Temporary()`

dtjm:
Main goal is to update the list of `Temporary` errors.  My application dropped many messages when a Kafka broker was down due to a `NotEnoughReplicas` error, so I thought it would be good to update them all. 

https://kafka.apache.org/protocol#protocol_error_codes specifies which errors are retriable.

### DIFF
--- a/error.go
+++ b/error.go
@@ -6,6 +6,7 @@ import (
 )
 
 // Error represents the different error codes that may be returned by kafka.
+// https://kafka.apache.org/protocol#protocol_error_codes
 type Error int
 
 const (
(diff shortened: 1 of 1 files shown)


dtjm:
@achille-roussel yes I just included the errors marked as "Retriable" in the documentation.

For some reason `CORRUPT_MESSAGE (2)` is called `InvalidMessage` here: 
https://github.com/segmentio/kafka-go/blob/master/error.go#L14

achille-roussel:
Got it, thanks for clarifying. Kafka must have changed the name of the error between versions, this code was written when 0.10 was the latest version.

dtjm:
thanks for the super-fast turnaround!

PR #375, opened 2019-11-06
lazily load API versions

achille-roussel:
This PR modifies the mechanism for negotiating API versions between the client and the broker so it happens lazily on first use of the connection instead of during the construction of the `kafka.Conn` value.

It addresses a rare edge case where clients could stay stuck waiting for a response during the construction of the connection, because no deadlines were configured yet.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1398,19 +1398,18 @@ var defaultApiVersions map[apiKey]ApiVersion = map[apiKey]ApiVersion{
 }
 
 func (c *Conn) ApiVersions() ([]ApiVersion, error) {
-	id, err := c.doRequest(&c.rdeadline, func(deadline time.Time, id int32) error {
-		if deadline.IsZero() {
-			// ApiVersions is called automatically when API version negotiation
-			// needs to happen, so we are not garanteed that a read deadline has
-			// been set yet. Fallback to use the write deadline in case it was
-			// set, for example when version negotiation is initiated during a
-			// produce request.
-			deadline = c.wdeadline.deadline()
-		}
+	deadline := &c.rdeadline
 
-		now := time.Now()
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -178,41 +194,41 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
...
+		return nil, err
+	}
+
+	v = make(apiVersions, len(brokerVersions))
+
+	for _, a := range brokerVersions {
+		v[apiKey(a.ApiKey)] = a
+	}
+
+	c.apiVersions.Store(v)
```
stevevls: I'm trying to determine whether there are any cases where we'd run into the thundering herd problem here.  The returned API versions will be consistent (at least I'd assume the broker we're connected to isn't changing its API versions out from under us!), but we could end up firing the call many times.  Do we expect much in the way of parallel produces or fetches?  I also am curious whether, as we add more uses cases for API negotiation, the thundering herd could become more an issue.  WDYT?

achille-roussel: I think the use cases we have right now are mostly single inflight requests, but you make a good point that the API is safe to use concurrently from multiple goroutines.

I can change it to use a blocking model so we guarantee that only one API Version request will be sent.

achille-roussel: On second thoughts, the number of API Version requests will never exceed the number of concurrent Produce or Fetch requests being sent, so while it may double the number of round trips it is strictly bound, there's no risk of exponentially increasing the request volume.

Let me know if you think that's good enough of a guarantee, otherwise I can change the synchronization mechanism.

stevevls: That sounds fine for now.  We can always revisit in the future if/when we find ourselves doing a lot more API version negotiation.  I think concurrent requests are also a pretty rare case on a single connection.

achille-roussel:
Thanks for the review!

PR #385, opened 2019-12-09
Fix log for syncing assignment

anandsinghkunwar:
* This log message should not be an error log, should be an info log instead


stevevls:
Good catch!  Thanks for the contribution.

PR #388, opened 2019-12-18
Wildcard topics

varsanojidan:
This PR revolves around giving the Reader capabilities to read from wildcard topics

There is a singleton routine called the topic scanner that gets triggered by the first reader that wants to use a wildcard topic evaluation. Those readers that want to use the wildcard topics will subscribe to the topic scanner and receive periodic updates from the scanner that keeps track of them and their wildcard pattern. The scanner also allows readers to unsubscribe when they close. The scanner will run each subscriber's regex against the list of all the topics in that broker to send it only matching topics (regex matching is based off of the golang regex.MatchString()). The scanner will cache the list of topics for a broker while its updating each subscriber to reduce the amount of calls made to kafka.

One decision I made was that if the reader is not associated with a consumer group and has wildcards enabled, it will ignore the partition field as there is no guarantee you'll find a specific partition across several topics, though that can be changed based on your input.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -9,7 +9,6 @@ import (
 	"net"
 	"os"
 	"path/filepath"
-	"regexp"
 	"runtime"
 	"sync"
 	"sync/atomic"
@@ -36,12 +35,6 @@ type Partition struct {
 	Replicas []Broker
 	Isr      []Broker
 	ID       int
-}
-
-// Topic carries the metadata associated with a kafka topic
-type Topic struct {
-	Name      string
-	ErrorCode int16
 }
 
 // Conn represents a connection to a kafka broker.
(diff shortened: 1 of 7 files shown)


achille-roussel:
@varsanojidan Thanks a lot for submitting a contribution to this project!

We'd like to avoid adding complexity to the `kafka.Reader` type (which is doing a lot already), but also recognize the need for dynamic discovery of topics, and would like to offer some building block for this in kafka-go.

Your concept of a Topic Scanner sounds interesting, I wonder if as a first step, we could try to introduce a public `TopicScanner` or `TopicWatcher` API which provides only the functionality of watching for changes to a topics that match a given pattern.

What do you think?

varsanojidan:
That sounds good. We can work through this incrementally to ensure the architecture of the feature is as seamless as possible. I will follow-up with changes to the topic scanner to make it a public interface that subscribes to a singular regex and sends back a channel that you could listen on to receive the updates, how does that sound?

matrikagupta:
Hi All, 

I noticed that this PR is not merged, but can someone please suggest alternative solution for the wildcard topics.

sneerin:
i don't think that topic scanner is a good idea, Kafka is not dynamic in terms of topic, so its better to have them once during start.

varsanojidan:
@sneerin the confluent kafka libraries support wildcard topics, it may not be the most optimal use of Kafka but it is a use-case people have when using it. The topic scanner would only be overhead if you choose to use wildcard topics.

achille-roussel:
Hello @varsanojidan, do you mind explaining what you mean by "overhead"? My understanding is that Kafka doesn't have this wildcard support built in, client libraries have to build it by periodically listing existing topic and filtering them based on a naming pattern, which is exactly what the scanner suggested here does.

achille-roussel:
Ah disregard my last message, I think I understood, you meant the scanner overhead only applies if you need it... (still need coffee today) 👍

varsanojidan:
@achille-roussel haha yeah, I opened up a PR for a public topic scanner here: #408 
Im unable to assign it to anyone but lmk if this is more in line with what you were talking about.
My team is eager to drop the confluent kafka library already but we need wildcard support for our usecase :P.

PR #396, opened 2020-01-15
Replace Zstandard wrapper with native Go implementation

pascaldekloe:
Follow-up on stale #303 with full use of the streaming API from @klauspost.

Decompression is slower while the (default) compression ratio went down from 5 to 3.

```
name                           old time/op    new time/op    delta
Compression/zstd/compress-4      5.64ms ± 2%    3.02ms ± 2%   -46.42%  (p=0.000 n=10+10)
Compression/zstd/decompress-4     911µs ± 2%    2386µs ± 2%  +162.10%  (p=0.000 n=10+10)

name                           old speed      new speed      delta
Compression/zstd/compress-4    21.0MB/s ± 2%  44.3MB/s ± 2%  +111.39%  (p=0.000 n=10+10)
Compression/zstd/decompress-4  2.13GB/s ± 2%  0.81GB/s ± 2%   -61.85%  (p=0.000 n=10+10)
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,8 +5,7 @@ go 1.11
 require (
 	github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21
 	github.com/golang/snappy v0.0.1
-	github.com/google/go-cmp v0.4.0 // indirect
-	github.com/klauspost/compress v1.9.7
+	github.com/klauspost/compress v1.9.8
 	github.com/pierrec/lz4 v2.0.5+incompatible
 	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c
 	github.com/xdg/stringprep v1.0.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -3,9 +3,10 @@ module github.com/segmentio/kafka-go
 go 1.11
 
 require (
-	github.com/DataDog/zstd v1.4.0
 	github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21
 	github.com/golang/snappy v0.0.1
+	github.com/google/go-cmp v0.4.0 // indirect
+	github.com/klauspost/compress v1.9.7
 	github.com/pierrec/lz4 v2.0.5+incompatible
 	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c
 	github.com/xdg/stringprep v1.0.0 // indirect
--- a/zstd/zstd.go
+++ b/zstd/zstd.go
@@ -15,7 +15,7 @@ func init() {
 
 const Code = 4
 
-const DefaultCompressionLevel = int(zstdlib.SpeedDefault)
+const DefaultCompressionLevel = 3
 
 type CompressionCodec struct{ level zstdlib.EncoderLevel }
 
--- a/zstd/zstd.go
+++ b/zstd/zstd.go
@@ -1,12 +1,11 @@
-// +build cgo
-
+// Package zstd implements Zstandard compression.
 package zstd
 
 import (
 	"io"
 	"sync"
 
-	"github.com/DataDog/zstd"
+	zstdlib "github.com/klauspost/compress/zstd"
 	kafka "github.com/segmentio/kafka-go"
 )
 
@@ -14,20 +13,18 @@ func init() {
 	kafka.RegisterCompressionCodec(NewCompressionCodec())
 }
 
-const (
-	Code = 4
+const Code = 4
 
-	DefaultCompressionLevel = zstd.DefaultCompression
-)
+const DefaultCompressionLevel = int(zstdlib.SpeedDefault)
 
-type CompressionCodec struct{ level int }
+type CompressionCodec struct{ level zstdlib.EncoderLevel }
 
 func NewCompressionCodec() *CompressionCodec {
 	return NewCompressionCodecWith(DefaultCompressionLevel)
 }
 
 func NewCompressionCodecWith(level int) *CompressionCodec {
-	return &CompressionCodec{level}
+	return &CompressionCodec{zstdlib.EncoderLevelFromZstd(level)}
 }
 
 // Code implements the kafka.CompressionCodec interface.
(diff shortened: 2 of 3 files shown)


achille-roussel:
Thanks a lot for this contribution @pascaldekloe !

What do you think about keeping the compression ratio to 5? I'm concerned that programs out there which rely on this package will suddenly see a change in behavior which could lead to exhausting disk space on kafka brokers. The default compression ratio in kafka-go would differ from what the dependency for zstd sets, but we can document why we're making this decision.

Could you share a CPU and memory profile of the decompression benchmark? I'm curious to see where the difference comes from and if there's anything we can do about it.

Let me know!

klauspost:
The numbers of the benchmark look a bit weird to me. Could you share the benchmark if it is not included?

go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/segmentio/kafka-go
 go 1.11
 
 require (
-	github.com/DataDog/zstd v1.4.0
 	github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21
 	github.com/golang/snappy v0.0.1
+	github.com/google/go-cmp v0.4.0 // indirect
```
achille-roussel: This seems to be caused by `klauspost/comrpess` not having a `go.mod` file, I opened https://github.com/klauspost/compress/issues/208 to follow up.

achille-roussel:
@klauspost you can run the benchmark with `go test -v -run '^$' -bench Compress/zstd` in the root directory.

I took a closer look at the `zstd` package, it looks like there are lots of unguarded `println` calls (some are done if the `debug` variable is `true`, some aren't). Do these show up in practice? I'm concerned about printing a bunch of random stuff to stdout on every program which depends on kafka-go (e.g. https://github.com/klauspost/compress/blob/master/zstd/decoder.go#L216)

klauspost:
Where do I find this?
```
--- FAIL: BenchmarkCompression
    compression_test.go:291: open src\encoding\json\testdata\code.json.gz: The system cannot find the path specified.
```

zstd/zstd.go
```diff
@@ -38,161 +35,91 @@ func (c *CompressionCodec) Name() string { return "zstd" }
...
-type buffer struct {
-	input  []byte
-	output []byte
-}
-
-func (b *buffer) reset() {
-	b.input = b.input[:0]
-	b.output = b.output[:0]
+	w.err = io.ErrClosedPipe
+	encPool.Put(w.enc)
```
klauspost: Defer this, otherwise someone might pick it up before it is closed.

klauspost:
There is something wonky with your benchmarks.

If you only do the compression the numbers are drastically different, eg:
```
func benchmarkCompression(b *testing.B, codec kafka.CompressionCodec, buf *bytes.Buffer, payload []byte) float64 {
	// In case only the decompression benchmark are run, we use this flags to
	// detect whether we have to compress the payload before the decompression
	// benchmarks.

	b.Run("compress", func(b *testing.B) {
		r := bytes.NewReader(payload)
		b.SetBytes(int64(len(payload)))
		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			buf.Reset()
			r.Reset(payload)
			w := codec.NewWriter(buf)

			_, err := io.Copy(w, r)
			if err != nil {
				b.Fatal(err)
			}
			if err := w.Close(); err != nil {
				b.Fatal(err)
			}
		}
	})

	return 1 - (float64(buf.Len()) / float64(len(payload)))
}
```

```
λ go test -v -run '^$' -bench Compress
goos: windows
goarch: amd64
BenchmarkCompression/none/compress-32              14197             80739 ns/op        24033.78 MB/s
BenchmarkCompression/gzip/compress-32                 51          21985308 ns/op          88.26 MB/s
BenchmarkCompression/snappy/compress-32             1110           1064202 ns/op        1823.41 MB/s
BenchmarkCompression/lz4/compress-32                 514           2313747 ns/op         838.67 MB/s
BenchmarkCompression/zstd/compress-32                420           2607726 ns/op         744.12 MB/s
input => 1.85 MB
  none:         0.00%
  gzip:         93.86%
  snappy:       86.37%
  lz4:          89.55%
  zstd:         93.10%
```

klauspost:
As a sidenote, since you already have the dependency, swapping out `compress/gzip` with `github.com/klauspost/compress/gzip` gives these numbers:

```
BenchmarkCompression/gzip/compress-32                160           7495428 ns/op         258.89 MB/s
gzip:         93.03%
```

achille-roussel:
Nice! Thanks for the tip on `compress/gzip`!

The compression ratio is different in your benchmarks, is it due to using a different default compression level? Or is it due to different trade offs in the implementation?

klauspost:
Yes, it is for "default" to be a more reasonable default in terms of speed/size tradeoff.

> This package attempts to provide a more smooth transition, where "1" is taking a lot of shortcuts, "5" is the reasonable trade-off and "9" is the "give me the best compression", and the values in between gives something reasonable in between. The standard library has big differences in levels 1-4, but levels 5-9 having no significant gains - often spending a lot more time than can be justified by the achieved compression.

https://blog.klauspost.com/rebalancing-deflate-compression-levels/

But it also has a lot other optimizations, for example in-compressible data being skipped more than 50x faster.

pascaldekloe:
To be clear, compression level 3 is the only compression level supported with the Go library. Other values will silently default to 3.

The compressed size win for level 5 is highly unlikely to exceed 20%. Kafka works well with fully packed disks thanks to the low fragmentation by design. I don't think anyone will run out of space unexpectedly due to this loss.

Those debug lines are more or less harmless.

```go
func printf(format string, a ...interface{}) {
        if debug {
                log.Printf(format, a...)
        }
}
```

The benchmark result in the description were measured on a mid-range iMac from 2017. I had a quick look, and most of the slowdown comes from the high amount of data copies. The decoder does not make use the advertised frame size. Also, the data parsing uses abstract methods, which prevent all sorts of optimisations. This is all fixable in the long run of course.

klauspost:
> To be clear, compression level 3 is the only compression level supported with the Go library.

(assuming you talk about zstd). A level 1 equivalent is also available. `zst.SpeedFastest`. But I see now you have a mixup in your code. I will add a comment.

zstd/zstd.go
```diff
@@ -1,33 +1,30 @@
...
 	kafka.RegisterCompressionCodec(NewCompressionCodec())
 }
 
-const (
-	Code = 4
+const Code = 4
 
-	DefaultCompressionLevel = zstd.DefaultCompression
-)
+const DefaultCompressionLevel = int(zstdlib.SpeedDefault)
```
klauspost: You cannot do this cast. The value is not interchangeable.

pascaldekloe: I don't understand what you mean? 🤔

pascaldekloe: Ah, now DefaultCompressionLevel is an integer, and before it was a plain constant, so this could be incompatible, right?

klauspost: Below you are doing `zstdlib.EncoderLevelFromZstd(level)` which assumes a reference zstd value.  

`zstdlib.SpeedDefault` has nothing to do with that. You can use `3` which is the reference library default value.

pascaldekloe: 😳 Now it makes even less sense.

```go
func EncoderLevelFromZstd(level int) EncoderLevel {
        switch {
        case level < 3:
                return SpeedFastest
        case level >= 3:
                return SpeedDefault
        }
        return SpeedDefault
}
```

Are you saying I can't use an int(SpeedDefault) as an argument to that function?

pascaldekloe: Right, so EncoderLevel has its *own* scale for some reason. 🤨Thanks for mentioning that. 👍

go.mod
klauspost: v1.9.8 is released which removes this dependency. Incidentally it also has faster/better gzip.

zstd/zstd.go
klauspost: Ping

pascaldekloe: Meh. I don't think a defer should be used with two lines of logic in place.  A swap makes the "unset" a bit less clear. Feel free to correct me @achille-roussel.

pascaldekloe:
So what should we do? 

### Conns
* lacks bounds checks (which could cause panics)
* performance worse than snappy
* limited compression levels

### Pro's
* no more C bindings
* active maintenance

stevevls:
@pascaldekloe Do you have an updated set of benchmarks now that you've changed the compression level and upgraded the library?

Also, zstd is expected to be slower than snappy, so I wouldn't list that as a con.  The compression ratio should be very favorable, though, so it's all about tradeoffs. 😄

pascaldekloe:
```
goos: darwin
goarch: amd64
BenchmarkCompression/none/compress-4         	   14690	     81149 ns/op	23912.51 MB/s
BenchmarkCompression/none/decompress-4       	   30531	     38266 ns/op	50710.10 MB/s
BenchmarkCompression/gzip/compress-4         	      60	  19582193 ns/op	   6.08 MB/s
BenchmarkCompression/gzip/decompress-4       	     442	   2682437 ns/op	 723.40 MB/s
BenchmarkCompression/snappy/compress-4       	     897	   1302823 ns/op	 203.01 MB/s
BenchmarkCompression/snappy/decompress-4     	    1507	    779842 ns/op	2488.29 MB/s
BenchmarkCompression/lz4/compress-4          	     370	   3211224 ns/op	  63.17 MB/s
BenchmarkCompression/lz4/decompress-4        	     878	   1342304 ns/op	1445.63 MB/s
BenchmarkCompression/zstd/compress-4         	     314	   3658719 ns/op	  35.74 MB/s
BenchmarkCompression/zstd/decompress-4       	     493	   2379608 ns/op	 815.46 MB/s
input => 1.85 MB
  none:		0.00%
  gzip:		93.86%
  snappy:	86.37%
  lz4:		89.55%
  zstd:		93.26%
```

zstd/zstd.go
```diff
@@ -38,161 +35,91 @@ func (c *CompressionCodec) Name() string { return "zstd" }
...
-func (r *reader) Read(b []byte) (int, error) {
-	if err := r.decompress(); err != nil {
-		return 0, err
-	}
-
-	if r.offset >= len(r.buffer.output) {
-		return 0, io.EOF
+// Close implements the io.Closer interface.
+func (r *reader) Close() error {
+	if r.err != io.ErrClosedPipe {
```
achille-roussel: This looks like a risky check, other code paths could be setting `r.err` to `io.ErrClosedPipe`.

How about a change like this?
```go
if r.dec != nil {
    decPool.Put(r.dec)
    r.dec = nil
    r.err = io.ErrClosedPipe
}
```

zstd/zstd.go
```diff
@@ -38,161 +35,91 @@ func (c *CompressionCodec) Name() string { return "zstd" }
...
-func (r *reader) WriteTo(w io.Writer) (int64, error) {
-	if err := r.decompress(); err != nil {
-		return 0, err
-	}
-
-	if r.offset >= len(r.buffer.output) {
-		return 0, nil
+// Read implements the io.Reader interface.
+func (r *reader) Read(p []byte) (n int, err error) {
+	println("zstd read", len(p))
```
achille-roussel: Should we remove this?

zstd/zstd.go
achille-roussel: I think @klauspost was referring to the call to `Close` after the encoder was given back to the pool. If the encoder gets reused before the `Close` call is executed, it would result in race conditions and unpredictable behaviors.

I'd have the same comment as for the reader path here, let's not use the `w.err` field as the sentinel for whether the encoder was closed, because other code paths may be setting this field as well. How about replacing it with something like this?
```go
if w.enc != nil {
    w.enc.Close()
    encPool.Put(w.enc)
    w.enc = nil
    w.err = io.ErrClosedPipe
}
```

zstd/zstd.go
```diff
@@ -38,161 +35,91 @@ func (c *CompressionCodec) Name() string { return "zstd" }
...
-	writer io.Writer
-	buffer *buffer
-	level  int
+	return p
 }
 
-func (w *writer) Write(b []byte) (int, error) {
-	return w.buffer.write(b)
-}
+var encPool sync.Pool
```
achille-roussel: Do we need to set the `New` function on this pool?

zstd/zstd.go
```diff
@@ -38,161 +35,91 @@ func (c *CompressionCodec) Name() string { return "zstd" }
...
-// The DataDog/zstd package exposes io.Writer and io.Reader implementations that
-// encode and decode streams, however there are no APIs to reuse the values like
-// other compression format have (through a Reset method usually).
-//
-// I first tried using these abstractions but the amount of state that gets
-// recreated and destroyed was so large that it was slower than using the
-// zstd.Compress and zstd.Decompress functions directly. Knowing that, I changed
-// the implementation to be more of a buffer management on top of these instead.
-// =============================================================================
+var decPool sync.Pool
```
achille-roussel: Do we need to set the `New` function on this pool?

zstd/zstd.go
```diff
@@ -38,161 +35,91 @@ func (c *CompressionCodec) Name() string { return "zstd" }
 
 // NewReader implements the kafka.CompressionCodec interface.
 func (c *CompressionCodec) NewReader(r io.Reader) io.ReadCloser {
-	return &reader{
-		reader: r,
-		buffer: bufferPool.Get().(*buffer),
+	p := new(reader)
+	if cached := decPool.Get(); cached == nil {
```
achille-roussel: I think if we set the `New` function on the pool this condition becomes impossible (`nil` will never be returned).

You may have found the error returned by `zstd.NewReader` challenging to handle in the context of the pool constructor, see how we've solved that in the [`kafka-go/gzip`](https://github.com/segmentio/kafka-go/blob/master/gzip/gzip.go#L13-L31) case, I think you can mirror it here.

klauspost:
Looking at the numbers the benchmarks still looks wrong. Try to compare to the "clean" one I posted above. Maybe just split encode/decode benchmarks, since simpler may be better in this case.

LZ4 should perform on level with Snappy, and only the fastest level is really worth it on lz4 (using pierrec library). The rest are slow and a waste of CPU. I would propose to set the default to that.

zstd/zstd.go
pascaldekloe: You're right. :-)

zstd/zstd.go
pascaldekloe: Yes!

zstd/zstd.go
pascaldekloe: Not required. The default `nil` value helps with applying the Writer.

zstd/zstd.go
pascaldekloe: Not required. The default nil value helps with applying the Writer.

zstd/zstd.go
pascaldekloe: That is exactly why I don't set the New function. The `nil` value is documented behaviour of sync.Pool.

zstd/zstd.go
pascaldekloe: That's much better indeed.

achille-roussel:
> LZ4 should perform on level with Snappy, and only the fastest level is really worth it on lz4 (using pierrec library). The rest are slow and a waste of CPU. I would propose to set the default to that.

Thanks for the tip! Let's make this another follow up to this PR as well to keep the focus on Zstd here 👍

Issue #364, opened 2019-10-16
Try to send entire batch before returning any MessageTooLargeErrors in Writer.WriteMessages


Evanjt1:
Currently, `Writer.WriteMessages` returns early if it encounters a message that is too large: https://github.com/segmentio/kafka-go/blob/5e37e0483bf9aac5ec283e3a76086e0289952b97/writer.go#L322

Since messages may have already been added to `w.msgs` before encountering the `MessageTooLargeError`, wouldn't it be better to try and submit all messages in the batch before returning this error? I have a feeling that the early return is due to only one error being returned from a `WriteMessages` call so ensuring this error gets returned allows the caller to handle appropriately.

Is there some other edge case that I'm missing here for why the receiver returns early when encountering this issue?

Linking #363 since it's related to the singular return from `WriteMessages`

Obligating myself to implement this, if needed.

achille-roussel:
Hey @Evanjt1, thanks for raising the question.

You're right that the current behavior isn't ideal, I think we should either attempt to send as many messages as possible (all messages that aren't too large), or do a check prior to sending the messages and either send them all or send nothing.

@stevevls you may have an opinion on this as well.

Evanjt1:
@achille-roussel I typically lean towards processing as much as possible and rejecting only the malformed/too large messages rather than failing out the entire batch. I perused a few clients for other systems that have similar batch send, and that seems to be the preferred way to go.

Issue #363, opened 2019-10-16
Return all errors that occur in Writer.WriteMessages


Evanjt1:
Currently, the `WriteMessages` receiver only returns a singular error if any message in a given batch fails to deliver to kafka. It's great to know that something went wrong, but it would even better to have information on a per message level.

There are a couple of limitations with the current single error signature:
- The error returned has the potential to be misleading. The single error in some cases is really just the last error read from the `res` channel. The caller ends up losing valuable information about other errors which may have occurred in the batch limiting error handling logic.
- The caller has no choice but to treat that entire batch as a failure.

The second bullet point is especially poignant for me. As an example, If I want to write the failed messages to a dead letter queue, I'm forced to write them all.

Here are a couple different approaches for making a more useful return for the receiver:

1. Use a library such as https://github.com/hashicorp/go-multierror to return all `writerError`s that occurred for the batch. I'm not sold on introducing a dependency, but the solution would be pretty straightforward.

2. Something along the lines of:
```go
func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) []WriterError

// WriterError is just the current writerError exposed publicly
type WriterError struct {
    msg Message
    err error
} 
```
Although, I'm not a big fan of returning error slices. We could always create a new error type to wrap the WriterErrors with some receivers to allow the caller to dig deeper into the errors if one so chooses. This would also allow us to preserve the signature.

Thoughts?

I'll obligate myself to implementing whatever solution is decided upon if this feature request is deemed useful.

achille-roussel:
Hello @Evanjt1, thanks for opening an issue!

The idea of exposing every error sounds good to me, but we can't make a breaking API change like returning a `[]WriterError` instead of an `error`.

I agree that we probably don't want to take an external dependency for this. The code to implement a multi-error type should be simple, and we can use a type that implements the `error` interface to add the feature without changing the API I think.

For example:
```go
type MessageErrorList []MessageError

func (e MessageErrorList) Error() string { ... }

type MessageError struct {
    Msg Message
    Err error
}
```
Then programs can use type assertion on the error returned by `WriteMessages` if they're interested in getting the detailed information (otherwise they can keep using the returned `error` value).

Let me know what you think about it!

Evanjt1:
@achille-roussel Agreed, I think that's the best way to go about this. Cool if I implement this?

Evanjt1:
@achille-roussel I'm finally revisiting this, anything important I should know before putting in PR?

[...]

PR #401, opened 2020-02-11
Return multiple errors from Writer.WriteMessages and MessageTooLargeError handling improvements

Evanjt1:
Addresses issues #363 and #364 

This PR revolves around 2 main improvements for the Writer type:
- Return information about all errors that occurred during a write of multiple messages while still preserving interface of `Writer.WriteMessages`
- Try to send all acceptable messages in a batch before returning the ones that are too large

I introduced two new public types, `WriterError` and `WriterErrors` to implement the first improvement. Both implement the error interface. Under the hood `Writer.WriteMessages` returns `WriterErrors`. The caller can cast the error returned as `WriterErrors` to use the new functionality if they so choose.

The second improvement is addressed by skipping messages that are too large as the `Writer` comes across them while sending to the `msgs` channel. They are instead stored as errors that will be returned after all the other messages have been tried.  

New tests have been added to `writer_test.go` to ensure these features behave as expected. Additionally, some of the preexisting tests have been updated so they also check the `WriterErrors` returned.

There would have been a bunch of merge conflicts and wasted work if I tried to made both improvements individually so I decided to knock out two birds with one stone.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"math/rand"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 )
@@ -148,6 +149,24 @@ func (e *WriterError) Temporary() bool {
 
 func (e *WriterError) Timeout() bool {
 	return isTimeout(e.Err)
+}
+
+type WriterErrors []WriterError
+
+func (es WriterErrors) Error() string {
+	if len(es) == 1 {
+		return fmt.Sprintf("1 WriterError occurred:\n\t* %s\n", es[0].Err)
+	}
+
+	points := make([]string, len(es))
+	for i, we := range es {
+		points[i] = fmt.Sprintf("* %s", we.Err)
+	}
+
+	return fmt.Sprintf(
+		"%d WriterErrors occurred:\n\t%s\n",
+		len(es),
+		strings.Join(points, "\n\t"))
 }
 
 // WriterStats is a data structure returned by a call to Writer.Stats that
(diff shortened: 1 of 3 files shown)


writer.go
```diff
@@ -133,6 +134,41 @@ type WriterConfig struct {
 	newPartitionWriter func(partition int, config WriterConfig, stats *writerStats) partitionWriter
 }
 
+type WriterError struct {
+	Msg Message
+	Err error
+}
```
achille-roussel: Could we have the `WriterError` and `WriterErrors` type implement `errors.Wrapper`?

Evanjt1: Will do, `Unwrap` is straightforward for `WriterError`. Do you have thoughts on behavior for `WriterErrors`?

Evanjt1:
Should I take care of merge conflicts as well?

writer.go
Evanjt1: @achille-roussel For `WriterErrors.Unwrap()` how do you feel about using the same strategy as https://github.com/prometheus/client_golang/blob/82ce871c27ca8919bcdb97d69bcf461499171b27/prometheus/registry.go#L240?

Evanjt1:
@achille-roussel This PR is pretty out-of-date at this point and your work on #438 covers the issues this PR was about so I think this should probably be closed.

achille-roussel:
Thanks for follow up @Evanjt1, I'll go ahead and close. Feel free to reopen an issue if the problem isn't properly addressed.

PR #408, opened 2020-02-26
Public topic scanner interface

varsanojidan:
Related to #388.
A public interface for a scanner to scan for all topics matching a specified regex

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1020,6 +1020,37 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
 	return
 }
 
+// Lists all topics matching specified regex
+func (c *Conn) ListTopics() (topics []string, err error) {
+
+	err = c.readOperation(
+		func(deadline time.Time, id int32) error {
+			return c.writeRequest(metadataRequest, v1, id, topicMetadataRequestV1(nil))
+		},
+		func(deadline time.Time, size int) error {
+			var res metadataResponseV1
+			topics = []string{}
+			if err := c.readResponse(size, &res); err != nil {
+				return err
+			}
+
+			for _, t := range res.Topics {
+
+				if t.TopicErrorCode != 0 && (c.topic == "" || t.TopicName == c.topic) {
+					// We only report errors if they happened for the topic of
+					// the connection, otherwise the topic will simply have no
+					// partitions in the result set.
+					return Error(t.TopicErrorCode)
+				}
+				topics = append(topics, t.TopicName)
+
+			}
+			return nil
+		},
+	)
+	return
+}
+
 // Write writes a message to the kafka broker that this connection was
 // established to. The method returns the number of bytes written, or an error
 // if something went wrong.
--- a/topicscanner.go
+++ b/topicscanner.go
@@ -1,0 +1,115 @@
+package kafka
+
+import (
+	"context"
+	"regexp"
+	"sync"
+	"time"
+)
+
+var (
+	defaultScanningIntervalMS = 10000
+)
+
+type (
+	TopicScannerConfig struct {
+		Brokers            []string
+		TopicRegex         string
+		ScanningIntervalMS int
+	}
+
+	TopicScanner struct {
+		config             TopicScannerConfig
+		currentFoundTopics *[]string
+		closeChan          chan struct{}
+		regex              regexp.Regexp
+		mutex              sync.Mutex
+	}
+)
+
+func NewTopicScanner(config TopicScannerConfig) (*TopicScanner, error) {
+
+	rgx, err := regexp.Compile(config.TopicRegex)
+	if err != nil {
+		return nil, err
+	}
+	//If the scanning interval isn't set properly, scan every 10 seconds
+	if config.ScanningIntervalMS <= 0 {
+		config.ScanningIntervalMS = defaultScanningIntervalMS
+	}
+	return &TopicScanner{
+		currentFoundTopics: new([]string),
+		config:             config,
+		regex:              *rgx,
+		closeChan:          make(chan struct{}),
+	}, nil
+}
+
+func (t *TopicScanner) StartScanning(ctx context.Context) {
+	//Initial fetching of the topics
+	t.updateTopics(ctx)
+	go func() {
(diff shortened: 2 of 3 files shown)


Programmer19:
Any updates on this? Should there be any changes or can we get this merged in?

varsanojidan:
Any update on this @achille-roussel

jbguerraz:
We also need wildcard support. We'll be happy to help. @achille-roussel  what's the status by now ? :)

alok87:
When can this get live?

hsawhney09:
This is cool .. can this please be merged in and released

conn.go
```diff
@@ -1020,6 +1020,37 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
 	return
 }
 
+// Lists all topics matching specified regex
+func (c *Conn) ListTopics() (topics []string, err error) {
```
nlsun, line 1024: Could we list topics through `ReadPartitions` instead? Something similar to what's in here https://github.com/segmentio/kafka-go/pull/521

```
partitions, err := conn.ReadPartitions()
if err != nil {
    return err
}
m := map[string]struct{}{}
for _, p := range partitions {
    m[p.Topic] = struct{}{}
}
```

We could have this as a private function in topicscanner.go as well

topicscanner.go
```diff
@@ -0,0 +1,115 @@
...
+				newTopics = append(newTopics, topic)
+			}
+		}
+		t.mutex.Lock()
+		t.currentFoundTopics = &newTopics
+		t.mutex.Unlock()
+	}
+}
+
+func (t *TopicScanner) initialize(ctx context.Context) (conn *Conn, err error) {
```
nlsun, line 102: I think it'd be more flexible if we allow the caller to pass in the `Conn` of their choosing instead?

topicscanner.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (t *TopicScanner) GetTopics() []string {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+	return *t.currentFoundTopics
+}
+
+func (t *TopicScanner) Close() {
+	t.closeChan <- struct{}{}
```
nlsun, line 75: we can close instead of sending on this channel to indicate that we're done

nlsun, line 75: we might also want to wait on a channel here (that is closed by the main loop) so we return when the loop actually exits

topicscanner.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"context"
+	"regexp"
+	"sync"
+	"time"
+)
+
+var (
+	defaultScanningIntervalMS = 10000
```
nlsun, line 11: we can use a `time.Duration` here instead of a number

topicscanner.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if len(topics) > 0 {
+		newTopics := []string{}
+		for _, topic := range topics {
+			if t.regex.MatchString(topic) {
+				newTopics = append(newTopics, topic)
+			}
+		}
+		t.mutex.Lock()
+		t.currentFoundTopics = &newTopics
```
nlsun, line 97: it'd be nice to move this update (with mutex handling) to a helper function

also to make a copy of slice rather than a reference
```
t.currentFoundTopics = newTopics[:]
```

topicscanner.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+		}
+
+	}()
+}
+
+func (t *TopicScanner) GetTopics() []string {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+	return *t.currentFoundTopics
```
nlsun, line 71: would be nice to return a copy instead of the original value
```
return t.currentFoundTopics[:]
```

topicscanner.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return
+	}
+	topics, err := conn.ListTopics()
+	if err != nil {
+		return
+	}
+	conn.Close()
+
+	if len(topics) > 0 {
```
nlsun, line 89: i think we could check this just for the actual update of `currentFoundTopics` and not for the rest of the logic

nlsun (commented):
Some high level comments

topicscanner.go
```diff
@@ -0,0 +1,115 @@
...
+func (t *TopicScanner) StartScanning(ctx context.Context) {
+	//Initial fetching of the topics
+	t.updateTopics(ctx)
+	go func() {
+		ticker := time.NewTicker(time.Duration(t.config.ScanningIntervalMS) * time.Millisecond)
+	outer:
+		for {
+			select {
+			case _ = <-ticker.C:
+				t.updateTopics(ctx)
```
nlsun, line 57: we probably want "do-while" loop semantics, so we update topics immediately upon startup

that could look something like
```
for {
    t.updateTopics(ctx)

    select {
    case <-ticker.C:
        // noop

    ...

    }
}
```

topicscanner.go
nlsun, line 11: what do you think of relaxing the interval to once per minute?

conn.go
aultimus: It seems like there is some demand for a public ListTopics function given the other PRs open to add that functionality, so perhaps we should add a public function here. I would be interested to understand the relative trade off between these differing implementations, using c.readOperation() and conn.ReadPartitions().

aultimus:
@varsanojidan is it possible for you to address these review comments?

varsanojidan:
@aultimus yes, apologies, I've been busy, Ill address the comments soon

abonec:
Am i right, that this PR address only to topic discovery, not for reading multiple topics at once and kafka-go still can't do this?

topicscanner.go
godwhoa: FWITW, both the Java client and rdkafka refresh topic metadata every 5 minutes by default.

nlsun:
@varsanojidan Hi, are you still interested in implementing this feature?

nlsun:
This might be superceded by https://github.com/segmentio/kafka-go/pull/663

nlsun:
This was implemented as a sub-package `topics` as of https://github.com/segmentio/kafka-go/pull/663

PR #409, opened 2020-02-29
Don't close connection when end of batch is read correctly

stevevls:
io.EOF is used as a marker that a batch of messages has been
completely read.  The batch will repackage an io.EOF from the
underlying connection as an io.ErrUnexpectedEOF.

The previous behavior closed the connection on an io.EOF and
restarted the reader loop, which will be much less efficient than
continuing to use the connnection.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1196,7 +1196,7 @@ func (r *reader) run(ctx context.Context, offset int64) {
 				// block relies on the batch repackaging real io.EOF errors as
 				// io.UnexpectedEOF.  otherwise, we would end up swallowing real
 				// errors here.
-				break readLoop
+				errcount = 0
 			case UnknownTopicOrPartition:
 				r.withErrorLogger(func(log Logger) {
 					log.Printf("failed to read from current broker for partition %d of %s at offset %d, topic or parition not found on this broker, %v", r.partition, r.topic, offset, r.brokers)


jiachinzhao:
from my test and code review, I found the reason it returns io.EOF after every batch  is here,  
https://github.com/segmentio/kafka-go/blob/f490bbc5ee1f4fe96e83b9f3efa1a971e8de500e/batch.go#L247
```go
        offset, timestamp, headers, err = batch.msgs.readMessage(batch.offset, key, val)
	switch err {
	case nil:
		batch.offset = offset + 1
	case errShortRead:
		// As an "optimization" kafka truncates the returned response after
		// producing MaxBytes, which could then cause the code to return
		// errShortRead.
		err = batch.msgs.discard()
		switch {
		case err != nil:
			// Since io.EOF is used by the batch to indicate that there is are
			// no more messages to consume, it is crucial that any io.EOF errors
			// on the underlying connection are repackaged.  Otherwise, the
			// caller can't tell the difference between a batch that was fully
			// consumed or a batch whose connection is in an error state.
			batch.err = dontExpectEOF(err)
		case batch.msgs.remaining() == 0:
			// Because we use the adjusted deadline we could end up returning
			// before the actual deadline occurred. This is necessary otherwise
			// timing out the connection for real could end up leaving it in an
			// unpredictable state, which would require closing it.
			// This design decision was made to maximize the chances of keeping
			// the connection open, the trade off being to lose precision on the
			// read deadline management.
			err = checkTimeoutErr(batch.deadline)
			batch.err = err
		}
	default:
		// Since io.EOF is used by the batch to indicate that there is are
		// no more messages to consume, it is crucial that any io.EOF errors
		// on the underlying connection are repackaged.  Otherwise, the
		// caller can't tell the difference between a batch that was fully
		// consumed or a batch whose connection is in an error state.
		batch.err = dontExpectEOF(err)
	}
```

Since  batch.msgs.readMessage return errShortRead and batch.msgs.remaining() == 0 and read did not timeout, then  checkTimeoutErr return io.EOF.
why   batch.msgs.readMessage return errShortRead, finally I found it's due to batch.msgs.remaining() == 0 from here
 https://github.com/segmentio/kafka-go/blob/f490bbc5ee1f4fe96e83b9f3efa1a971e8de500e/message.go#L479

So I have a question, since we already know remain bytes to read, shouldn't we break ReadMessage by judge remain bytes?  https://github.com/segmentio/kafka-go/blob/f490bbc5ee1f4fe96e83b9f3efa1a971e8de500e/reader.go#L1366

another library github.com/Shopify/sarama/ had Similar logic.

https://github.com/Shopify/sarama/blob/b5764af1c47d0f6718dba3be6a3d75e8c97b351a/message_set.go#L70

tomeliaz:
Testing completed successfully locally. No issues noted.

PR #414, opened 2020-03-12
Skip control message

jame2981:
If not consumer receive empty message.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -221,7 +221,7 @@ func (batch *Batch) readMessage(
 
 	offset, timestamp, headers, err = batch.msgs.readMessage(batch.offset, key, val)
 	switch err {
-	case nil:
+	case nil, errControlMessage:
 		batch.offset = offset + 1
 	case errShortRead:
 		// As an "optimization" kafka truncates the returned response after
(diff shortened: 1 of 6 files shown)


dominicbarnes:
I don't believe we have encountered control messages in our usage of Kafka, but that seems to be because we haven't really configured for it. We found some documentation about [Control Batches](https://kafka.apache.org/documentation/#controlbatch) which seems to be related to this and can help us handle them appropriately.

A feature like this would benefit a lot from some unit tests, so could you maybe share some more about your Kafka setup so we can reproduce this in our test suite?

dominicbarnes:
From my understanding, the problem here is that these control messages end up manifesting with empty values, which can be problematic for applications. The only reason these control messages are generated at all is when transactions in use, so it seems very risky to work around this by simply ignoring these types of messages.

This exposes a functional gap in kafka-go, as we don't have any true support for transactions, which we already have an issue for tracking with #216.

As a workaround, your application/consumer code can check for empty messages and ignore them as you're trying to accomplish here. The PR is very much appreciated, but I think ultimately we should try to address this problem with transaction support.

PR #429, opened 2020-04-14
Fixed #428 by making reader.read respect context closure

### LINKED ISSUE #428: Reader.Close hangs for ReaderConfig.MaxWait
**Describe the bug**
After reading once, `Reader.Close` hangs until the end of a `MaxWait` duration.

**Kafka Version**
2.3.0

**To Reproduce**
```go
package main

import (
	"context"
	"fmt"
	"time"

)

func main() {
		Brokers: []string{"localhost:9092"}, // assuming a running Kafka instance
		Topic:   "test",
		MaxWait: time.Hour,
	})

	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()

	fmt.Println("Trying to read...")
	_, _ = reader.FetchMessage(ctx)

	fmt.Println("Closing reader...")
	reader.Close()

	fmt.Println("Test passed")
}
```

**Expected behavior**
`Reader.Close` should close the connection and return immediately.

feldgendler:
Closes segmentio/kafka-go#428.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1337,6 +1337,30 @@ func (r *reader) initialize(ctx context.Context, offset int64) (conn *Conn, star
 	return
 }
 
+// readBatch wraps the call to conn.ReadBatchWith to make it interruptible.
+// Conn methods are written in a non-interruptible style, so the only way to
+// interrupt them is to close the connection in another goroutine.
+func (r *reader) readBatch(ctx context.Context, conn *Conn) (*Batch, error) {
+	done := make(chan struct{})
+	defer close(done)
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			conn.Close()
+		case <-done:
+			return
+		}
+	}()
+
+	batch := conn.ReadBatchWith(ReadBatchConfig{
+		MinBytes:       r.minBytes,
+		MaxBytes:       r.maxBytes,
+		IsolationLevel: r.isolationLevel,
+	})
+	return batch, ctx.Err()
+}
+
 func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, error) {
 	r.stats.fetches.observe(1)
 	r.stats.offset.observe(offset)
@@ -1344,18 +1368,17 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
 	t0 := time.Now()
 	conn.SetReadDeadline(t0.Add(r.maxWait))
 
-	batch := conn.ReadBatchWith(ReadBatchConfig{
-		MinBytes:       r.minBytes,
-		MaxBytes:       r.maxBytes,
-		IsolationLevel: r.isolationLevel,
-	})
+	batch, err := r.readBatch(ctx, conn)
+	if err != nil {
+		return offset, err
+	}
+
 	highWaterMark := batch.HighWaterMark()
 
 	t1 := time.Now()
 	r.stats.waitTime.observeDuration(t1.Sub(t0))
 
 	var msg Message
-	var err error
 	var size int64
 	var bytes int64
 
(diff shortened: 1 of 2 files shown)


feldgendler:
@achille-roussel Hi! Could you please look at this fix?

achille-roussel:
Thanks for the contribution!

Spawning a goroutine for each batch we read from kafka seems like it could add significant overhead for high throughput readers. I feel like we would want to measure the impact before merging this change.

May I ask which value you're using for MaxWait? Typically I've seen programs use values in the 500ms-1s range, which is why I originally assumed that blocking Close for that duration would be fine (since it often happens when the program exits).

feldgendler:
I set `MaxWait` to a very large value, essentially infinity. I don't see why I'd want `FetchMessage` to return empty-handed. Whenever it did, I would just restart it because I've got nothing better to do.

I guess I could set it to 1s and keep restarting `FetchMessages` (although that sounds like a lot of needless activity), but it would cause a problem in our unit tests. We use `kafka-go` in a complex component that we create and destroy for each unit test. I wouldn't want the teardown of the environment to block each of the tests for a second.

baconalot:
I had a similar issue where MaxWait = 5 * time.Minutes would make the reader hang on a rebalance. I set this MaxWait this high because I do not need to be hammered by EOF messages when I am at the head of a topic/partition. (Using groups with multiple topics)

feldgendler:
@achille-roussel Hey, any chance for this to be merged?

achille-roussel:
I'll take another look at it this week.

Antonboom:
+1

feldgendler:
@achille-roussel any news? Sorry to bother, but this is important for us, and it's bothersome to live with a local package replacement.

reader_test.go
```diff
@@ -1313,3 +1313,24 @@ func getOffsets(t *testing.T, config ReaderConfig) offsetFetchResponseV1 {
...
+	r := NewReader(ReaderConfig{
+		Brokers: []string{"localhost:9092"},
+		Topic:   makeTopic(),
+		MaxWait: 2 * time.Second,
+	})
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	_, _ = r.FetchMessage(ctx)
```
Antonboom, line 1329: err handling?

and second `r.Close` via defer?

feldgendler: Fixed.

feldgendler:
@Antonboom Does it look mergeable now?

Antonboom:
> @Antonboom Does it look mergeable now?

@feldgendler, I'm not the owner of repo :(
I'm also interested in this fix and was just passing by.

@achille-roussel, merge please...

reader.go
```diff
@@ -1337,25 +1337,48 @@ func (r *reader) initialize(ctx context.Context, offset int64) (conn *Conn, star
...
-	t0 := time.Now()
-	conn.SetReadDeadline(t0.Add(r.maxWait))
+	go func() {
+		select {
+		case <-ctx.Done():
+			conn.Close()
+		case <-done:
+			return
+		}
+	}()
```
achille-roussel, line 1354: We would need to wait for this goroutine to finish before returning from the `readBatch` function, otherwise the goroutine will execute asynchronously and may close the connection when `ctx` is canceled after the function returned.

achille-roussel, line 1354: I'm concerned that having this goroutine spawn for every batch read by the consumer may greatly increase pressure on the Go scheduler and GC, which could impact performance of high throughput consumers.

Could we either add a benchmark to evaluate the impact of this change?

achille-roussel (commented):
Thanks for your contribution!

Generally this looks good, but the overhead may not be practical for high throughput readers, which tend to use low value MaxWait and therefore aren't impacted by the closure waiting on inflight requests. We would need to validate that we aren't introducing performance regressions with this change before we merge it in.

Issue #227, opened 2019-03-11
Endless waiting for consumer connection


pshopper:
**Describe the bug**
If kafka unavailable (connection refused or other error), refreshCoordinator func performs endless.

**To Reproduce**
Try to fetch message (by reader.FetchMessage with consumer group) with unavailable kafka.

**Expected behavior**
Reader will return error and logging this.

reprah:
@achille-roussel I'm interested in helping with this issue. If we just want to log that we got a dial timeout, we could log the error inside `refreshCoordinator()` [here](https://github.com/segmentio/kafka-go/blob/3a0e564053e187bc9c18a4e69966d89a4fd4499a/reader.go#L161) and perhaps return the error instead of continuing to the next iteration of the loop. Thoughts?

achille-roussel:
@reprah thanks for offering to help.

Logging the error seems like a good first step, let me know if you're available to make the change 👍

alamages:
Hello,

I see the code has change quite a bit since this issue was reported but I would like to mention that this is still an issue.

This is now [the line that causes the problem](https://github.com/segmentio/kafka-go/blob/48f0c131b6a036cfeff519cc8d6ff09de3141d9e/consumergroup.go#L576). The `nextGeneration` function returns an error if a connection can not be established successfully (e.g invalid hostname, wrong port, wrong credentials etc). The error is not really handled but instead is ignored and the for{} loop will try after some "backoff" time again, infinitely. 

The only way to figure out whether there is an error is to pass an error logger in the reader config and look at the logs. This might be good for now but I would expect the consumer to fail if a connection can not be established due to invalid configuration.

Cheers!

xiaopeifeng:
what's the PR status now?

stevevls:
This should be fixed on `master` now thanks to #430 .

PR #430, opened 2020-04-16
Fix Endless waiting for consumer connection (#227)

dcepelik:
When the high-level Reader API is used, e.g. when calling
Reader.FetchMessage, and the Reader is configured with a non-empty
GroupID, make sure that the call returns an error when the call to
ConsumerGroup.Next fails. This can happen, for example, when the Kafka
server isn't reachable.

The call will fail only after Reader.config.MaxAttempts attempts
calling ConsumerGroup.Next were performed internally by Reader.run
and failed.

Previously, Reader.FetchMessage blocked even when the Kafka cluster
could not possibly be reached, making it impractical to distinguish
between not getting messages because there are none and not getting
messages because we cannot connect to the server.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -59,6 +59,16 @@ type Reader struct {
 	lag     int64
 	closed  bool
 
+	// Without a group subscription (when Reader.config.GroupID == ""),
+	// when errors occur, the Reader gets a synthetic readerMessage with
+	// a non-nil err set. With group subscriptions however, when an error
+	// occurs in Reader.run, there's no reader running (sic, cf. reader vs.
+	// Reader) and there's no way to let the high-level methods like
+	// FetchMessage know that an error indeed occurred. If an error in run
+	// occurs, it will be non-block-sent to this unbuffered channel, where
+	// the high-level methods can select{} on it and notify the caller.
+	runError chan error
+
 	// reader stats are all made of atomic values, no need for synchronization.
 	once  uint32
 	stctx context.Context
@@ -253,8 +263,15 @@ func (r *Reader) run(cg *ConsumerGroup) {
 	})
 
 	for {
-		gen, err := cg.Next(r.stctx)
-		if err != nil {
+		// Limit the number of attempts at waiting for the next
+		// consumer generation.
+		var err error
+		var gen *Generation
+		for attempt := 1; attempt <= r.config.MaxAttempts; attempt++ {
+			gen, err = cg.Next(r.stctx)
+			if err == nil {
+				break
+			}
 			if err == r.stctx.Err() {
 				return
 			}
@@ -262,6 +279,17 @@ func (r *Reader) run(cg *ConsumerGroup) {
 			r.withErrorLogger(func(l Logger) {
 				l.Printf(err.Error())
 			})
+			// Continue with next attempt...
+		}
+		if err != nil {
+			// All attempts have failed.
+			select {
+			case r.runError <- err:
+				// If somebody's receiving on the runError, let
+				// them know the error occurred.
+			default:
+				// Otherwise, don't block to allow healing.
+			}
 			continue
 		}
 
@@ -531,7 +559,6 @@ type readerStats struct {
 // NewReader creates and returns a new Reader configured with config.
 // The offset is initialized to FirstOffset.
 func NewReader(config ReaderConfig) *Reader {
-
 	if err := config.Validate(); err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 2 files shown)


reader.go
```diff
@@ -59,6 +59,16 @@ type Reader struct {
...
 
+	// Without a group subscription (when Reader.config.GroupID == ""),
+	// when errors occur, the Reader gets a synthetic readerMessage with
+	// a non-nil err set. With group subscriptions however, when an error
+	// occurs in Reader.run, there's no reader running (sic, cf. reader vs.
+	// Reader) and there's no way to let the high-level methods like
+	// FetchMessage know that an error indeed occurred. If an error in run
+	// occurs, it will be non-block-sent to this unbuffered channel, where
+	// the high-level methods can select{} on it and notify the caller.
+	runError chan error
```
stevevls, line 70: I think it would be a good idea to make this a buffered channel of capacity 1.  Otherwise, there would be a decent chance of dropping the error.  In the case where e.g. the connect string is bad, it's likely that the caller is blocking on the `Reader` and the synchronous handoff via `runError` will work.  If something goes awry after messages have already started flowing, though, it's less likely that the client will be blocking because it may be either be processing a message or the buffered `r.msgs` channel will have messages queued.

dcepelik: My line of thinking was that this way, you'll only hit the error if it actually affects you, meaning you don't have anything to process, and you're not able to connect. E.g., if network was choppy but healed while you were processing a message, you will now (always) get the past error, though it's passé. When the channel isn't buffered, that won't happen.

But I agree that when you're in a known-wrong state (i.e., you exceeded your attempts inside Reader.run) it will take ReaderConfig.MaxAttempts (times the respective time-outs per attempt) to learn about that.

Perhaps a non-blocking read from buffered runError channel upon succeeding in Reader.run would be a good compromise between learning of the error promptly while not seeing past errors which are irrelevant?

I will address this plus the test tomorrow.

stevevls: It's an interesting case.  I wonder...if your reader can't maintain its membership with the consumer group, should you even be processing messages in the first place even if they're already buffered?  I would expect that the reader would have lost its assignments and that they'd have been given to another consumer already, so you may run the risk of duplicate processing because you wouldn't be able to commit offsets.

rjshrjndrn:
I didn't get the whole context, sorry to be a noob. But I'm facing the issue if with `GroupID` my client will wait for ever, else it'll start consuming message. Doesn't even know what's the error.

rjshrjndrn:
And when I used @dcepelik fork, got an error saying that
```
err reading message: [15] Group Coordinator Not Available: the broker returns this error code for group coordinator requests, offset commits, and most group management requests if the offsets topic has not yet been created, or if the group coordinator is not active
```
It's a huge help to figure out what's wrong.

dcepelik:
Hi, I've added two test cases. Let me know if anything else is needed on my part to get this merged.

reader_test.go
```diff
@@ -1313,3 +1313,45 @@ func getOffsets(t *testing.T, config ReaderConfig) offsetFetchResponseV1 {
...
 }
+
+const (
+	connTO     = 1 * time.Second
+	connTestTO = 2 * connTO
+)
+
+func TestErrorCannotConnect(t *testing.T) {
+	r := NewReader(ReaderConfig{
+		// FIXME
```
stevevls, line 1324: What needs to be fixed here?  We run tests using this docker-compose setup: https://github.com/segmentio/kafka-go/blob/master/docker-compose.yml.  The way I read it, I would expect this test to pass b/c port 9093 requires SASL authentication and would reply with an error.

One other question: if there's no consumer group ID set, are we exercising the logic being adding?

reader_test.go
```diff
@@ -1313,3 +1313,45 @@ func getOffsets(t *testing.T, config ReaderConfig) offsetFetchResponseV1 {
...
+	if err == nil || ctx.Err() != nil {
+		t.Errorf("Reader.FetchMessage must fail when it cannot " +
+			"connect")
+	}
+}
+
+func TestErrorCannotConnectGroupSubscription(t *testing.T) {
+	r := NewReader(ReaderConfig{
+		// FIXME
+		Brokers:     []string{"localhost:9093"},
```
stevevls, line 1343: What requires fixing here?  I think, like above, this test will pass b/c it will be unable to SASL authenticate.

MacTynow:
Hi @dcepelik, @stevevls, would you require any help on this to get it merged?

reader_test.go
dcepelik: Nothing, really. I think I just forgot to remove the FIXME.

reader_test.go
dcepelik: Same here.

dcepelik:
Hi @MacTynow @stevevls, I'm really sorry I didn't get to this earlier. I see no issues with the test, those FIXMEs we got stuck upon were no longer relevant IMO, or at least I don't see an issue now :).

dcepelik:
@stevevls This should hopefully be resolved now. Please let me know if anything else is needed.

PR #438, opened 2020-05-04
0.4

achille-roussel:
Hello friends! I believe this work is getting to the point where it could use a second pair of eyes.

I've been working on preparing a next major release (0.4) with the addition of long-wanted features:
* `kafka/protocol` is a new package containing a rewrite of the protocol implementation
* `kafka.Transport` provides the implementation of a connection manager and request routing layer which minimizes the number of open connections
* `kafka.Client`, along with `*Request` `*Response` types, provides a new API for interacting with kafka clusters, less oriented towards pub/sub use cases and mapping closer to the kafka RPC protocol

I tried to document and comment as much of the internal and external APIs, but this is still a large amount of code, happy to discuss live if it helps.

One thing to keep in mind is this work is grown "in parallel" to the existing implementations, I have not yet modified any of the internal implementations to leverage the new protocol types for example. The intent is to reduce the amount of changes we ship at once, and with it the risk of seeing unexpected changes in behavior.


time.go
```diff
@@ -11,18 +11,18 @@ const (
...
 	}
 	return t.UnixNano() / int64(time.Millisecond)
 }
 
-func timestampToTime(t int64) time.Time {
-	return time.Unix(t/1000, (t%1000)*int64(time.Millisecond))
-}
-
-func duration(ms int32) time.Duration {
+func makeDuration(ms int32) time.Duration {
```
achille-roussel, line 25: I renamed `duration` to `makeDuration` to unify the naming pattern of a few unexported functions.

record.go
```diff
@@ -0,0 +1,148 @@
...
+// used in both produce and fetch requests to represent the sequence of records
+// that are sent to or receive from kafka brokers.
+//
+// RecordSet values are not safe to use concurrently from multiple goroutines.
+type RecordSet interface {
+	io.Closer
+	// Returns the next record in the set, or nil if all records of the sequence
+	// have already been returned.
+	Next() Record
+}
```
achille-roussel, line 121: This was one of the main reasons I undertook this work: having the ability to abstract the sequence of records in fetch and produce requests unlocks the next level of opportunities for performance improvements by allowing programs to read and write record batches directly into internal buffers managed by this package.

The APIs we had only supported passing message keys and values as `[]byte` which required the allocation of intermediary, and often large, temporary buffers on the heap, which I've measured to be a performance bottleneck on multiple high throughput programs.

README.md
```diff
@@ -279,22 +308,26 @@ w := kafka.NewWriter(kafka.WriterConfig{
...
+import (
+    "github.com/segmentio/kafka-go/compress/snappy"
+)
+
+...
+
 w := kafka.NewWriter(kafka.WriterConfig{
 	Brokers: []string{"localhost:9092"},
 	Topic:   "topic-A",
 	CompressionCodec: snappy.NewCompressionCodec(),
```
achille-roussel, line 320: Need to fix this example.

offsetfetch.go
```diff
@@ -2,8 +2,119 @@ package kafka
...
+	//
+	// The error contains both the kafka error code, and an error message
+	// returned by the kafka broker. Programs may use the standard errors.Is
+	// function to test the error against kafka error codes.
+	Error error
+}
+
+// OffsetFetch sends a offset fetch request to a kafka broker and returns the
+// response.
+func (c *Client) OffsetFetch(ctx context.Context, req *OffsetFetchRequest) (*OffsetFetchResponse, error) {
```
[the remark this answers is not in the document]
achille-roussel: I actually reused the Kafka API names https://kafka.apache.org/protocol.html#The_Messages_OffsetFetch

metadata_test.go
```diff
@@ -0,0 +1,75 @@
...
+				t.Error("too many partitions were returned in the topic metadata response")
+			}
+		}
+
+		if topicMetadata.Error != nil {
+			t.Error("unexpected error found in the topic metadata response:", topicMetadata.Error)
+		}
+
+		if len(metadata.Topics) > 1 {
+			t.Error("too mnay topics were returned in the metadata response")
```
jnjackins, line 72:
```suggestion
			t.Error("too many topics were returned in the metadata response")
```
maybe worth quickly running a spell checker to catch some more typo?

record.go
```diff
@@ -0,0 +1,42 @@
+package kafka
+
+import (
+	"github.com/segmentio/kafka-go/protocol"
+)
+
+// Header is a key/value pair type representing headers set on records.
+type Header = protocol.Header
```
jnjackins, line 8: > Type aliases are not meant for everyday use. They were introduced to support gradual code repair while moving a type between packages during large-scale refactoring.

This is the advice I always see about type aliases. That's now what we're doing here, so I think it's a code smell. Can you talk a bit about why type aliases are the right choice here?

jnjackins: For example, RecordReader could alternately be defined like this:
```
type RecordReader interface {
    protocol.RecordReader
}
```
and so on.

jnjackins: (or, presumably, the `protocol` types could be used directly everywhere they are needed)

achille-roussel: For the interfaces that's doable, but not for the concrete types, we'd have to redeclare them, which then means they're a different type and would not be compatible when assigning anymore (e.g. we can't assign a `kafka.Header` to `protocol.Header` without explicitly converting the values). I unified everything to use type aliases to avoid having two models.

We had `kafka.Header` before, with the same signature, which is why I used a type alias here. The main concern here is to keep the exported APIs as clean as possible, and avoid breaking compilation as much as possible. This code is very much a transition, we now have two implementations of the protocol in one repository, and more worked to do to simplify it. Type aliases were introduced to help with these large code refactors, so I think the use case is fine here, I expect that we'll evolve the APIs throughout the versions, and these type aliases will go away.

Regarding using the `protocol` package directly, I'd like to discourage programs from doing so, to give us the opportunity to make breaking changes to this package without breaking programs. Maybe we could make it `internal/protocol`, but some programs may be OK with breaking changes if it means they can use the lower level APIs, and the current design of `kafka.RoundTripper` kinds of require `protocol` to be exported.
I didn't want to bring all low-level protocol symbols in the top-level package either (which would have avoided the aliases), because it extends the exported API surface beyond my comfort zone.

If we were starting from scratch, we would probably have gotten a cleaner design, but retaining compatibility with existing code is important here.

protocol/listoffsets/listoffsets.go
```diff
@@ -0,0 +1,230 @@
...
+		})
+	}
+
+	return r, nil
+}
+
+var (
+	_ protocol.BrokerMessage = (*Request)(nil)
+	_ protocol.Mapper        = (*Request)(nil)
+	_ protocol.Reducer       = (*Response)(nil)
```
stevevls, line 229: I'm not finding the definition of `protocol.Reducer` now, but what do you think about `Merge` instead of `Reduce`?  The map/reduce verbiage really throws me b/c it seems out of place in a kafka context.  Still thinking on what to name `Map`.  Also, technically the map operation here is more of a flat map. 😉   `FanOut`?

.circleci/config.yml
```diff
@@ -123,6 +123,37 @@ jobs:
       - run: go get -v -t . ./gzip ./lz4 ./sasl ./snappy
       - run: go test -v -race -cover -timeout 150s $(go list ./... | grep -v examples)
 
+  kafka-231:
```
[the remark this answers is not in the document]
achille-roussel: See my last commit message.

I tried and we have one test failing on kafka 2.4.1 (`TestConn/nettest/PingPong`), which is unrelated to the code changes in this PR.

I would prefer to investigate and tackle this in a separate PR.

docker-compose.yml
```diff
@@ -1,15 +1,16 @@
 version: "3"
 services:
   kafka:
-    image: wurstmeister/kafka:2.11-0.11.0.3
+    image: wurstmeister/kafka:2.12-2.3.1
+    #image: wurstmeister/kafka:0.10.1.1
```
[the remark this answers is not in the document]
achille-roussel: We are, we run CI against 0.10, this is just for local testing.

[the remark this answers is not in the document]
stevevls: Okay.  Let's clean up all the commented out lines, then.

protocol.go
```diff
@@ -102,17 +102,17 @@ func (k apiKey) String() string {
 type apiVersion int16
```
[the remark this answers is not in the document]
achille-roussel: It looks like it's still used in a few places:
```
$ grep apiVersion *.go
conn.go:	apiVersions atomic.Value // apiVersionMap
conn.go:type apiVersionMap map[apiKey]ApiVersion
conn.go:func (v apiVersionMap) negotiate(key apiKey, sortedSupportedVersions ...apiVersion) apiVersion {
conn.go:		if apiVersion(x.MaxVersion) >= s {
conn.go:func (c *Conn) negotiateVersion(key apiKey, sortedSupportedVersions ...apiVersion) (apiVersion, error) {
conn.go:func (c *Conn) loadVersions() (apiVersionMap, error) {
conn.go:	v, _ := c.apiVersions.Load().(apiVersionMap)
conn.go:	v = make(apiVersionMap, len(brokerVersions))
conn.go:	c.apiVersions.Store(v)
conn.go:	var produceVersion apiVersion
conn.go:func (c *Conn) writeRequestHeader(apiKey apiKey, apiVersion apiVersion, correlationID int32, size int32) {
conn.go:	hdr := c.requestHeader(apiKey, apiVersion, correlationID)
conn.go:func (c *Conn) writeRequest(apiKey apiKey, apiVersion apiVersion, correlationID int32, req request) error {
conn.go:	hdr := c.requestHeader(apiKey, apiVersion, correlationID)
conn.go:func (c *Conn) requestHeader(apiKey apiKey, apiVersion apiVersion, correlationID int32) requestHeader {
conn.go:		ApiVersion:    int16(apiVersion),
conn.go:			ApiKey:        int16(apiVersions),
protocol.go:	apiVersions                 apiKey = 18
protocol.go:type apiVersion int16
protocol.go:	apiVersions:                 "ApiVersions",
```

protocol/listoffsets/listoffsets.go
achille-roussel: `Merge` sounds good to me 👍

achille-roussel: How about `Dispatch` instead of `Map`?

achille-roussel: `Split` can be descriptive as well (split as an opposite to merge).

stevevls: `Split` sounds good!

PR #441, opened 2020-05-06
Use SwapInt64 to snapshot counters

Nicofuma:
Using LoadInt64 and then AddInt64 can lead to incoherent results if
snapshot() is called twice at the same time.

We recently encountered this situation in production and it produced a panic somewhere else because of the negative number.


stevevls:
Thanks for the fix!

PR #446, opened 2020-05-13
Fix zstd decoder leak

biasedbit:
**Description**
Using `sync.Pool` to pool encoders/decoders that are not eligible for
deallocation once the pool loses their reference leads to goroutine & mem leaks.

**Kafka Version**
Irrelevant; library issue.

**To Reproduce**
See added test.

**Expected behavior**
There should be no memory or goroutine leaks.

**Additional context**
The docs from `sync.Pool` state:
> Any item stored in the Pool may be removed automatically at any time without
> notification. If the Pool holds the only reference when this happens, the
> item might be deallocated.

`zstd.Decoder`s spin up their own goroutines, which hold a strong ref to owning
decoder. When `decPool` downsizes (usually on GC runs), an arbitrary number of
decoders are evicted from the pool. Since their goroutines are still running
(observed: waiting on `*blockDec.startDecoder()`, at instruction
`for range b.input { ... }`) and thus holding references, these decoders are not
eligible for deallocation — instead becoming unreachable. As new decoders are
created and evicted, the number of running goroutines continuously increases,
eventually causing the program to run out of memory and/or peg the CPU.

I originally noticed this under heavy concurrent load, but managed to create a
trivial, serial test case where the problem is demonstrated.

This change introduces a pretty naive fix. A more robust solution should be
discussed and implemented (or at least a more configurable one). The goal here
was to minimize changes to existing code in order to help focus on the existing
issue and foster discussion.

Likewise, there isn't much value to the test case other than demonstrating the
original problem, and should probably be dropped once a robust solution is put
in place to prevent this leak.

---

I've been benchmarking a system using this patch and both memory and number of goroutines hold steady under heavy, highly concurrent load. We should probably consider a pool that is capable of downsizing (as `sync.Pool` did), or perhaps add a variant of `NewCompressionCodec` that allows specifying the size of the pool (a pool per codec?)

Think of this proposal as a conversation starter. I'll happily help drive implementation :)

### DIFF
--- a/zstd/zstd.go
+++ b/zstd/zstd.go
@@ -17,6 +17,9 @@ const Code = 4
 
 const DefaultCompressionLevel = 3
 
+// DefaultDecoderPoolMaxSize is the default max number of cached decoders for a decoderPool.
+const DefaultDecoderPoolMaxSize = 100
+
 type CompressionCodec struct{ level zstdlib.EncoderLevel }
 
 func NewCompressionCodec() *CompressionCodec {
@@ -39,13 +42,13 @@ func (c *CompressionCodec) NewReader(r io.Reader) io.ReadCloser {
 	if cached := decPool.Get(); cached == nil {
 		p.dec, p.err = zstdlib.NewReader(r)
 	} else {
-		p.dec = cached.(*zstdlib.Decoder)
+		p.dec = cached
 		p.err = p.dec.Reset(r)
 	}
 	return p
 }
 
-var decPool sync.Pool
+var decPool = newDecoderPool(DefaultDecoderPoolMaxSize)
 
 type reader struct {
 	dec *zstdlib.Decoder
--- a/zstd/zstd_test.go
+++ b/zstd/zstd_test.go
@@ -1,0 +1,82 @@
+package zstd
+
+import (
+	"bytes"
+	"crypto/rand"
+	"github.com/segmentio/kafka-go"
+	"io"
+	"runtime"
+	"testing"
+)
+
+func Test2(t *testing.T) {
+	codec := NewCompressionCodec()
+
+	// Run a compress/decompress cycle to create the first decoder and encoder.
+	compressDecompress(codec)
+
+	// Take a snapshot of the number of goroutines after the first comp+dec cycle.
+	// Since all compression and decompression is serial and in this test case
+	// goroutines can only go up if new decoders are created, this number is
+	// expected to remain the same. If it doesn't, there's a leak.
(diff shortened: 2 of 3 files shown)


zstd/zstd_test.go
```diff
@@ -0,0 +1,82 @@
...
+	comp, _ := compress(codec, data)
+	_, _ = decompress(codec, comp)
+}
+
+func randomData(len int) []byte {
+	b := make([]byte, 0, len)
+	_, _ = rand.Read(b)
+	return b
+}
+
```
biasedbit, line 56: Copied `compress` and `decompress` from `../compression_test.go`, rather than refactor to make those reusable since a) I don't think this test has much value other than showing the issue* and b) minimize changes.

If keeping this test turns out to be desirable, I'll happily refactor to remove duplication.

*To observe problem, check out PR branch, undo changes to `zstd/zstd.go` and run this test (or just copy this standalone test and run it against master).

zstd/zstd.go
```diff
@@ -17,6 +17,9 @@ const Code = 4
 
 const DefaultCompressionLevel = 3
 
+// DefaultDecoderPoolMaxSize is the default max number of cached decoders for a decoderPool.
+const DefaultDecoderPoolMaxSize = 100
```
biasedbit, line 21: Arbitrarily chosen. Picked a relatively small number since each `zstd.Decoder` will spin up multiple goroutines.

I've been benchmarking an app with loads >>100 QPS per pod (small EC2 instances) and haven't seen the pool max out yet.

achille-roussel:
I noticed this issue while working on https://github.com/segmentio/kafka-go/pull/438 as well.

The solution I put in place uses finalizers to make sure we close the decoders when they are removed from the pool, but the one you suggest here is just as good 👍 

Thanks for the fix!

biasedbit:
> The solution I put in place uses finalizers to make sure we close the decoders when they are removed from the pool, but the one you suggest here is just as good

Tried an alternative with finalizers and it seems to be a bit more performant (for reference, here's [the benchmark](https://github.com/biasedbit/compression-tests) I used). For an identical (synthetic) workload, I'm seeing roughly extra 100MB/s for `sync.Pool` w/ Finalizers vs `chan` pool.

I think I like the finalizer approach better... Performance aside, the pool gets to grow and shrink as necessary, which isn't the case for the channel based pool I proposed here (which can grow to 100, but never shrinks). I'll open another PR with the finalizers and let you decide if that makes more sense!

PR #450, opened 2020-05-21
Added requests/responses for consumer group admin APIs

stevevls:
This PR adds requests and responses for `ListGroups` and `DescribeGroups`.

Because the consumer groups protocol has some nested binary-encoded structs that are not described as part of the Kafka protocol, there are structs defined so that clients can easily unmarshal the byte slices.

### DIFF
--- a/protocol/decode.go
+++ b/protocol/decode.go
@@ -105,21 +105,6 @@ func (d *decoder) decodeArray(v value, elemType reflect.Type, decodeElem decodeF
 	}
 }
 
-func (d *decoder) decodeMap(v value, typ reflect.Type, decodeVal decodeFunc) {
-	if n := d.readInt32(); n < 0 {
-		v.setArray(array{})
-	} else {
-		m := array{val: reflect.MakeMap(typ)}
-		for i := 0; i < int(n) && d.remain > 0; i++ {
-			k := reflect.ValueOf(d.readString())
-			v := reflect.New(typ.Elem()).Elem()
-			decodeVal(d, value{v})
-			m.val.SetMapIndex(k, v)
-		}
-		v.setArray(m)
-	}
-}
-
 func (d *decoder) discardAll() {
 	d.discard(d.remain)
 }
@@ -321,8 +306,6 @@ func decodeFuncOf(typ reflect.Type, version int16, tag structTag) decodeFunc {
 			return bytesDecodeFuncOf(tag)
 		}
 		return arrayDecodeFuncOf(typ, version, tag)
-	case reflect.Map:
-		return mapDecodeFuncOf(typ, version, tag)
 	default:
 		panic("unsupported type: " + typ.String())
 	}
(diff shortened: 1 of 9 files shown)


protocol/decode.go
```diff
@@ -105,6 +105,21 @@ func (d *decoder) decodeArray(v value, elemType reflect.Type, decodeElem decodeF
 	}
 }
 
+func (d *decoder) decodeMap(v value, typ reflect.Type, decodeVal decodeFunc) {
+	if n := d.readInt32(); n < 0 {
+		v.setArray(array{})
+	} else {
+		m := array{val: reflect.MakeMap(typ)}
```
achille-roussel, line 112: Is this going to break if the `unsafe` build tag is specified (see `protocol/reflect_unsafe.go`)?

protocol/describegroups/describegroups.go
```diff
@@ -0,0 +1,73 @@
...
+// ConsumerGroupAssignment is the structured representation of
+// GroupMember.MemberAssignment when the group is using Kafka's consumer group
+// protocol.  This is the usual case and will be indicated by the Group's
+// ProtocolType being set to 'consumer'.
+//
+// After checking the protocol type, this struct can be unmarshalled using
+// kafka.Unmarshal.
+type ConsumerGroupAssignment struct {
+	Version  int16
+	Topics   map[string][]int32
```
achille-roussel, line 71: I have mixed feelings about supporting maps, since the data type doesn't really exists in the kafka protocol.

What would you think about changing this to an array?
```go
type ConsumerGroupAssignment struct {
    Version  int16
    Topics   []ConsumerGroupTopic
    UserData []byte
}
```
```go
type ConsumerGroupTopic struct {
    Topic      string
    Partitions []int32
}
```

stevevls: Fancy...that hadn't occurred to me.  I like it!

protocol/describegroups/describegroups.go
```diff
@@ -0,0 +1,105 @@
...
+	if len(r.GroupIDs) != 1 {
+		return ""
+	}
+	return r.GroupIDs[0]
+}
+
+func (r *Request) Map(cluster protocol.Cluster) ([]protocol.Message, protocol.Reducer, error) {
+	requests := make([]protocol.Message, len(r.GroupIDs))
+	for i, group := range r.GroupIDs {
+		requests[i] = &Request{GroupIDs: []string{group}}
```
achille-roussel, line 36: To reduce heap allocations you could do two things here:

* Allocate all requests in a single slice:
```go
requests := make([]Request, len(r.GroupIDs))
messages := make([]protocol.Message, len(r.GroupIDs))

for i := range r.GroupIDs {
    requests[i] = ...
    messages[i] = &requests[i]
}

return messages, new(Response), nil
```

* Takes slices in the input `GroupIDs` instead of constructing new slices with `[]string{group}`:
```go
for i := range r.GroupIDs {
    requests[i] = Request{GroupIDs: r.GroupIDs[i : i+1 : i+1]} // 3-index form to cap the capacity
    ...
}
```

achille-roussel:
Closing, these APIs were added via a different PR.

Issue #451, opened 2020-05-21
Producer stuck on TCP connection error


abuchanan-nr:
When a producer encounters a TCP connection error, it becomes completely blocked and doesn't recover with a new connection.

I suspect I introduced a bug in #382. I haven't verified that. If so, it would affect v0.3.5 and v0.3.6.

**Kafka Version**
Tested against kafka v0.10.2.1

**To Reproduce**
```go
package main

import (
	"context"
	"log"
	"os"
	"time"

	"github.com/segmentio/kafka-go"
)

func main() {
	logger := log.New(os.Stderr, "kafkago", 0)
	writer := kafka.NewWriter(kafka.WriterConfig{
		Brokers:     []string{"127.0.0.1:9093"},
		Topic:       "test_topic",
		Logger:      logger,
		ErrorLogger: logger,
	})
	ctx := context.Background()

	for range time.Tick(time.Second) {
		log.Println("tick")

		err := writer.WriteMessages(ctx, kafka.Message{
			Key:   []byte("Key"),
			Value: []byte("Val"),
		})
		if err != nil {
			log.Println(err)
		}
	}
}
```

Then run `docker stop kafka` (assuming you're running a test kafka broker in a container).

The producer will correct retry once, but get stuck on the second attempt. I suspect that's because nothing is reading from the result channel being written to at this line:
https://github.com/segmentio/kafka-go/blob/master/writer.go#L748

Issue #426 could be related.

lukemassa:
I believe I am seeing the same issue, recording some info here in case it's useful:

I create NewWriter() and try to write to it, however my brokers are unreachable (for unrelated reasons). I then see these log lines:

```
time="2020-07-08T22:25:24Z" level=error msg="error dialing kafka brokers for topic testtopic (partition 0): dial tcp 10.41.10.178:29092: connect: connection refused"
time="2020-07-08T22:25:24Z" level=error msg="error dialing kafka brokers for topic testtopic (partition 0): dial tcp 10.41.10.178:29092: connect: connection refused"
time="2020-07-08T22:25:25Z" level=error msg="error dialing kafka brokers for topic testtopic (partition 0): dial tcp 10.41.10.178:29092: connect: connection refused"
```

Followed by the code hanging. I killed go and grabbed the stack trace that it was hung up on:

```
goroutine 186 [chan send, 4 minutes]:
[...]

PR #452, opened 2020-05-21
return write error only when all retries have failed

abuchanan-nr:
This is a potential fix for #451. It passes my quick-and-dirty local test.

This changes the code to only return an error (via the message "res" channel) when all write retries have failed.

This also moves the stat counters associated with a successful write (messages, bytes, and batchSize), so that they are not inflated by each call to write (i.e. each retry/attempt).

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -719,20 +719,34 @@ func (w *writer) writeWithRetries(conn *Conn, batch []Message, resch [](chan<- e
 		w.stats.retries.observe(1)
 		time.Sleep(backoff(attempt+1, 100*time.Millisecond, 1*time.Second))
 	}
+
+	if err != nil {
+		for i, res := range resch {
+			res <- &writerError{msg: batch[i], err: err}
+		}
+	} else {
+		for _, res := range resch {
+			res <- nil
+		}
+		for _, m := range batch {
+			w.stats.messages.observe(1)
+			w.stats.bytes.observe(int64(len(m.Key) + len(m.Value)))
+		}
+		w.stats.batchSize.observe(int64(len(batch)))
+	}
+
 	return conn, err
 }
 
 func (w *writer) write(conn *Conn, batch []Message, resch [](chan<- error)) (ret *Conn, err error) {
 	w.stats.writes.observe(1)
+
 	if conn == nil {
 		if conn, err = w.dial(); err != nil {
 			w.stats.errors.observe(1)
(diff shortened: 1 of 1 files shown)


VibhorGupta1991:
This change needs to be pushed really quick. I have been struggling to find the cause but finally came across this [#451 ]. 
This is affecting only **0.3.6 version** as of now. @abuchanan-nr your change [#382 ] was merged on 22nd Feb in this version.

writer.go
```diff
@@ -719,20 +719,34 @@ func (w *writer) writeWithRetries(conn *Conn, batch []Message, resch [](chan<- e
...
+	if err != nil {
+		for i, res := range resch {
+			res <- &writerError{msg: batch[i], err: err}
+		}
+	} else {
+		for _, res := range resch {
+			res <- nil
+		}
+		for _, m := range batch {
+			w.stats.messages.observe(1)
```
achille-roussel, line 732: Can we take this out of the loop? It seems like it could be `w.stats.messages.observe(int64(len(batch)))`

stevevls:
Thanks for the fix @abuchanan-nr !  I traced through and agree with your assessment that there may not be a channel receiver on a second attempt.

I would prefer that we leave the meaning of the stats as-is, though.  It's possible that someone out there is depending on them, so I'd prefer we don't change that and create a surprise for someone.

jnjackins:
Given the nature of the breakage I believe we should revert #382 in the meantime -- see #462.  Since there are low expectations for the reliability of async writes, we should give preference to sync writes which are meant to be reliable, and they are currently broken. Also, we should be sure to add more tests to cover these changes -- currently there is no test coverage for the partitionWriter abstraction, which is mocked over in tests (specifically the only test we have for retries).

If we do want to move forward with doing retries in the `partitionWriter`, I'd suggest we add a new interface which abstracts the `write` method, which can have a simple (non-retrying) implementation, and then we can add a `retryWriter` implementation which wraps the simple one. `partitionWriter`s can be configured to use the `retryWriter` implementation normally, and in tests the simple writer can be faked and we can still exercise the retry functionality.

Note that this would mean there is no more use for a partitionWriter abstraction, which currently only exists so that it can be faked in tests.

achille-roussel:
Considering we reverted the change that introduced the original regression, I believe we can close this PR. Feel free to reopen if it needs further discussion 👍

Issue #459, opened 2020-06-12
Closing a Writer is "deadlocked" on `func (w *writer) close() {`


JordyMoos:
**Describe the bug**
If I create a basic Writer (as in the readme) and write messages to it and make sure all works fine (kafka up and running). Now if I kill kafka then I will get a `EOF` error when writing to kafka. So far as expected.

But now if I want to close that writer. Then it will hang in a dead lock

**Kafka Version**
wurstmeister/kafka:2.12-2.4.0

**To Reproduce**
Steps to reproduce the behavior.  Bonus points for a code sample.

**Expected behavior**
I expect that Kafka will be successfully closed within a minute.

**Additional context**
I think the problem lies here:
This will be called and adds 1 to the waitgroup
https://github.com/segmentio/kafka-go/blob/v0.3.6/writer.go#L502
```go
func (w *Writer) close(writer partitionWriter) {
	w.join.Add(1)
	go func() {
		writer.close()
		w.join.Done()
	}()
}
```

Which in turn will execute this piece of code:
https://github.com/segmentio/kafka-go/blob/v0.3.6/writer.go#L568
```go
func (w *writer) close() {
	close(w.msgs)
	w.join.Wait()
}
```

Which will then Wait for the waitgroup. Although that will never succeed because of the addition on line 502.

What will happen is something like this
```go
func (w *Writer) close(writer partitionWriter) {
	w.join.Add(1)
	go func() {
		close(w.msgs)
                // This will wait forever because the Done is the line below
	        w.join.Wait()
		w.join.Done()
	}()
}
```

Removing the wait on line 570 will solve the neverending wait.

stevevls:
Thanks for submitting this bug report and for including code samples!  Could you also share the `WriterConfig` that you are using minus the hostnames?

One clarification I'd like to make is that the there are two types in play.  There's the exported `Writer` and the unexported `writer`, each of which have a `close` function and a wait group named `join`.  In the first code block above, the `Writer` (capital W) calls `close` on the `writer` (lower w).  With that in mind, the inlined code would look like this instead:

```
func (w *Writer) close(writer partitionWriter) {
	w.join.Add(1)
	go func() {
		close(writer.msgs)
	        writer.join.Wait()
		w.join.Done()
	}()
}
```

Thanks!

JordyMoos:
Thanks for answering the report @stevevls 

You are indeed right, I overlooked that piece and that looks fine. Thanks for clarifying, that makes way more sense now!

[...]

Issue #445, opened 2020-05-09
No error message or summary stats of error 


nporsche:
w := kafka.NewWriter(kafka.WriterConfig{
		Brokers:       strings.Split(kafkaServers, ";"),
		Topic:         kafkaTopic,
		Balancer:      &kafka.LeastBytes{},
		Async:         true,
		QueueCapacity: defaultKafkaQueueCapacity,
		BatchSize:     defaultKafkaBatchSize,
		Logger:        &kafkaInfoLogger{},
		ErrorLogger:   &kafkaErrLogger{},
	})
	defer func() {
		w.Close()
		wg.Done()
		glog.V(1).Info("output goroutine exits")
	}()

	counter := 0
	exit := false
	for !exit {
		select {
		case c := <-ch:
			w.WriteMessages(ctx,
				kafka.Message{
					Key:   []byte(fmt.Sprintf("%s_%d", host, time.Now().Unix())),
					Value: c,
				},
			)
			counter++

			if glog.V(1) && counter%10 == 0 {
				prettyJSON, _ := json.MarshalIndent(w.Stats(), "", "\t")
				glog.V(1).Infof("Kafka Stats Summary:\n%s", string(prettyJSON))
			}
			glog.V(2).Infof("success WriteMessage to kafka client")
		case <-ctx.Done():
			exit = true
		}
	}

If given an invalid broker address, I cannot find any error message or summary of error. The root cause may be is:
func (w *Writer) partitions() (partitions []int, err error) {
	for _, broker := range shuffledStrings(w.config.Brokers) {
		var conn *Conn
		var plist []Partition

		if conn, err = w.config.Dialer.Dial("tcp", broker); err != nil {
			continue
		}

		conn.SetReadDeadline(time.Now().Add(w.config.ReadTimeout))
		plist, err = conn.ReadPartitions(w.config.Topic)
		conn.Close()

		if err == nil {
			partitions = make([]int, len(plist))
			for i, p := range plist {
				partitions[i] = p.ID
			}
			break
		}
	}

	sort.Ints(partitions)
	return
}

no error report out if dial is error

achille-roussel:
Hello @nporsche, and thank you for reporting and identifying the source of the problem.

We always welcome contributions, would you be available to submit a fix for this?

PR #420, opened 2020-03-27
accept asynchronous error handler

dmarkhas:
Addresses #405 by allowing setting a callback to handle errors in an Async producer

Issue #419, opened 2020-03-27
WriteMessages no longer waits for retries on synchoronous writes before returning an error


WillAbides:
**Describe the bug**

I believe that https://github.com/segmentio/kafka-go/pull/382 introduces a breaking change to WriteMessages. Updating a project from `v0.3.5` to `efce7b6e0bf4da4d8a76aaf26fac880a023396d1` caused our integration tests to fail. The tests use `WriteMessages` to publish messages to a new topic. Before the updating, `WriteMessages` would succeed after retries. After the update, it immediately returns `LeaderNotAvailable`. 

One workaround is putting `WriteMessages` in a retry loop.

**Kafka Version**
What version(s) of Kafka are you testing against?

1.1.0.0

**To Reproduce**

I don't have time to produce a code sample now. I can come up with one later if needed.

I think the only thing you should need to do is write to a new topic. 

The general steps are:

1. Start a kafka cluster with auto-create topics enabled.
2. Use WriteMessages to write to a new topic.
3. See an immediate LeaderNotAvailable response.

**Expected behavior**

`WriteMessages` should wait until retries are exhausted to return an error.

**Additional context**

I'm not sure whether the writes get retried after returning the error.

achille-roussel:
Hello @WillAbides, I believe the issue you described was addressed in a 0.3.6 with this fix https://github.com/segmentio/kafka-go/pull/462

Let me know if you have still been running into it with later kafka-go versions.

achille-roussel:
Hello @zydovech 

I'm not actively contributing to this project anymore but I'm sure someone from the maintainer team at Twilio can help take a look if you share a reproducible example.

PR #461, opened 2020-06-14
0.4: kafka.Writer

achille-roussel:
This PR modifies the implementation of the `kafka.Writer` type to adopt the 0.4 APIs and leverage connection pooling of `kafka.Transport` in the writers.

I am also introducing a couple new APIs to address known flaws of the existing API:
* Enhanced errors returned by `WriteMessages` so programs can distinguish which messages have succeeded and failed.
* An asynchronous completion API to support reporting the status of writes when the writers are configured with `Async: true`.

A major change in the implementation is the removal of all queuing, and with it the need for blocking and synchronizing error reporting. Batches are now aggregated in a shared state instead of using message passing to aggregate batches in each partition writer.

I've also done documentation work, which hopefully is going to help address some of the open issues as well.

Here are a couple of issues that this PR will likely address:
https://github.com/segmentio/kafka-go/issues/53
https://github.com/segmentio/kafka-go/issues/131
https://github.com/segmentio/kafka-go/issues/298
https://github.com/segmentio/kafka-go/issues/342
https://github.com/segmentio/kafka-go/issues/350
https://github.com/segmentio/kafka-go/issues/356
https://github.com/segmentio/kafka-go/issues/364
https://github.com/segmentio/kafka-go/issues/391
https://github.com/segmentio/kafka-go/issues/419
https://github.com/segmentio/kafka-go/issues/445
https://github.com/segmentio/kafka-go/issues/451
https://github.com/segmentio/kafka-go/issues/459
https://github.com/segmentio/kafka-go/pull/176
https://github.com/segmentio/kafka-go/pull/420
https://github.com/segmentio/kafka-go/pull/452

### DIFF
--- a/produce.go
+++ b/produce.go
@@ -12,6 +12,27 @@ import (
 	produceAPI "github.com/segmentio/kafka-go/protocol/produce"
 )
 
+type RequiredAcks int
+
+const (
+	RequireNone RequiredAcks = 0
+	RequireOne  RequiredAcks = 1
+	RequireAll  RequiredAcks = -1
+)
+
+func (acks RequiredAcks) String() string {
+	switch acks {
+	case RequireNone:
+		return "none"
+	case RequireOne:
+		return "one"
+	case RequireAll:
+		return "all"
+	default:
+		return "unknown"
+	}
+}
+
 // ProduceRequest represents a request sent to a kafka broker to produce records
 // to a topic partition.
 type ProduceRequest struct {
@@ -25,7 +46,7 @@ type ProduceRequest struct {
 	Partition int
 
 	// The level of required acknowledgements to ask the kafka broker for.
-	RequiredAcks int
+	RequiredAcks RequiredAcks
 
 	// The message format version used when encoding the records.
 	//
@@ -88,6 +109,9 @@ type ProduceResponse struct {
 //
 // If the request contained no records, an error wrapping protocol.ErrNoRecord
 // is returned.
+//
+// When the request is configured with RequiredAcks=none, both the response and
+// the error will be nil on success.
 func (c *Client) Produce(ctx context.Context, req *ProduceRequest) (*ProduceResponse, error) {
 	attributes := protocol.Attributes(req.Compression) & 0x7
 
@@ -113,6 +137,10 @@ func (c *Client) Produce(ctx context.Context, req *ProduceRequest) (*ProduceResp
 		return new(ProduceResponse), nil
 	default:
 		return nil, fmt.Errorf("kafka.(*Client).Produce: %w", err)
+	}
+
+	if req.RequiredAcks == RequireNone {
+		return nil, nil
 	}
 
 	res := m.(*produceAPI.Response)
--- a/writer.go
+++ b/writer.go
@@ -208,7 +208,7 @@ type Writer struct {
 // WriterConfig is a configuration type used to create new instances of Writer.
 //
 // DEPRECATED: writer values should be configured directly by assigning their
-// exported fields. This type is kept for backward compatibilty, and will be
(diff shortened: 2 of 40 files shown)


achille-roussel:
I was being blocked by flaky tests so I added fixes for those, also fixed the CI config so it would run all tests, and made it run against all kafka versions from 0.10 to 2.3.

writer.go
```diff
@@ -1,37 +1,215 @@
...
+	// for extra synchronization to access this field.
+	roundRobin RoundRobin
+
+	// non-nil when a transport was created by NewWriter, remove in 1.0.
+	transport *Transport
 }
 
 // WriterConfig is a configuration type used to create new instances of Writer.
+//
+// DEPRECATED: writer values should be configured directly by assigning their
```
jnjackins, line 210: I'm questioning whether doing a two-phase deprecation is worthwhile, it seems OK to remove API while in v0. seems like we're delaying ripping off the band-aid :)

jnjackins: but I'm fully open to being wrong about that of course

writer_test.go
```diff
@@ -144,38 +142,16 @@ func TestValidateWriter(t *testing.T) {
...
-
+	topic := makeTopic()
 	createTopic(t, topic, 1)
+	defer deleteTopic(t, topic)
+
 	w := newTestWriter(WriterConfig{
+		Brokers:     []string{"localhost:9999"}, // nothing is listening here
 		Topic:       topic,
-		MaxAttempts: 1,
+		MaxAttempts: 3,
```
jnjackins, line 153: It would be nice to have a bit more thorough testing around how many attempts were actually made, and that writes actually succeed on subsequent attempts if the temporary issue resolves. In the old version, I had started going down the route of putting the `write` method (now `produce`) in an interface so that test code could substitute it with a dumb implementation which counts attempt numbers. Do you have any thoughts on that?

jnjackins: also, I think this test will fail if I happen to have a kafka server running on port 9999 😄

produce.go
```diff
@@ -25,7 +46,7 @@ type ProduceRequest struct {
 	Partition int
 
 	// The level of required acknowledgements to ask the kafka broker for.
-	RequiredAcks int
+	RequiredAcks RequiredAcks
```
jnjackins, line 49: Interesting. I didn't even know these were code for none/one/all, I thought I was specifying exactly how many acks I wanted.

writer.go
achille-roussel: I thought about it, but I feel like we're doing the right thing by optimizing for backward compatibility. I've had too many pains with library's making breaking changes, it's true that we are below 1.0 but it doesn't mean we shouldn't be mindful of the extra work we push down to users when we make breaking changes. There are a few breaking changes in 0.4 already (all trivial to address), I'd like to keep the count as low as possible and do the API cleanup when we're ready to release 1.0.

writer.go
```diff
@@ -289,523 +541,531 @@ func NewWriter(config WriterConfig) *Writer {
...
+			// are that the program will check if WriteMessages returned a
+			// MessageTooLargeError, discard the message that was exceeding
+			// the maximum size, and try again.
+			return messageTooLarge(msgs, i)
 		}
+	}
 
-		wm := writerMessage{msg: msg, res: res}
+	numPartitions, err := w.partitions(ctx)
+	if err != nil {
```
sagarkrkv, line 588: If I'm reading this right, we would make an expensive call to the broker to fetch the number of partitions for every WriteMessage call.

sagarkrkv: I see that caching is done in the new Transport, maybe a comment would help?

PR #462, opened 2020-06-15
Revert #382

jnjackins:
CC @abuchanan-nr 

There were a few issues with #382 that we should address. In the meantime, this PR:
- reverts the change, and
- updates `testWriterMaxAttemptsErr` to ensure it's testing the behavior it thinks it's testing.

To summarize the issues we know of with #382:
- `WriteMessages` returns early with initial write errors rather than waiting for retries, as we write directly to the response channels before retrying.
- The response channel can become full after a batch fails, and further retries are blocked on writes to the full channel.
- There is a subtle behavior change that was perhaps not considered well enough: rather than `WriteMessages` blocking on sleeps during retry, the background `partitionWriter` threads are now doing the sleeping. This may mean that other concurrent writes to the same partition, are blocked, while previously they would not have been.
- Moving retries to the partition writer also means that our current hook for faking writes in tests, the `partitionWriter` interface, is not useful -- using it means we do not cover retry logic in our tests.

We can most likely overcome these problems, but we currently have systems which need a fix and this deserves more time than I can spend on it right now. Additionally, there is a new implementation of `Writer` coming in 0.4, so it may not be worthwhile to spend too much time on this.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -308,50 +308,83 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
 		res = make(chan error, len(msgs))
 	}
 	t0 := time.Now()
-	defer w.stats.writeTime.observeDuration(time.Since(t0))
 
-	w.mutex.RLock()
-	closed := w.closed
-	w.mutex.RUnlock()
+	for attempt := 0; attempt < w.config.MaxAttempts; attempt++ {
+		w.mutex.RLock()
 
-	if closed {
-		return io.ErrClosedPipe
-	}
+		if w.closed {
+			w.mutex.RUnlock()
+			return io.ErrClosedPipe
+		}
 
-	for i, msg := range msgs {
-
-		if int(msg.size()) > w.config.BatchBytes {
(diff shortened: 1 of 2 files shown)


writer.go
```diff
@@ -308,50 +308,83 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
...
+			}:
+			case <-ctx.Done():
+				w.mutex.RUnlock()
+				return ctx.Err()
 			}
-			return err
 		}
 
-		wm := writerMessage{msg: msg, res: res}
+		w.mutex.RUnlock()
```
jnjackins, line 340: Note: I don't understand why we keep this mutex for so long in this version of the code. We only seem to need to lock `w.closed`

writer_test.go
```diff
@@ -140,14 +140,17 @@ func TestValidateWriter(t *testing.T) {
 	}
 }
 
-type fakeWriter struct{}
+type fakeWriter struct {
```
jnjackins, line 143: This file contains all the changes not part of the revert.

Issue #477, opened 2020-08-10
nil pointer dereference in tls check in v0.4.1


dangogh:
**Describe the bug**
Trying to use a minimal tls.Config,  I get this crash:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb74d33]

goroutine 38 [running]:
github.com/segmentio/kafka-go.(*connGroup).connect(0xc0001f3480, 0xed4dc0, 0xc000077620, 0x0, 0x0, 0x0)
	/go/pkg/mod/github.com/segmentio/kafka-go@v0.4.1/transport.go:1016 +0xc63
github.com/segmentio/kafka-go.(*connGroup).grabConnOrConnect.func1(0xc0001f3480, 0xed4d40, 0xc0001f3400, 0xc000101e00, 0xc000101da0)
	/go/pkg/mod/github.com/segmentio/kafka-go@v0.4.1/transport.go:881 +0x6e
created by github.com/segmentio/kafka-go.(*connGroup).grabConnOrConnect
	/go/pkg/mod/github.com/segmentio/kafka-go@v0.4.1/transport.go:880 +0x101
```

**Kafka Version**
2.2.1 (MSK on AWS)

**To Reproduce**
I wrote this test to show the crash:
```
package kafka

import (
	"context"
	"crypto/tls"
	"testing"
)

func TestMakeItCrash(t *testing.T) {
	cg := connGroup{
		pool: &connPool{
			tls: &tls.Config{},
		},
	}

	cg.connect(context.Background())
}
```

**Expected behavior**
Successful connection with kafka when a minimal `tls.Config` is used.

**Additional context**
It appears the `netAddr` variable is never being set in this function.  It's declared as `net.Addr` which is an interface, but never set otherwise, so it would be nil.

achille-roussel:
Thanks for the bug report! I'm looking into this 👍

achille-roussel:
Note that v0.4.x is not yet officially stable, we're still ironing out the details. If you need a stable version you can pin your dependency at v0.3.7

dangogh:
thanks!   I'll lock on v0.3.7 for a while, then..

achille-roussel:
[...]

PR #478, opened 2020-08-11
0.4 fix tls address panic

achille-roussel:
Fixes https://github.com/segmentio/kafka-go/issues/477

### DIFF
--- a/transport_test.go
+++ b/transport_test.go
@@ -28,5 +28,7 @@ func TestIsse477(t *testing.T) {
 		// An error is expected here because we are not actually establishing
 		// a TLS connection to a kafka broker.
 		t.Log(err)
+	} else {
+		t.Error("no error was reported when attempting to establish a TLS connection to a non-TLS endpoint")
 	}
 }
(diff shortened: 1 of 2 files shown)


transport_test.go
```diff
@@ -0,0 +1,32 @@
+package kafka
+
+import (
+	"context"
+	"crypto/tls"
+	"net"
+	"testing"
+)
+
+func TestIsse477(t *testing.T) {
```
stevevls, line 10:
```suggestion
func TestIssue477(t *testing.T) {
```

transport_test.go
```diff
@@ -0,0 +1,32 @@
...
+	cg := connGroup{
+		addr: l.Addr(),
+		pool: &connPool{
+			dial: defaultDialer.DialContext,
+			tls:  &tls.Config{},
+		},
+	}
+
+	if _, err := cg.connect(context.Background()); err != nil {
+		// An error is expected here because we are not actually establishing
```
stevevls, line 28: If the error is expected, would it be a stronger verification is we fail the test if it's `nil`?

achille-roussel: Yep, thanks for keeping me honest :)

PR #479, opened 2020-08-12
0.4: fix short writes

achille-roussel:
This PR addresses an issue where producing to a kafka 0.10 broker using kafka-go 0.4 would result in silently dropping parts of the batch.

The issues was an inverted logic in a error check, which caused compressed message sets to be truncated after writing the first buffer page. This code path was specialized for the message v1 format, which explains why it didn't hit on kafka brokers in order version (which use the version 2 of the message format).

I added a test which allowed me to reproduce the issue and verify that the changed corrected the behavior.

### DIFF
--- a/protocol/buffer.go
+++ b/protocol/buffer.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"sort"
 	"sync"
 	"sync/atomic"
 )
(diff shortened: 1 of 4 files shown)


protocol/buffer.go
```diff
@@ -461,9 +460,10 @@ func (pages contiguousPages) slice(begin, end int64) contiguousPages {
 }
 
 func (pages contiguousPages) indexOf(offset int64) int {
-	return sort.Search(len(pages), func(i int) bool {
-		return offset < (pages[i].offset + pages[i].Size())
-	})
+	if len(pages) == 0 {
+		return 0
+	}
+	return int((offset - pages[0].offset) / pageSize)
```
achille-roussel, line 466: I simplified this function, the search was overkill since the pages are contiguous, we can simply compute the index from the offset.

PR #481, opened 2020-08-12
make the dial function configurable on the kafka.Dialer type

achille-roussel:
This PR adds the ability to configure the function used by `kafka.Dialer` to establish network connections. This customization is useful to add monitoring on the network connections opened by the various components (Reader, Writer, etc...),  as they can all be configured to use a `kafka.Dialer`.

I chose to name it `DialFunc` because the `kafka.Dialer` type already had methods called `Dial` and `DialContext`, but I'm open to better naming suggestions.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -23,7 +23,7 @@ type Dialer struct {
 	//
 	// When DialFunc is set, LocalAddr, DualStack, FallbackDelay, and KeepAlive
 	// are ignored.
-	DialFunc func(context.Context, string, string) (net.Conn, error)
+	DialFunc func(ctx context.Context, network string, address string) (net.Conn, error)
 
 	// Timeout is the maximum amount of time a dial will wait for a connect to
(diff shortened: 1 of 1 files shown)


dialer.go
```diff
@@ -18,6 +18,13 @@ type Dialer struct {
 	// Unique identifier for client connections established by this Dialer.
 	ClientID string
 
+	// Optionally specifies the function that the dialer uses to establish
+	// network connections. If nil, net.(*Dialer).DialContext is used instead.
+	//
+	// When DialFunc is set, LocalAddr, DualStack, FallbackDelay, and KeepAlive
+	// are ignored.
+	DialFunc func(context.Context, string, string) (net.Conn, error)
```
pelletier, line 26:
```suggestion
	DialFunc func(context context.Context, network string, address string) (net.Conn, error)
```
:point_up: suggest adding the names of arguments in case you don't have the `DialContext` on top of your head.

PR #482, opened 2020-08-13
ReadBatchWith bug fix and new MaxWaitTime option

stevevls:
This fixes a bug where ReadBatchWith could incorrectly return `io.EOF`
in certain `errShortRead` scenarios.  Since `io.EOF` means the batch
was handled correctly, this could cause consumers to continue to use
a connection that is in a bad state.

This also adds a MaxWaitTime option that can be used to improve the
way that the conn deadline and the fetch request interact.  Instead
of trying to calculate a correct timeout and change the max wait
time, this allows the caller to explicitly configure both the wait
and the deadline independently of one another.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -121,6 +121,13 @@ type ReadBatchConfig struct {
 	// ReadUncommitted makes all records visible. With ReadCommitted only
 	// non-transactional and committed records are visible.
 	IsolationLevel IsolationLevel
+
+	// The amount of time for the broker while waiting to hit the min/max byte
+	// targets.  This setting is independent of any network-level timeouts or
+	// deadlines.
+	//
+	// For backward compatibility, the default value here will todo
+	MaxWaitTime time.Duration
 }
 
 type IsolationLevel int8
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -790,7 +797,20 @@ func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
...
+			timeout = cfg.MaxWaitTime
+		} else {
+			// default case: use the original logic to adjust the conn's
+			// deadline.T
+			deadline = adjustDeadlineForRTT(deadline, now, defaultRTT)
+			timeout = deadlineToTimeout(deadline, now)
+		}
+		// save this variable outside of the closure for later use in detecting
+		// truncated messages.
+		adjustedDeadline = deadline
```
stevevls, line 813: Prior to this change, this variable was never assigned so would always be the zero value.

conn.go
```diff
@@ -121,6 +121,13 @@ type ReadBatchConfig struct {
 	// ReadUncommitted makes all records visible. With ReadCommitted only
 	// non-transactional and committed records are visible.
 	IsolationLevel IsolationLevel
+
+	// The amount of time for the broker while waiting to hit the min/max byte
+	// targets.  This setting is independent of any network-level timeouts or
+	// deadlines.
+	//
+	// For backward compatibility, the default value here will todo
+	MaxWaitTime time.Duration
```
achille-roussel, line 130: Since we have this field named `MaxWait` in other places, how about sticking to that name to keep the APIs consistent?

stevevls: SGTM.  Also looks like I didn't finish the docs. 🤦

PR #495, opened 2020-08-29
Added code and test for GetControllerConn() in conn.go. This correctl…

Zell45:
…y provides the controller connection for the kafka cluster which is required to avoid error 41 when attempting to use segmentio to create topics on a connection to a node in the cluster which is not the controller on a server which has replication factor > 1.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -295,6 +295,44 @@ func (c *Conn) Brokers() ([]Broker, error) {
 	return brokers, err
 }
 
+// GetControllerConn returns the connection object for the Controller of the kafka cluster 
+// required to avoid the error encountered when attempting to create topics on a conn-
+// ection to a non-controller connection to a kafka server with replication factor > 1;
+// "error = [41] Not Controller: this is not the correct controller for this cluster"
+func (c *Conn) GetControllerConn() (*Conn, error) {
+	// Ranges over the list of brokers returned from the current connection,
+	// then returns the connection object for the controller of the cluster.
+	var controllerConn *Conn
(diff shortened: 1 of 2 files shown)


Zell45:
I hope this addition can make it easier for people who might encounter kafka error 41 in kafka server environments which use replication to identify, resolve and make use of this extra functionality.

Pryz:
Hi @Zell45, thanks for your PR.

This looks a bit redundant with the existing https://pkg.go.dev/github.com/segmentio/kafka-go?tab=doc#Conn.Controller where the user would just have to use something like : 

```
conn, _ := kafka.Dial("tcp", "localhost:9092")
controller, _ := conn.Controller()
conn, _ = kafka.Dial("tcp", net.JoinHostPort(controller.Host, strconv.Itoa(controller.Port)))
// ... create topics, etc etc
```

Did you take a look at the existing Controller method ?

Thanks

Zell45:
Hi @Pryz, thanks for your review.

I posted this function to provide the convenience of guaranteeing a defined connection object accesses the controller of the cluster (throwing errors where appropriate along the way). It incorporates the existing Conn.Controller() method but rather than returning the broker it returns the direct connection object to the controller.

In my working group we actually found it to be a bit of a goose chase to create the dialler object and make the right connection after dealing with the surprise error 41 on our remote replicated cluster- mainly before this we were using the higher level abstractions of Reader and Writer objects, and the problem was not apparent on the locally hosted kafka servers. 

Above you point out you can use the kafka.Dial method which might be more lightweight than the way I have created a Dialer object. Happy to hear your thoughts on it.

Thanks again!

Issue #487, opened 2020-08-18
Release tag 0.4.2: response from kafka is on unknown type while using kafka.Writer


riteshmodi:
**Describe the bug**
When publishing the message using `kafka.Writer` from release tag v0.4.2 then I am getting error like this: `panic: BUG: promise resolved with impossible value of type <nil>` and message is getting written to kafka successfully.

**Kafka Version**
kafka_2.11-2.0.0

**To Reproduce**
Set up local kafka broker and send the message using following code:
```
import (
	"net/http"

	kafka "github.com/segmentio/kafka-go"
)

func GetKafkaProducer() *kafka.Writer {
	return &kafka.Writer{
		Addr:     kafka.TCP("localhost:9092"),
		Topic:    "topicA",
		Balancer: &kafka.LeastBytes{},
	}
}

func PublishMessage(r *http.Request, w *kafka.Writer, msg kafka.Message) {
	w.WriteMessages(r.Context(), msg)
}
```
**Test code**
```
import (
	"context"
	"net/http"
	"testing"
	"time"

	kafka "github.com/segmentio/kafka-go"
	"github.com/stretchr/testify/assert"
)

func TestWriteToKafka(t *testing.T) {

	kw := GetKafkaProducer()
	defer kw.Close()

	msg := kafka.Message{
		Key:   []byte("KeyA"),
		Value: []byte("It seems that issue is in latest release. Check again!!")}

	ctx, cancel := context.WithTimeout(context.Background(), 2000*time.Millisecond)
	defer cancel()
	req, _ := http.NewRequestWithContext(ctx, "POST", "http://google.com", nil)

	PublishMessage(req, kw, msg)
	assert := assert.New(t)
	assert.Fail("")

}
```

**Expected behavior**
I am not sure. But I tried the code (provided in README) with release tag 0.4.0 and there is no issue.

**Additional context**
I tried to debug the issue. It seems to me that we are expecting response from kafka of type `type Response = protocol.Message` and we are getting different response. I checked console of kafka and zookeeper and there seems to be no issue

Stack trace of the error:
```
panic: BUG: promise resolved with impossible value of type <nil>

goroutine 36 [running]:
github.com/segmentio/kafka-go.async.await(0xc00007a120, 0x1435480, 0xc00018c140, 0x142f040, 0xc00014a2d0, 0xc0001660c0, 0x0)
[...]

PR #504, opened 2020-09-09
Fix for panic when RequiredAcks is set to RequireNone

neilcook:
This also includes a new regression test in writer_test.go, because the previous tests were not testing RequiredAcks = 0.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -778,6 +778,8 @@ func (p async) await(ctx context.Context) (Response, error) {
 	select {
 	case x := <-p:
 		switch v := x.(type) {
+		case nil:
+			return nil, nil // A nil response is ok (e.g. when RequiredAcks is None)
 		case Response:
 			return v, nil
 		case error:
--- a/writer_test.go
+++ b/writer_test.go
@@ -51,6 +51,10 @@ func TestWriter(t *testing.T) {
 			scenario: "setting a non default balancer on the writer",
 			function: testWriterSetsRightBalancer,
 		},
+		{
+			scenario: "setting RequiredAcks to None in Writer doesn't cause a panic",
+			function: testWriterRequiredAcksNone,
+		},
 	}
 
 	for _, test := range tests {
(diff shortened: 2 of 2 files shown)


neilcook:
Fixes #487

turfaa:
Looks good to me. I need this PR to be merged 😅

transport.go
```diff
@@ -777,13 +777,17 @@ type async chan interface{}
...
 	select {
 	case x := <-p:
-		switch v := x.(type) {
-		case Response:
-			return v, nil
-		case error:
-			return nil, v
-		default:
-			panic(fmt.Errorf("BUG: promise resolved with impossible value of type %T", v))
+		if x == nil {
```
dominicbarnes, line 780: Would it be possible to put `case nil` in the `switch` below rather than have a separate block here?

writer_test.go
```diff
@@ -83,6 +87,30 @@ func testWriterClose(t *testing.T) {
...
+	}
+	defer w.Close()
+
+	msg := Message{
+		Key:   []byte("ThisIsAKey"),
+		Value: []byte("Test message for required acks test")}
+
+	err := w.WriteMessages(context.Background(), msg)
+	if err != nil {
+		t.Error(err)
```
dominicbarnes, line 109: Can we switch this to `t.Fatal(err)` rather than following up with `t.FailNow()` since that's the intent anyways?

dominicbarnes (changes requested):
I've asked for a couple small tweaks, but overall the change looks good to me!

writer_test.go
neilcook: Sure

transport.go
neilcook: np

achille-roussel (approved):
Thanks a lot of the fix!

achille-roussel:
v0.4.4 was released with this fix.

PR #509, opened 2020-09-14
consumer groups: ensure read deadline is set on broker interactions

stevevls:
Since the Conn does not have a timeout by default, it's possible for a
connection to hang if the remote side crashes before responding.  This
PR ensures that every interaction with the broker has a deadline so that
the consumer group won't get stuck if the coordinator goes down.

### DIFF
--- a/consumergroup.go
+++ b/consumergroup.go
@@ -55,6 +55,10 @@ const (
 	// defaultPartitionWatchTime contains the amount of time the kafka-go will wait to
 	// query the brokers looking for partition changes.
 	defaultPartitionWatchTime = 5 * time.Second
+
+	// defaultReadTimeout is the deadline to set when interacting with the
+	// consumer group coordinator.
+	defaultReadTimeout = 5 * time.Second
 )
 
 // ConsumerGroupConfig is a configuration object used to create new instances of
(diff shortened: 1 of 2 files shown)


consumergroup.go
```diff
@@ -143,6 +147,17 @@ type ConsumerGroupConfig struct {
...
+	// ReadTimeout is the network timeout used when communicating with the
+	// consumer group coordinator.  This value should not be too small since
+	// errors communicating with the broker will generally cause a consumer
+	// group rebalance, and it's undesirable that a transient network error
+	// intoduce that overhead.  Similarly, it should not be too large or the
+	// consumer group may be slow to respond to the coordinator failing over
+	// to another broker.
+	//
+	// Default: 5s
+	ReadTimeout time.Duration
```
achille-roussel, line 159: It looks like a few operations are writes instead of reads. Should we maybe only expose one `Timeout` and use `SetDeadline` instead of `SetReadDeadline`?

stevevls: Sure.  I guess I don't fully understand the subtleties of the read v. write deadlines.  Will change to your suggestion.

achille-roussel (approved):

PR #510, opened 2020-09-15
Fix consumer group timeout handling

stevevls:
PR #509 introduced a bug that could create infinite rebalance loops
when the Timeout is less than the RebalanceTimeout because the JoinGroup
call could timeout before the rebalance completes.


achille-roussel (approved):
Looking good, thanks for the fix!

PR #513, opened 2020-09-18
docs(readme): add error-handling to example code

dominicbarnes:
The current README includes quite a few code snippets that include no error-handling. If someone is trying to learn this library for the first time, it can make understanding why things fail much more difficult without first adding this error-handling to their own code.

It's just good hygiene in Go to handle errors, so I think our examples should reflect that, rather than try so hard to keep the code shorter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,23 +50,34 @@ Here are some examples showing typical use of a connection object:
 topic := "my-topic"
 partition := 0
 
-conn, _ := kafka.DialLeader(context.Background(), "tcp", "localhost:9092", topic, partition)
+conn, err := kafka.DialLeader(context.Background(), "tcp", "localhost:9092", topic, partition)
+if err != nil {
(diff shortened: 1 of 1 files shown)


achille-roussel (approved):
Thanks for improving the examples @dominicbarnes 👍

stevevls (approved):
Overall LGTM.  One philosophical question: should we use `panic` instead of `log.Fatal`?  My understanding is that `log.Fatal` has generally fallen out of favor since it calls `os.Exit` and therefore does not execute deferred statements.

dominicbarnes:
@stevevls I'm open to either approach, just as long as we're consistent. I tend to lean towards this as it's likely closer to what a user would do in practice, aside from `return err` which wouldn't compile in these particular examples. That being said, I do see your point about panic and could be convinced to change this if others chime in (otherwise, will probably just merge as-is).

Issue #440, opened 2020-05-05
How to create topic for Kafka Cluster


nitiprabhu:
Can I get an example of creating a topic programmatically? 

 I tried creating a topic as below: 

c, _ := kafka.Dial("tcp", "host:port")
kt := kafka.TopicConfig{Topic: "sometopic", NumPartitions: 1, ReplicationFactor: 1}
e := c.CreateTopics(kt)

But this will work only if the given **host:port** is Kafka Leader. If the **host:port** is not Leader, then I will be getting this error: _**Not Controller: this is not the correct controller for this cluster**_

What is the right way of passing clusters address and creating a topic?

z2665:
you may try this code
```golang
        conn, err := kafka.Dial("tcp", "localhost:9092")
	if err != nil {
		panic(err.Error())
	}
	controller, _ := conn.Controller()
	fmt.Printf("%s %d ", controller.Host, controller.Port)
	conncontroller, err := kafka.Dial("tcp", net.JoinHostPort(controller.Host, strconv.Itoa(controller.Port)))
	if err != nil {
		panic(err.Error())
	}
	err = conncontroller.CreateTopics(topic)
	if err != nil {
		panic(err.Error())
	}
```

aultimus:
I added documentation to the readme which should help answer your question. I'm going to close this issue now, please let me know if your issue is not solved

PR #514, opened 2020-09-18
add docs on how to create topics

aultimus:
* add docs on how to create topics
* add docs on how to connect to leader via another connection

in response to issue https://github.com/segmentio/kafka-go/issues/440

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ topic := "my-topic"
 partition := 0
 
 conn, _ := kafka.DialLeader(context.Background(), "tcp", "localhost:9092", topic, partition)
+defer conn.Close()
 topicConfigs := []kafka.TopicConfig{
     kafka.TopicConfig{
         Topic:             topic,
--- a/README.md
+++ b/README.md
@@ -84,6 +84,35 @@ batch.Close()
 conn.Close()
 ```
 
+```go
+// to create topics
+topic := "my-topic"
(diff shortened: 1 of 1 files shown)


achille-roussel (approved):
Thanks for adding this example @aultimus !

README.md
```diff
@@ -84,6 +84,50 @@ batch.Close()
 conn.Close()
 ```
 
+```go
+// to create topics
+topic := "my-topic"
+partition := 0
+
+conn, err := kafka.DialLeader(context.Background(), "tcp", "localhost:9092", topic, partition)
```
stevevls, line 92: For a new topic, will this call succeed or will it fail with something like `UnknownTopicOrPartition`?

stevevls (commented):
We get a lot of bug reports where people are confused about the asynchronous nature of topic creation.  Would it make sense to add some docs or an example of waiting for the topic to exist?

README.md
aultimus: I ran the code and calling DialLeader with an uncreated topic does not appear to yield an error

aultimus:
> We get a lot of bug reports where people are confused about the asynchronous nature of topic creation. Would it make sense to add some docs or an example of waiting for the topic to exist?

This sounds like another good topics for docs, I hope to address this in a future pr

PR #518, opened 2020-09-23
Test reader.ReadLag() to return zero lag when using consumer group.

ardhitama:


reader_test.go
```diff
@@ -570,6 +570,19 @@ func TestReaderLagWhenConsumerGroupsEnabled(t *testing.T) {
 	}
 }
 
+func TestReaderReadLagReturnsZeroLagWhenConsumerGroupsEnabled(t *testing.T) {
+	r := &Reader{config: ReaderConfig{GroupID: "not-zero"}}
+	lag, err := r.ReadLag(context.Background())
+
+	if err != errNotAvailableWithGroup {
+		t.Fatalf("expected %s; got %s", errNotAvailableWithGroup.Error(), err.Error())
```
nlsun, line 578: no need to call `.Error()` on these

nlsun, line 578: no need to call .Error() on these

nlsun (commented):
small comment, otherwise lgtm

reader_test.go
ardhitama: changed to use %v

nlsun (approved):

PR #519, opened 2020-09-23
fix LeastBytes Balancer logic for partition changes

### LINKED ISSUE #517: panic in Writer run loop
**Describe the bug**

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x9e7650]

goroutine 85 [running]:
github.com/segmentio/kafka-go.(*Writer).run(0xc0000f8b40)
        /data/orange-ci/workspace/vendor/github.com/segmentio/kafka-go/writer.go:488 +0x560
created by github.com/segmentio/kafka-go.NewWriter
        /data/orange-ci/workspace/vendor/github.com/segmentio/kafka-go/writer.go:272 +0x2bd
```

**Kafka Version**
0.10

kafka-go version: v0.3.4
```go
if len(partitions) != 0 {
    selectedPartition := w.config.Balancer.Balance(wm.msg, partitions...)
    writers[selectedPartition].messages() <- wm
} 
```
it seams that the selectedPartition is not in writers, but i dont know why.

Masterlvng:
LeastBytes balancer may select wrong partition, when partitions have changed.

### DIFF
--- a/balancer.go
+++ b/balancer.go
@@ -67,11 +67,8 @@ type leastBytesCounter struct {
 
 // Balance satisfies the Balancer interface.
 func (lb *LeastBytes) Balance(msg Message, partitions ...int) int {
-	for _, p := range partitions {
-		if c := lb.counterOf(p); c == nil {
-			lb.counters = lb.makeCounters(partitions...)
-			break
-		}
+	if lb.partitionChanged(partitions...) {
+		lb.counters = lb.makeCounters(partitions...)
 	}
 
 	minBytes := lb.counters[0].bytes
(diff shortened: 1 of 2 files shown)


dominicbarnes (changes requested):
Just requesting a little change to the tests, overall the solution here looks correct to me so would like to merge it once this minor change is made.

balancer.go
```diff
@@ -89,6 +86,18 @@ func (lb *LeastBytes) Balance(msg Message, partitions ...int) int {
 	return c.partition
 }
 
+func (lb *LeastBytes) partitionChanged(partitions ...int) bool {
+	if len(partitions) != len(lb.counters) {
+		return true
+	}
+	for _, p := range partitions {
```
nlsun, line 93: if it's true that partitions are always passed in 0-indexed and monotonically increasing with no gaps, i think it's possible we only need the length check that you added (and don't need to iterate `partitions` or the `counterOf` function anymore), thoughts?

Masterlvng: Yes, you 're right.  Length checking is enough.  
May I remove the counterOf function in this pr, because it 's not used?

balancer_test.go
```diff
@@ -277,3 +277,50 @@ func TestMurmur2Balancer(t *testing.T) {
 		}
 	})
 }
+
+func TestLeastBytes(t *testing.T) {
+	testCases := []struct {
+		Label      string
```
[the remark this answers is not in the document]
Masterlvng: ok, change to map[string]struct{ ... }

dominicbarnes (approved):

PR #520, opened 2020-09-24
Added topic list get functionality

yartashes:


aultimus:
looks like a duplicate of https://github.com/segmentio/kafka-go/pull/408 which is currently under review

nlsun:
i think https://github.com/segmentio/kafka-go/pull/408 is specifically for a topic scanner that filters based on a regex whereas this one appears to be a function that lists topics which seems to overlap with https://github.com/segmentio/kafka-go/pull/521

aultimus:
> i think #408 is specifically for a topic scanner that filters based on a regex whereas this one appears to be a function that lists topics which seems to overlap with #521

@nlsun I think that they are duplicates as #408 still adds a public ListTopic method which this PR also adds

nlsun:
hi @yartashes, does the following documentation sufficiently address this use case? https://github.com/segmentio/kafka-go/blob/9cb6db918c3b84d0f50788c550fa0f7a5706c9b5/README.md#to-list-topics

yartashes:
Hello, but the option that you propose is a workaround, if the kafka gives
the option only get the topics why all the partitions were taken and will
get the topics

nlsun:
This functionality has been implemented in a sub-package `topics` as of https://github.com/segmentio/kafka-go/pull/663

PR #521, opened 2020-09-24
add docs on how to list topics

aultimus:
This pull request adds documentation on how to get a list of all topics in kafka.

This documentation addresses how to list topics in the current version of kafka-go without adding new functionality.
This removes the need for this PR: https://github.com/segmentio/kafka-go/pull/520

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ if err := conn.Close(); err != nil {
 }
 ```
 
-### create topics
+### To Create Topics
 ```go
 // to create topics
 topic := "my-topic"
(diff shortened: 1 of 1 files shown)


aultimus:
If this is popular functionality would it make more sense to add a function to return a unique list of topics rather than the documentation?

achille-roussel:
Potentially, let's see how things evolve after improving our documentation? We could also open an issue to collect feedback, and maybe other ideas about which APIs we'd like to add?

achille-roussel (approved):

yartashes:
@aultimus if kafka api makes it possible to directly get a list of topics why not use it. I don't understand what is the point of getting the list in an implicit way

aultimus:
@nlsun @yartashes Once the public topic scanner interface lands lets update the docs to show how to use it, but these docs will suffice in the meantime

Issue #386, opened 2019-12-14
TopicReader: A reader for all partitions of a topic without using a consumer group


fvosberg:
**Describe the solution you'd like**

Now and then, I need to read messages from a topic with no need for consumer groups, because we want to:
- Have all instances of a service consume all partitions
- Want to build a little in memory projection of this topic while booting the service (so no persistent offset management is needed)

Most of the time, we are implementing something like this:

```
func InitTopicReader(cfg kafka.ReaderConfig) (TopicReader, error) {
   // make a connection to the first broker
  // get alle partitions for the topic
  // init a reader for each partition
  // fanInMessages and errors
}
```

My questions regarding this:
- Are you interested in this feature? I would be happy to help
- How should rebalancing be handled in this screnario?
- What functions should it support?

I think these would be useful:

```
func (r *TopicReader) FetchMessage(context.Context) (Message, error)
func (r *TopicReader) ReadLag(context.Context) (int64, error)
func (r *TopicReader) Lag() (int64)
func (r *TopicReader) ReadPartitionLag(context.Context, int) (int64, error)
func (r *TopicReader) PartitionLag(int) int64
func (r *TopicReader) Partitions() ([]Partition, error)
func (r *TopicReader) Stats() ReaderStats
func (r *TopicReader) Close() error
```

Cheers
Fredi

nlsun:
Hi there, I'm probably missing some context but is there a reason you want to avoid reading from a consumer group?

fvosberg:
Hi,

yes, I've tried to point the reasons out above, but I'm not a native speaker so let me try again:

We are using kafka as our main persistence. The services, which don't need a separate persistence, read the data they need on startup. This is fast and migrations of data schemas are easy to deploy. The services use an ephemeral database or use the data just in memory. So:

- every instance of the service has to read all partitions (so if you want to use a consumer group, every instance would have to need one)
- the offsets should not be persisted, as a restart should reset the offsets

Am I missing another option (I don't want a consumer group for every lifecycle of an instance and have to have so many stale consumer groups in my brokers)

Cheers

nlsun:
One option is to pin a consumer group to each instance and reset the offset of the consumer group upon startup of the instance. Although this has other complications like how to associate a consumer group without an external datastore or hardcoding.

[...]

PR #522, opened 2020-09-24
feat: first draft of the reader group

fvosberg:
Reader groups are a way of consuming messages from multiple partitions
without consumer groups. This is useful, e.g. when you are using
compacted topics and don't replicate the state from Kafka. Then you can
use these reader groups for consuming all messages to build a projection
of the data in memory or in an ephemeral storage

Relates to #386

### DIFF
--- a/readergroup.go
+++ b/readergroup.go
@@ -1,0 +1,139 @@
+package kafka
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
+)
+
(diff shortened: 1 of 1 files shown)


fvosberg:
This is just a first draft to get some input if this is something which has the potential of beeing merged into the master. If thats the case, I'll start writing tests and make sure the code is working as expected, it's more kind of a scribble. 

It is related to #386 as mentioned in the PR

fvosberg:
> This does not seem to do what was expected ;-)

Thanks, I hope this works better ;) But this is a draft anyway, so I would love to get some feedback, if this has a chance of beeing merged, so I can invest time in tests

PR #525, opened 2020-09-29
Add PartitionBalancer to sepcific a single partition in single topic

Dashift:
I need write messages to a specific partition, and I noticed that the `Partition` field in `Message` is readonly ( I don't understand why design as this...), so I provide this balancer.  

I don't know whether this is a good idea...

How about use `Message.Partition` when  using specific a `NoneBalancer` like this:

```go
type NoneBalancer struct {
}

func (p *PartitionBalancer) Balance(msg Message, _ ...int) (partition int) {
	return msg.Partition
}
```

### DIFF
--- a/balancer.go
+++ b/balancer.go
@@ -283,3 +283,16 @@ func murmur2(data []byte) uint32 {
 
 	return h
 }
+
+
+// PartitionBalancer is a Balancer that always write Message to a specific partition
+//
+// Note that if you specific a none exist partition, write func will return an error,
+// and the logger of the Writer instance will write down this error.
+type PartitionBalancer struct {
+	p int
+}
+
+func (p *PartitionBalancer) Balance(_ Message, _ ...int) (partition int) {
+	return p.p
+}
--- a/balancer_test.go
+++ b/balancer_test.go
@@ -354,7 +354,7 @@ func TestLeastBytes(t *testing.T) {
 }
 
 func TestPartitionBalancer(t *testing.T) {
-	soecificPartition:=2
+	specificPartition:=2
 
 	testCases := map[string]struct {
 		Keys       [][]byte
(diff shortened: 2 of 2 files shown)


nlsun:
Hi there, could you explain what you're planning on using the ability to write to a specific partition for?

Dashift:
> Hi there, I think I'm missing some context, could you explain what you're planning on using the ability to write to a specific partition for?

Hello, I'm trying to keep the the messages produced, consumed in same sequence. I think I should produce them in a partition of a single topic  and consume in a single thread.   

In sarama I can use  [manualPartitioner](https://github.com/Shopify/sarama/blob/master/partitioner.go#L72) like this
```
        conf := sarama.NewConfig()
	conf.Producer.Return.Successes=true

	conf.Producer.Partitioner=sarama.NewManualPartitioner // ManualPartitioner

	producer, err := sarama.NewSyncProducer(splitBrokers, conf)
	producer.SendMessage(&sarama.ProducerMessage{
		Topic: *topic, Key: nil,
		Value: sarama.StringEncoder(fmt.Sprintf("test message %v/%v from kafka-client-go-test at %s", i+1, bulkSize, now)),
		Partition: 7, // specific the partition 
	})
```

nlsun:
I see, have you run into issues with this approach (implementing a custom balancer)? Or are you looking to add this to the package?

Dashift:
My apologies for not reply sooner. 

There are two types of balancer in my pr:
    1.  Specific partition in the balancer (the codes I had committed, and I use it in my own project)
    2. A custom balancer that allows user specific partition in `Message.Partition` field (might be conflict with origin design)

I think the kafka client library should keep expose the ability of partitioning to users. And I want provide some help, but I don't know which one is better.

nlsun:
Just to check my understanding:
1. This is the `PartitionBalancer` you implemented in this PR?
2. Is this the test case in this PR that uses partition 2? Or is this referring to a different Balancer?

Dashift:
1. Yes, this is the `PartitionBalancer` I had implemented.
2. The test case uses partition 2. It should be only publish to partition 2.

nlsun:
I see, glad to see that you were able to implement a Balancer that suits your needs. However, we're not looking to maintain this use case at this time

Dashift:
That's ok, I'll close this pr.

PR #529, opened 2020-10-05
docs(writer): indicate that v0.3 does not support 0 required acks

dominicbarnes:
We've decided that fixing the lack of support for 0 required acks is not going to happen in v0.3, due to the extra complexity to add this coupled with the fact that v0.4 already includes support for this. This documentation update makes this clear while recommending users upgrade if needed.


achille-roussel (approved):
Thanks for the update @dominicbarnes !

PR #543, opened 2020-10-17
fix zstd decoder leak

achille-roussel:
This PR aims to resolve a memory leak that would occur under some circumstances when reading from partitions containing batches compressed with zstd.

The zstd encoders and decoders are retained by the goroutines they start internally, which causes their finalizers to never go called. The solution implemented here consists in adding a layer of indirection by wrapping the encoders and decoders with an intermediary object on which we install the finalizers.

Related issue https://github.com/klauspost/compress/issues/264

### DIFF
--- a/compress/compress_test.go
+++ b/compress/compress_test.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"io/ioutil"
 	"math/rand"
-	"net"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -117,7 +116,8 @@ func TestCompressedMessages(t *testing.T) {
 
 func testCompressedMessages(t *testing.T, codec pkg.Codec) {
 	t.Run("produce/consume with"+codec.Name(), func(t *testing.T) {
-		topic := createTopic(t, 1)
+		topic := makeTopic()
+		createTopic(t, topic, 1)
 		defer deleteTopic(t, topic)
 
 		w := &kafka.Writer{
@@ -185,18 +185,22 @@ func testCompressedMessages(t *testing.T, codec pkg.Codec) {
 }
 
 func TestMixedCompressedMessages(t *testing.T) {
-	topic := createTopic(t, 1)
+	topic := makeTopic()
+	createTopic(t, topic, 1)
 	defer deleteTopic(t, topic)
 
 	offset := 0
 	var values []string
 	produce := func(n int, codec pkg.Codec) {
 		w := &kafka.Writer{
-			Addr:        kafka.TCP("127.0.0.1:9092"),
-			Topic:       topic,
-			Compression: kafka.Compression(codec.Code()),
+			Addr:  kafka.TCP("127.0.0.1:9092"),
+			Topic: topic,
 		}
 		defer w.Close()
+
+		if codec != nil {
+			w.Compression = kafka.Compression(codec.Code())
+		}
 
 		msgs := make([]kafka.Message, n)
 		for i := range msgs {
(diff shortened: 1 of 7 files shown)


compress/zstd/zstd.go
```diff
@@ -90,52 +102,73 @@ func (r *reader) WriteTo(w io.Writer) (n int64, err error) {
...
 }
 
 // Close implements the io.Closer interface.
 func (w *writer) Close() error {
-	if w.enc == nil {
-		return nil // already closed
+	if w.enc != nil {
+		// Close needs to be called to write the end of stream marker and flush
+		// the buffers. The zstd package documents that the encoder is re-usable
+		// fater being closed.
```
nlsun, line 142:
```suggestion
		// after being closed.
```

compress/zstd/zstd.go
```diff
@@ -90,52 +102,73 @@ func (r *reader) WriteTo(w io.Writer) (n int64, err error) {
...
+			w.err = err
+		}
+		w.enc.Reset(devNull{}) // don't retain the underyling writer
+		w.c.encoderPool.Put(w.enc)
+		w.enc = nil
+		return err
 	}
-	err := w.enc.Close()
-	w.c.encPool.Put(w.enc)
-	w.c = nil
```
nlsun, line 121: we don't set this nil anymore?

compress/zstd/zstd.go
```diff
@@ -90,52 +102,73 @@ func (r *reader) WriteTo(w io.Writer) (n int64, err error) {
...
-	w.enc = nil
-	w.err = io.ErrClosedPipe
```
nlsun, line 123: how come we store `err` instead of `io.ErrClosedPipe` now?

nlsun (commented):
haven't looked at the test code yet

compress/compress_test.go
```diff
@@ -88,19 +88,24 @@ func testEncodeDecode(t *testing.T, m kafka.Message, codec pkg.Codec) {
 	t.Run("encode with "+codec.Name(), func(t *testing.T) {
 		r1, err = compress(codec, m.Value)
 		if err != nil {
-			t.Error(err)
+			t.Fatal(err)
 		}
 	})
 
 	t.Run("decode with "+codec.Name(), func(t *testing.T) {
+		if r1 == nil {
```
nlsun, line 96: is r1 only nil if the first test fails? in that case does the `t.Fatal` cover this already?

compress/compress_test.go
```diff
@@ -407,58 +417,72 @@ func benchmarkCompression(b *testing.B, codec pkg.Codec, buf *bytes.Buffer, payl
...
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
 func makeTopic() string {
 	return fmt.Sprintf("kafka-go-%016x", rand.Int63())
 }
 
-func createTopic(t *testing.T, partitions int) string {
+func newLocalClientAndTopic() (*kafka.Client, string, func()) {
```
nlsun, line 428: seems to mostly be a copy of what's in client_test.go, would it help to move this to the `testing` package too?

compress/compress_test.go
achille-roussel: `r1` may be `nil` if the first test is not run, which is what this check is intended to cover (for example if `-run` is used in the `go test` invocation).

compress/compress_test.go
achille-roussel: Unfortunately it would result in a cycle between the top level `kafka` package and `kafka/testing`, we'd have to refactor the tests to all be in the `kafka_test` package for that to work.

compress/zstd/zstd.go
achille-roussel: It's accessory, we're not making decisions based on this field so setting it to `nil` or not.

compress/zstd/zstd.go
achille-roussel: I missed this, I'll fix.

nlsun (approved):

achille-roussel:
Thanks for the reviews @nlsun !

naughtyGitCat:
encode have leak too, go 1.18, kafka-go v0.4.31, centos7
```
File: rhizobium
Type: inuse_space
Time: May 27, 2022 at 2:46pm (CST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 1082.17MB, 99.01% of 1093.03MB total
Dropped 77 nodes (cum <= 5.47MB)
Showing top 10 nodes out of 38
      flat  flat%   sum%        cum   cum%
  990.72MB 90.64% 90.64%   990.72MB 90.64%  github.com/klauspost/compress/zstd.encoderOptions.encoder
   54.26MB  4.96% 95.60%    54.26MB  4.96%  Rhizobium/cli.Prepare
   21.73MB  1.99% 97.59%    21.73MB  1.99%  github.com/klauspost/compress/zstd.(*blockEnc).init
   11.51MB  1.05% 98.64%    11.51MB  1.05%  runtime.allocm
    3.96MB  0.36% 99.01%    28.69MB  2.62%  github.com/klauspost/compress/zstd.(*Encoder).Reset
         0     0% 99.01%   988.21MB 90.41%  github.com/klauspost/compress/zstd.(*Encoder).Close
         0     0% 99.01%   988.21MB 90.41%  github.com/klauspost/compress/zstd.(*Encoder).EncodeAll
         0     0% 99.01%   974.30MB 89.14%  github.com/klauspost/compress/zstd.(*Encoder).initialize
         0     0% 99.01%   988.21MB 90.41%  github.com/klauspost/compress/zstd.(*Encoder).nextBlock
         0     0% 99.01%    15.67MB  1.43%  github.com/klauspost/compress/zstd.(*doubleFastEncoder).Reset
```

naughtyGitCat:
```go
type KafkaRawMSG struct {
	Topic      string
	KeyBytes   []byte
	ValueBytes []byte
}

type Ungeziefer map[string]*kafka.Writer


func InitKafka() Ungeziefer {
	ungeziefer := make(map[string]*kafka.Writer)
	for key, value := range KafkaTopicMap {
		kafkaLogger.Debugf("Now spawn kafka %s writer %s", key, value)
		ungeziefer[value] = newKafkaWriter(value)
	}
	return ungeziefer
}


func (u Ungeziefer) PublishLog() {
	for msg := range KafkaPublishChan {
		w := u[msg.Topic]
		kafkaLogger.Debug(fmt.Sprintf("Now publish log content to kakfa, key:%s, value:%s, channel approximately size %d",
			string(msg.KeyBytes), string(msg.ValueBytes), len(KafkaPublishChan)))
		err := w.WriteMessages(context.Background(), kafka.Message{Key: msg.KeyBytes, Value: msg.ValueBytes})
		kafkaLogger.Infof(" [x] Sent %#s\n", msg)
		if err != nil {
			kafkaLogger.Warnf("Failed to publish a message: %s, message size: %d bytes", msg, unsafe.Sizeof(msg), err)
			SendError(err, "Failed to publish a kafka message")
		}
	}
}

/*
	new kafka writer
*/
func newKafkaWriter(topic string) *kafka.Writer {
	d := &kafka.Dialer{SASLMechanism: plain.Mechanism{Username: common.CONF.KafkaLQ.User,
		Password: common.CONF.KafkaLQ.Password}}
		Dialer:           d,
		Brokers:          []string{fmt.Sprintf("%s:%d", common.CONF.KafkaLQ.Host, common.CONF.KafkaLQ.Port)},
		Topic:            topic,
		BatchBytes:       56 * 1024 * 1024, // MB,
		CompressionCodec: zstd.NewCompressionCodec(),
		BatchTimeout:     8 * time.Millisecond,
		Async:            true,
	})
	return w
}
```

```go
		adapter.RabbitPublishChan = make(chan adapter.RabbitMessage, 888)
		franz := adapter.InitKafka()
		go franz.PublishLog()
```

PR #545, opened 2020-10-21
improve custom resolver support by allowing port to be overridden

### LINKED ISSUE #231: Respect the port from Resolver in Dialer.
We'd like to implement the `Resolver` interface to adapt our name discovery system.
Our name resolving sys returns not only host but also port. However, the current `Dialer` implementation abandons the resolved port. Would we add an option to opt in port resolving?

dominicbarnes:
This PR improves the support for a custom `Resolver` by allowing it to return a port that overrides the original address. I took the opportunity to consolidate this same logic to a helper function and added some test cases for each of the scenarios I could think of.

This should fix #231

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -65,7 +65,13 @@ type Dialer struct {
 	// support keep-alives ignore this field.
 	KeepAlive time.Duration
 
-	// Resolver optionally specifies an alternate resolver to use.
+	// Resolver optionally gives a hook to convert the broker address into an
+	// alternate host or IP address which is useful for custom service discovery.
+	// If a custom resolver returns any possible hosts, the first one will be
+	// used and the original discarded. If a port number is included with the
+	// resolved host, it will only be used if a port number was not previously
+	// specified. If no port is specified or resolved, the default of 9092 will be
+	// used.
 	Resolver Resolver
 
 	// TLS enables Dialer to open secure connections.  If nil, standard net.Conn
--- a/dialer.go
+++ b/dialer.go
@@ -320,20 +320,10 @@ func (d *Dialer) authenticateSASL(ctx context.Context, conn *Conn) error {
 	return nil
 }
 
-func (d *Dialer) dialContext(ctx context.Context, network string, address string) (net.Conn, error) {
-	if r := d.Resolver; r != nil {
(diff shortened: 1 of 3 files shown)


dialer.go
```diff
@@ -441,3 +435,23 @@ func splitHostPort(s string) (host string, port string) {
...
+	}
+
+	// if the resolver doesn't return anything, we'll allow the caller to continue
+	// with the original address
+	if len(resolved) == 0 {
+		return "", nil
+	}
+
+	resolvedHost, resolvedPort := splitHostPort(resolved[0])
+	if resolvedPort == "" {
```
achille-roussel, line 453: Should we check for `inputPort != ""` instead?

Basically having a logic where we pick the port number from the address if it's set, otherwise fallback to the one from the resolver, and otherwise default to 9092?

dialer.go
```diff
@@ -441,3 +435,23 @@ func splitHostPort(s string) (host string, port string) {
 	}
 	return
 }
+
+func runCustomResolver(ctx context.Context, address string, resolver Resolver) (string, error) {
```
achille-roussel, line 439: How about renaming to `lookupHost` since this is a helper to call `LookupHost` on the resolver?
```suggestion
func lookupHost(ctx context.Context, address string, resolver Resolver) (string, error) {
```

dialer.go
dominicbarnes: I guess that depends on the priority order we want to use, my thinking here was that we should prefer the resolved port, and if not supplied we'll fall back on the one provided in the original `address`. But yeah, just realizing if neither includes a port number, we'll get into a weird state, so I'll add a test case to cover that.

dialer.go
dominicbarnes: I'm fine with this, the name I chose is a little awkward.

dialer.go
achille-roussel: Thinking about user experience, it would be really strange to explicitly specify a port number and have the system connect to something else because the resolver had another port registered. Potentially, this could also be a security concern.

achille-roussel (approved):
This is looking great, thanks for the PR 👍 

Could we document this new behavior on the `Resolver` interface?

Issue #540, opened 2020-10-15
Topic Created as side effect of DialLeader


tlinford:
**Describe the bug**
Following the example in [README](https://github.com/segmentio/kafka-go#to-create-topics), the topic gets created with the call to DialLeader. This seems to rely on `auto.create.topics.enable` being on by default in kafka config.

Running the example with the option set to false causes DialLeader to never return.

Creating a connection to leader explicitly works as expected, so possibly the documentation is just a bit misleading.
I think this could also be related to #436.

**Kafka Version**
2.6.0

**To Reproduce**
set `auto.create.topics.enable=false` in `server.properties`
run example code from [README](https://github.com/segmentio/kafka-go#to-create-topics)

**Expected behavior**
the topic gets created with the call to CreateTopics, and not as a side effect of DialLeader.

aultimus:
Thank you for your bug report @tlinford. I was able to reproduce this behaviour using this example program:

```
package main
import (
	"context"
	"fmt"
	"math/rand"
	"time"
	"github.com/segmentio/kafka-go"
)
func listTopics(conn *kafka.Conn) {
	partitions, err := conn.ReadPartitions()
	if err != nil {
		panic(err.Error())
	}
	m := map[string]struct{}{}
	for _, p := range partitions {
		m[p.Topic] = struct{}{}
	}
	fmt.Println("topics that exist:")
	for k := range m {
		fmt.Println(k)
	}
}
func main() {
	rand.Seed(time.Now().UTC().UnixNano())
	topic := fmt.Sprintf("topic-%d", rand.Intn(1000000))
	fmt.Printf("creating topic %s\n", topic)
	partition := 0
	conn, err := kafka.DialLeader(context.Background(), "tcp", "localhost:9092", topic, partition)
	if err != nil {
		panic(err.Error())
	}
	defer conn.Close()
	listTopics(conn)
}
```

example output:
```
12:24:28 matthew.ault@aulty-mac:runkafka$ go run create-topics/main.go 
creating topic topic-149182
topics that exist:
topic-149182
```

As a result, I have updated the documentation here https://github.com/segmentio/kafka-go/pull/546. I hope that this satisfies your request.

aultimus:
@tlinford does this satisfy your request?

tlinford:
totally, thanks a lot for the clarification!

aultimus:
thanks for your contribution :)

PR #546, opened 2020-10-23
correct kafka create-topics docs

aultimus:
This pr resolves issue https://github.com/segmentio/kafka-go/issues/540. This PR updates the documentation so that it is consistent with `KAFKA_AUTO_CREATE_TOPICS_ENABLE` config option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,18 +101,18 @@ if err := conn.Close(); err != nil {
 ```
 
 ### To Create Topics
-By default kafka has the `KAFKA_AUTO_CREATE_TOPICS_ENABLE='true'`. If this value is set to `'true'` then topics will be created as a side effect of `kafka.DialLeader` like so:
+By default kafka has the `auto.create.topics.enable='true'` (`KAFKA_AUTO_CREATE_TOPICS_ENABLE='true'` in the wurstmeister/kafka kafka docker image). If this value is set to `'true'` then topics will be created as a side effect of `kafka.DialLeader` like so:
 ```go
-// to create topics when KAFKA_AUTO_CREATE_TOPICS_ENABLE='true'
+// to create topics when auto.create.topics.enable='true'
 conn, err := kafka.DialLeader(context.Background(), "tcp", "localhost:9092", "my-topic", 0)
 if err != nil {
     panic(err.Error())
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -101,12 +101,22 @@ if err := conn.Close(); err != nil {
 ```
 
 ### To Create Topics
+By default kafka has the `KAFKA_AUTO_CREATE_TOPICS_ENABLE='true'`. If this value is set to `'true'` then topics will be created as a side effect of `kafka.DialLeader` like so:
```
achille-roussel, line 104: `KAFKA_AUTO_CREATE_TOPICS_ENABLE` is an environment variable of the docker image we use, I don't think it's a standard variable recognized by kafka itself (tho I could be wrong).

Maybe refer to this feature more literally as "automatic topic creation" ?

tlinford: This also does not allow passing options for the topic

README.md
```diff
@@ -101,12 +101,22 @@ if err := conn.Close(); err != nil {
...
+```
+
+If `KAFKA_AUTO_CREATE_TOPICS_ENABLE='false'` then you will need to create topics explicitly like so:
+```go
+// to create topics when KAFKA_AUTO_CREATE_TOPICS_ENABLE='false'
 topic := "my-topic"
 partition := 0
 
-conn, err := kafka.DialLeader(context.Background(), "tcp", "localhost:9092", topic, partition)
+conn, err := kafka.Dial("tcp", "localhost:9092")
```
tlinford, line 119: As mentioned in #440 This could fail with `error = [41] Not Controller: this is not the correct controller for this cluster`

README.md
aultimus: @achille-roussel good point, in kafka downloaded from [here](https://kafka.apache.org/quickstart) `auto.create.topics.enable` is a config setting in `config/server.properties`, I can update the docs to reflect the two different representations for the same setting

README.md
aultimus: https://github.com/segmentio/kafka-go/pull/546/commits/099f26bc8cd695f9fb926383499dd16c565befec should address this error

README.md
aultimus: addressed by https://github.com/segmentio/kafka-go/pull/546/commits/fd254c74da45417c13b65fa6ce127969e582eacf

achille-roussel (approved):
Thanks for the update!

PR #547, opened 2020-10-25
0.4: reduce memory footprint

achille-roussel:
The goal of this PR is to reduce the memory footprint of internal data types by using `int16` or `int32` instead of `int` where possible.

The switch also brings extra type safety because we don't need to convert integers types to `int` anymore, allowing for stricter compiler checks.


Pryz (approved):
lgtm !

achille-roussel:
Thanks for the review!

PR #544, opened 2020-10-19
Add DescribeConfigs, AlterConfigs & CreatePartition implementation

Luffy110:
Implemented DescribeConfigs, AlterConfigs and CreatePartition functionalities for fixing [496](https://github.com/segmentio/kafka-go/issues/469) and [310](https://github.com/segmentio/kafka-go/issues/310).

Kafka Improvement Proposals :
https://cwiki.apache.org/confluence/display/KAFKA/KIP-195%3A+AdminClient.createPartitions .
https://cwiki.apache.org/confluence/display/KAFKA/KIP-133%3A+Describe+and+Alter+Configs+Admin+APIs.

APIs:
[The_Messages_DescribeConfigs](https://kafka.apache.org/protocol#The_Messages_DescribeConfigs).
[The_Messages_AlterConfigs](https://kafka.apache.org/protocol#The_Messages_AlterConfigs)
[The_Messages_CreatePartitions](https://kafka.apache.org/protocol#The_Messages_CreatePartitions)

detailed information ,please looking at [470](https://github.com/segmentio/kafka-go/pull/470).

PR #559, opened 2020-11-19
[WIP] Support additional APIs and flexible messages

yolken-segment:
## Description
This change extends kafka-go to support some additional APIs and newer protocols.

The new APIs (added into the existing client) are:

1. `AlterPartitionReassignments`
2. `CreatePartitions`
3. `DescribeConfigs`
4. `ElectLeaders`
5. `IncrementalAlterConfigs`

It also adds some existing APIs into the client more directly so they can be more easily used by `topicctl` and other apps:

1. `ApiVersions`
2. `DescribeGroups`
3. `ListGroups`

The `AlterPartitionReassignments` API only supports the new, "flexible" message format described in https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields. Since this isn't supported in kafka-go currently, this change also adds basic support for this new protocol.

The proposed interface for this is via extra tags on struct fields, e.g.:

```
type Request struct {
    [non-optional fields]

    TaggedField0 `kafka:"min=v0,max=v0,tagId=0"`
    TaggedField1 `kafka:"min=v0,max=v0,tagId=1"`
   ...
}
```

If a version has any `tagId` fields set, then it's considered "flexible" and the new protocol is used. Otherwise, we continue to use the old, non-flexible protocol.

For now, it looks like tags are not actively used in the API, so I've just put placeholders into the `AlterPartitionReassignments` structs so that the encoder and decoder know to use the "flexible" format.

I've tested these changes on a v2.4.1 cluster via `topicctl`, but if the initial approach looks good I can add more unit tests into this change.

Some other, miscellaneous notes:
1. KIP-482 defines new, "compact" encodings that are different from the ones currently defined in the kafka-go procotol. I've renamed the latter from "compact" to "var", e.g. `VarString`, `VarBytes`, etc. to distinguish them.
2. In "flexible" messages, it seems that all strings and arrays are compact by default. So, there is no need to add extra tags to indicate these things.

### DIFF
--- a/alterpartitionreassignments.go
+++ b/alterpartitionreassignments.go
@@ -25,6 +25,14 @@ type AlterPartitionReassignmentsRequestAssignment struct {
 type AlterPartitionReassignmentsResponse struct {
 	ErrorCode    int16
 	ErrorMessage string
+
+	PartitionResults []AlterPartitionReassignmentsResponsePartitionResult
+}
+
+type AlterPartitionReassignmentsResponsePartitionResult struct {
+	PartitionID  int32
+	ErrorCode    int16
+	ErrorMessage string
 }
 
 func (c *Client) AlterPartitionReassignments(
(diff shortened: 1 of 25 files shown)


godwhoa:
Correct me if I'm wrong but I think #544 is also trying to achieve the same thing.

yolken-segment:
@godwhoa thanks for pointing that out, hadn't seen that. The overlap is only partial, so will wait for that other change to land in the v0.4 branch and then update my pull request to just have the additional features (e.g., tagged field support + `AlterPartitionReassignments`).

yolken-segment:
Closing this out and then splitting it into two separate pull requests, one for the flexible message protocol changes and the second for the additional APIs.

Issue #553, opened 2020-11-05
Is is possible to write to different topics with a single writer?


Nossiac:
The `Writer` example implies that it is bond to a single topic, but i have to dynamically route my messages to different topics based on message content. 
So, is there any way to specifing message topic dynamically with a single writer?

salakhovilia:
when creating a writer via NewWriter (), panic is called with the message "cannot create a kafka writer with an empty topic" same message when trying to create a writer directly and sending messages via WriteMessages

dominicbarnes:
@salakhovilia have you updated to the latest version of the v0.4 branch? #561 was recently merged and added support for this feature. If you have any trouble with that, please open a new issue so we can track it.

Issue #53, opened 2017-12-29
allow Writer to publish to any topic


savaki:
Now that Message supports Topic, it seems like there's no reason not to allow WriteMessages to publish to more than one topic.  Until the next major release, we could leave the public api alone and allow topics passed directly on the message to override the default topic.

Thoughts?

achille-roussel:
I'm not against this change, my only concern is this will require quite a bit of a rework of the internals of `Writer` which would increase the complexity of the code, since we'd have to maintain a connection pool for each topic. There will be limitations tho since a `Writer` still has only one `Balancer` attached, it means the balancing logic will have to be the same for every topics that the writer publishes to.

Also if we take this route, should we also allow direct addressing of partitions when writing to a topic (effectively bypassing the balancer)?

In practice it seems to me that programs rarely need to publish to multiple topics, and if they do a `map[string]*kafka.Writer` may be more practical.

Feel free to chime in if you have examples of use cases that would clearly benefit from having the `Writer` support publishing to multiple topics.

savaki:
While I like the idea of maintaining a connection pool per broker, a smaller step might be just to expand writeProduceRequestV2 to publish to more than one topic.  The underlying kafka protocol supports it, but it looks like for simplicity, it's been limited to just 1.

https://github.com/segmentio/kafka-go/blob/master/write.go#L216

As for only a single balancer, I think it's quite common for organizations to use a just a single balancer.  However, if multiple balancers are needed, multiple writers could be created.

savaki:
> Also if we take this route, should we also allow direct addressing of partitions when writing to a topic (effectively bypassing the balancer)?

We don't have any use cases where we'd ever directly assign to a partition without a balancer.  Most cases that I can think of that would desire direct routing, like QoS and priority routing, seems like they'd be better served by being put in a balancer anyway.

savaki:
> In practice it seems to me that programs rarely need to publish to multiple topics, and if they do a map[string]*kafka.Writer may be more practical.

We have a number of services that publish to multiple topics.  Some examples include:

* database change data capture where each table is published to a separate topic
* data extraction.  a service might extract N types of data from a single provider e.g. Salesforce.  Each data type would be published onto a different topic.

While it's possible to use map[string]*kafka.Writer for such use cases, it feels like it pushes complexity down to the users for common use cases for something that seems like a reasonably small change to the writer.

achille-roussel:
[...]

PR #561, opened 2020-11-20
feat(writer): add support for writing messages to multiple topics

dominicbarnes:
This PR adds support for writing messages to topics dynamically with a single writer.

Previously, you needed to define `Writer.Topic` and it was always required. Now, if you choose not to specify this configuration, then `Message.Topic` **must** always be set for each written message.

These patterns are mutually exclusive, to help avoid some potential pitfalls, such as consuming a message from 1 topic and auto-writing it back to the same topic in an infinite loop. If both the writer and a message are configured with a topic, an error will be returned. In the opposite case, if neither the writer nor the message are configured with a topic, an error will be returned.

This should address a few issues that have been created:
 - #553 
 - #53

### DIFF
--- a/message.go
+++ b/message.go
@@ -10,10 +10,13 @@ import (
 
 // Message is a data structure representing kafka messages.
 type Message struct {
-	// Topic is reads only and MUST NOT be set when writing messages
+	// Topic may be set to specify the topic for this message, but only if the
+	// Writer is not already configured with a Topic.
+	//
+	// See the documentation for WriterConfig.Topic for more information.
 	Topic string
 
-	// Partition is reads only and MUST NOT be set when writing messages
+	// Partition is read-only and MUST NOT be set when writing messages
 	Partition int
 	Offset    int64
 	Key       []byte
--- a/message.go
+++ b/message.go
@@ -10,10 +10,10 @@ import (
 
 // Message is a data structure representing kafka messages.
 type Message struct {
-	// Topic may be set to specify the topic for this message, but only if the
-	// Writer is not already configured with a Topic.
+	// Topic indicates which topic this message was consumed from via Reader.
 	//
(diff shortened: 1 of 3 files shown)


message.go
```diff
@@ -10,10 +10,13 @@ import (
 
 // Message is a data structure representing kafka messages.
 type Message struct {
-	// Topic is reads only and MUST NOT be set when writing messages
+	// Topic may be set to specify the topic for this message, but only if the
```
achille-roussel, line 13: Maybe add documentation as well to explain that for messages returned by a reader, the field is set to the topic that the message was read from?

writer.go
```diff
@@ -957,6 +966,24 @@ func (w *Writer) Stats() WriterStats {
...
+
+		return w.Topic, nil
+	} else {
+		if msg.Topic == "" {
+			// either the writer or the message must specify a topic
+			return "", InvalidTopic
+		}
+
+		return msg.Topic, nil
+	}
```
achille-roussel, line 984: Maybe this function can be expressed with a bit less branching to help readers grasp the logc?

```go
if (w.Topic != "" && msg.Topic != "") || (w.Topic == "" && msg.Topic == "") {
    return "", InvalidMessage
}
if msg.Topic != "" {
    return msg.Topic, nil
}
return w.Topic, nil
```
This could be a personal preference tho, if the form looks harder to read to you it's probably fine to keep it the way it is.

achille-roussel (approved):
🚀

writer.go
dominicbarnes: Ah yeah, I was going back and forth with various ways to accomplish this, this looks more concise so I'll adopt it.

message.go
dominicbarnes: Good callout!

Issue #539, opened 2020-10-14
More auth mechanisms support


yuzhichang:
sarama supports [following SASL mechanisms](https://github.com/Shopify/sarama/blob/master/broker.go):

- PLAIN
- SCRAM
- GSSAPI(Kerberos V5)
- OAUTH10A or OAUTHBEARER (I'm not sure which one)

kafka-go doesn't support GSSAPI and OAUTH currently. I wish kafka-go provides them, especially GSSAPI(Kerberos V5).

yuzhichang:
I notice there's a branch named `sasl-gssapi` tries to add Kerberos support. However this branch is inactive for several months.

achille-roussel:
There is work in progress in https://github.com/segmentio/kafka-go/pull/598 to support GSSAPI.

Generally speaking, we are happy to provide exposure to third party packages that implement kafka-go extensions like new SASL mechanisms, but it is unlikely that we will be extending the number of authentication mechanisms that we maintain in kafka-go.

Issue #237, opened 2019-03-18
SASL: add support for GSSAPI mechanism


stevevls:
**Describe the solution you'd like**
Once #223 is merged, we can add support for GSSAPI (Kerberos) authentication.

**Supporting documentation**
https://cwiki.apache.org/confluence/display/KAFKA/KIP-43%3A+Kafka+SASL+enhancements
https://docs.confluent.io/current/kafka/authentication_sasl/authentication_sasl_gssapi.html

psilvaro:
Is there any news about having the support for GSSAPI? 
Thanks

stevevls:
No movement as of yet.  If you're interested in taking a stab at it, I think it's a matter of:
1. Adding a kerberos server to the docker compose setup
1. Configuring the kafka container to add another listener that connects to the kerberos server
1. Implementing the `sasl.Mechanism`.  There are examples in `sasl/plain` and `sasl/scram` that could serve as a jumping off point.

This isn't currently a critical feature for Segment, so it's one of those when we get around to it sorts of things.  We're happy to accept PRs, though! 😄

PR #563, opened 2020-11-23
feat(sasl): add GSSAPI as authentication mechanism

mrolla:
This PR adds GSSAPI via Kerberos authentication mechanism.

Addresses https://github.com/segmentio/kafka-go/issues/237 and partly https://github.com/segmentio/kafka-go/issues/539.

Currently, it's missing docker compose support for testing e2e, but I'll add that asap.


chy373:
I need this feature very much! can this pull request be reviewed?

mentalisttraceur:
**Caution**! As far as I know, no implementation of GSSAPI for this library can work in the general case unless it has a change in `dialer.go` to pass the broker's hostname/FQDN from the dialer code into the SASL mechanism!

It will only work in the special case where you only need to dial one broker with that mechanism. (Or I guess when every broker somehow has the same SPN, but I'm not sure if this is common or deprecated/discouraged?)

(More comments when I have time.)

mentalisttraceur:
I say my last comment from having added a GSSAPI mechanism for this library back in mid-November for my employer, where we have a cluster with many Kafka brokers, and we have the use case of reading/dialing multiple brokers with one `kafka.Reader`/`kafka.Dialer`.

We intend to do a pull request soon (but for legal reasons I'm not speaking for my employer, and I'm not promising/guaranteeing anything). But y'all know how it goes when it's part of a job:

1. I'm on a small team with many competing priorities, and

2. it's still going through some prolonged testing in our environment to see if there are any subtle issues that crop up after prolonged execution, with large volumes of data, etc.

If this was my own personal contribution I would just paste in here all of the relevant diffs/code I wrote, but I'm not authorized to just casually make that decision on my own. So for now I'll just say:

1. In the SPN (`kafka/domain` in this pull request's example), only the `kafka` part seems reasonably possible to configure for an entire Kafka cluster - the `domain` part needs to be different each broker.

2. Currently, in this library, one `sasl.Mechanism`/`kafka.Dialer` can be used for connecting to multiple brokers.

3. It is a non-obvious regression and pitfall for human error if the GSSAPI SASL mechanism in particular can only work with one broker at a time. It would force users of this library to connect to each broker individually when for example a single `kafka.Reader` might have been sufficient. It would take some knowledge of the internals/details of Kerberos/GSSAPI for a user of this library to even know that this is a potential problem.

4. The `sasl.Mechanism` code can't know which broker `Start` is being called for unless code in `dialer.go` passes that information through before/during calling `Start`.

mrolla:
That's some great feedback @mentalisttraceur, thanks!

I did develop and test this in a very controlled environment (you guessed, one broker) and wanted to have this out in the open to collect some feedbacks and use cases as my exposure to Kerberos and GSSAPI in general is kinda limited.

I will be back working on this very soon with the changes to the dialer to propagate the domain and in general testing with multiple brokers scenarios.

If in the meantime you manage to open a PR that supersedes this one, please feel free to ping me and I'll gladly close mine.

achille-roussel:
Should we close in favor of https://github.com/segmentio/kafka-go/pull/598 ?

mrolla:
@achille-roussel I'm going ahead and close this PR. Reading what you suggested in #598, I'm going ahead and open a new PR which basically just injects the address (needed for GSSAPI) into the authentication functions via the context (idiomatic in a way).

PR #565, opened 2020-11-26
fix default fetch.min.bytes to 1,

eaglexiang:
following [official document](https://kafka.apache.org/documentation/#fetch.min.bytes)

> The default setting of 1 byte means that fetch requests are answered as soon as a single byte of data is available or the fetch request times out waiting for data to arrive. Setting this to something greater than 1 will cause the server to wait for larger amounts of data to accumulate which can improve server throughput a bit at the cost of some additional latency.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -24,6 +24,13 @@ const (
 	defaultCommitRetries = 3
 )
 
+const (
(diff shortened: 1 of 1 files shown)


achille-roussel:
The change looks good, thanks for the fix!

PR #567, opened 2020-11-30
add auto create topic

salakhovilia:


achille-roussel:
Hello @salakhovilia, thanks for the submission!

Do you mind giving a bit more details about the use case for this feature? I believe kafka servers can be configured to automatically create topics as well, any reasons this wasn't working for you?

salakhovilia:
I tried to send a message to a non-existent topic and this caused the Unknown topic error, I thought that this function was not implemented and as I understood that to implement this function, you need to send a request to the broker, maybe I'm wrong, because I don't fully understand the work with api. These commits are not working, I'm still trying to solve this problem. @achille-roussel,Please explain what I can do wrong?

PR #571, opened 2020-12-08
Add support for additional broker APIs

yolken-segment:
## Description
This change updates the kafka-go `Client` to expose some additional APIs that are needed to support zk-less `topicctl`:

1. `AlterPartitionReassignments`
2. `ApiVersions`
3. `DescribeGroups`
4. `ElectLeaders`
5. `IncrementalAlterConfigs`
6. `ListGroups`

It also fixes the implementation of `DescribeConfigs` when fetching broker configs- based on my testing, I think these need to be split by broker.

## Testing
- Added some unit tests
- Tested the new functionality in the `topicctl` v1 branch: https://github.com/segmentio/topicctl/pull/32


alterpartitionreassignments.go
```diff
@@ -0,0 +1,121 @@
...
+	BrokerIDs []int
+}
+
+// AlterPartitionReassignmentsResponse is a response from the AlterPartitionReassignments API.
+type AlterPartitionReassignmentsResponse struct {
+	// ErrorCode is set to a non-zero value if a top-level error was encountered.
+	ErrorCode int
+
+	// ErrorMessage describes the top-level error that occured.
+	ErrorMessage string
```
achille-roussel, line 42: We've expose errors as `error` types in other exported APIs of the client (for example https://pkg.go.dev/github.com/segmentio/kafka-go#FetchResponse), what do you think about continuing this pattern?

(and similarly in other types that include ErrorCode/ErrorMessage)

describegroups_test.go
```diff
@@ -48,3 +53,114 @@ func TestDescribeGroupsResponseV0(t *testing.T) {
...
 	}
 }
+
+func TestClientDescribeGroups(t *testing.T) {
+	if os.Getenv("KAFKA_VERSION") == "2.3.1" {
+		// There's a bug in 2.3.1 that causes the MemberMetadata to be in the wrong format and thus
+		// leads to an error when decoding the DescribeGroupsResponse.
+		//
+		// See https://issues.apache.org/jira/browse/KAFKA-9150 for details.
+		t.Skip("Skipping because kafka version is 2.3.1")
```
achille-roussel, line 63: Oh Kafka...

protocol/alterpartitionreassignments/alterpartitionreassignments.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Detailed API definition: https://kafka.apache.org/protocol#The_Messages_AlterPartitionReassignments
+type Request struct {
+	TimeoutMs int32          `kafka:"min=v0,max=v0"`
+	Topics    []RequestTopic `kafka:"min=v0,max=v0"`
+
+	// We need at least one tagged field to indicate that this is a "flexible" message
+	// type.
+	_ struct{} `kafka:"min=v0,max=v0,tag"`
```
achille-roussel, line 16: If it doesn't change the semantics, moving this field to the first position will offer a tighter packing in memory:
```suggestion
	// We need at least one tagged field to indicate that this is a "flexible" message
	// type.
	_ struct{} `kafka:"min=v0,max=v0,tag"`

	TimeoutMs int32          `kafka:"min=v0,max=v0"`
	Topics    []RequestTopic `kafka:"min=v0,max=v0"`

```

protocol/alterpartitionreassignments/alterpartitionreassignments.go
```diff
@@ -0,0 +1,61 @@
...
+
+type Response struct {
+	ThrottleTimeMs int32            `kafka:"min=v0,max=v0"`
+	ErrorCode      int16            `kafka:"min=v0,max=v0"`
+	ErrorMessage   string           `kafka:"min=v0,max=v0,nullable"`
+	Results        []ResponseResult `kafka:"min=v0,max=v0"`
+
+	// We need at least one tagged field to indicate that this is a "flexible" message
+	// type.
+	_ struct{} `kafka:"min=v0,max=v0,tag"`
```
achille-roussel, line 45: If it doesn't change the semantics, moving this field to the first position will offer a tighter packing in memory:

```suggestion
	// We need at least one tagged field to indicate that this is a "flexible" message
	// type.
	_ struct{} `kafka:"min=v0,max=v0,tag"`

	ThrottleTimeMs int32            `kafka:"min=v0,max=v0"`
	ErrorCode      int16            `kafka:"min=v0,max=v0"`
	ErrorMessage   string           `kafka:"min=v0,max=v0,nullable"`
	Results        []ResponseResult `kafka:"min=v0,max=v0"`
```

describegroups.go
```diff
@@ -1,8 +1,147 @@
...
+	apiResp := protoResp.(*describegroups.Response)
+	resp := &DescribeGroupsResponse{
+		Groups: []DescribeGroupsResponseGroup{},
+	}
+
+	for _, apiGroup := range apiResp.Groups {
+		group := DescribeGroupsResponseGroup{
+			GroupID:    apiGroup.GroupID,
+			GroupState: apiGroup.GroupState,
+			Members:    []DescribeGroupsResponseMember{},
```
nlsun, line 115: initializing the slice here and above is usually not necessary

describegroups.go
```diff
@@ -174,3 +313,119 @@ func (t *describeGroupsResponseV0) readFrom(r *bufio.Reader, sz int) (remain int
...
+// for protocol details.
+func decodeMemberMetadata(rawMetadata []byte) (DescribeGroupsResponseMemberMetadata, error) {
+	mm := DescribeGroupsResponseMemberMetadata{}
+
+	if len(rawMetadata) == 0 {
+		return mm, nil
+	}
+
+	buf := bytes.NewBuffer(rawMetadata)
+	bufReader := bufio.NewReader(buf)
```
nlsun, line 330: you can use the `bytes.Buffer` returned from `bytes.NewBuffer` as a reader below, unless you wanted the buffering here

describegroups.go
```diff
@@ -174,3 +313,119 @@ func (t *describeGroupsResponseV0) readFrom(r *bufio.Reader, sz int) (remain int
...
+	var err error
+	var version16 int16
+
+	if remain, err = readInt16(bufReader, remain, &version16); err != nil {
+		return mm, err
+	}
+	mm.Version = int(version16)
+
+	if remain, err = readStringArray(bufReader, remain, &mm.Topics); err != nil {
+		return mm, err
```
nlsun, line 342: might be more clear return the zero value, the `mm` here has `Version` set
```suggestion
		return DescribeGroupsResponseMemberMetadata{}, err
```

nlsun (commented):
just minor comments, +1 on @achille-roussel 's comments, LGTM otherwise

alterpartitionreassignments_test.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if resp.ErrorCode != 0 {
+		t.Error(
+			"Bad error code in response",
+			"expected", 0,
+			"got", resp.ErrorCode,
+		)
+	}
```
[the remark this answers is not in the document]
yolken-segment: Yeah, so I deliberately didn't use testify (as much as I'd like to) because it seems like the vast majority of the tests in the repo don't. If it's ok to use testify, maybe we should migrate all of our existing tests in a separate change?

protocol/alterpartitionreassignments/alterpartitionreassignments.go
yolken-segment: Fixed.

protocol/alterpartitionreassignments/alterpartitionreassignments.go
yolken-segment: Fixed.

alterpartitionreassignments.go
yolken-segment: Makes sense, done!

describegroups.go
achille-roussel: Alternatively, `bytes.Reader` with `bytes.NewReader(rawMetadata)` could be an interesting option.

protocol/describeconfigs/describeconfigs.go
```diff
@@ -16,13 +24,55 @@ type Request struct {
...
 
+func (r *Request) Split(cluster protocol.Cluster) (
+	[]protocol.Message,
+	protocol.Merger,
+	error,
+) {
+	messages := []protocol.Message{}
+	topicsMessage := Request{
+		Resources: []RequestResource{},
+	}
```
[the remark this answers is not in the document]
yolken-segment: Fixed.

describegroups.go
yolken-segment: Fixed.

describegroups.go
yolken-segment: I think the `read...` helpers we use below specifically require a `bufio.Reader` as opposed to a generic `io.Reader` since they're peeking ahead. Let me know if I'm missing something here?

describegroups.go
yolken-segment: I think this is ok since the caller of this function just discards the decoded metadata if `err` is not `nil`.

describegroups.go
achille-roussel: You are correct 👍

protocol/incrementalalterconfigs/incrementalalterconfigs.go
```diff
@@ -0,0 +1,65 @@
...
+type RequestConfig struct {
+	Name            string `kafka:"min=v0,max=v0"`
+	ConfigOperation int8   `kafka:"min=v0,max=v0"`
+	Value           string `kafka:"min=v0,max=v0,nullable"`
+}
+
+func (r *Request) ApiKey() protocol.ApiKey { return protocol.IncrementalAlterConfigs }
+
+func (r *Request) Broker(cluster protocol.Cluster) (protocol.Broker, error) {
+	// TODO: Support updating multiple broker configs in single request?
```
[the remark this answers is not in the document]
yolken-segment: Done and added a test for this.

describegroups.go
yolken-segment: Ok :-).

yolken-segment:
Thanks for the feedback @achille-roussel and @nlsun! I think I addressed all of the feedback, and I also tested the latest changes with `topicctl` v1 to ensure that the new APIs still work.

achille-roussel (approved):
Changes look good to me 👍 

@nlsun are there more changes you would like to see?

nlsun (approved):
lgtm

PR #576, opened 2020-12-21
Export Batch.Partition

vmihailenco:
This allows to commit processed batch with:

```go
gen.CommitOffsets(map[string]map[int]int64{
    "topic": map[int]{
        batch.Partition(): batch.Offset(),
    },
})
```

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -346,7 +346,7 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
 		}
 	}()
 
-	var state = p.grabState()
+	state := p.grabState()
 	var response promise
 
 	switch m := req.(type) {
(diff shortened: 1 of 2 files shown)


transport.go
```diff
@@ -346,7 +346,7 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
 		}
 	}()
 
-	var state = p.grabState()
+	state := p.grabState()
```
achille-roussel, line 349: Were these style changes done by a formatting tool?

achille-roussel (approved):
Looking good, thanks for the contribution!

transport.go
vmihailenco: Yes, this is how https://github.com/mvdan/gofumpt formats.

The second commit fixes https://github.com/segmentio/kafka-go/issues/582 and slipped into this PR accidentally. But hopefully it is still fine...

wziww:
When do this PR can be released ？We do need [this](https://github.com/segmentio/kafka-go/pull/576/files#diff-5982a4ca350449c8a5deba675bc4bc4bb0f1823c45e6c7ba0076cc6b159fef8fR497-R499) to fix the reconnection-bug in some situation    :)

PR #596, opened 2021-01-24
Kafka Client FindCoordinator implementation

live-wire:
### This pull request contains
- FIndCoordinator implementation that utilizes the `kafka.(*Client)` abstraction.
- A tiny unit test for the same.

---
According to [this design doc](https://docs.google.com/document/d/11Jqy_GjUGtdXJK94XGsEIK7CP1SnQGdp2eF0wSw9ra8/edit#heading=h.dz78fc8xlsd5), the first step in implementing a **Transactional Producer** is looking for a TransactionCoordinator for the specified transactionalId using `FindCoordinatorRequest`. 

---
This PR is a step towards solving #233

CC: @achille-roussel

### DIFF
--- a/error.go
+++ b/error.go
@@ -9,12 +9,6 @@ import (
 // https://kafka.apache.org/protocol#protocol_error_codes
 type Error int
 
-// ErrorWithCode retains Error code information while creating an error message
-type ErrorWithCode struct {
-	Error error
-	Code  int16
-}
-
 const (
 	Unknown                            Error = -1
 	OffsetOutOfRange                   Error = 1
@@ -560,13 +554,6 @@ func makeError(code int16, message string) error {
 		return Error(code)
 	}
 	return fmt.Errorf("%w: %s", Error(code), message)
-}
-
-func makeErrorWithCode(code int16, message string) *ErrorWithCode {
-	return &ErrorWithCode{
-		Error: makeError(code, message),
-		Code:  code,
-	}
 }
 
 // WriteError is returned by kafka.(*Writer).WriteMessages when the writer is
--- a/error.go
+++ b/error.go
@@ -9,6 +9,12 @@ import (
 // https://kafka.apache.org/protocol#protocol_error_codes
 type Error int
 
+// ErrorWithCode retains Error code information while creating an error message
+type ErrorWithCode struct {
+	Error error
+	Code  int16
+}
+
 const (
 	Unknown                            Error = -1
 	OffsetOutOfRange                   Error = 1
@@ -554,6 +560,13 @@ func makeError(code int16, message string) error {
 		return Error(code)
 	}
 	return fmt.Errorf("%w: %s", Error(code), message)
+}
+
+func makeErrorWithCode(code int16, message string) *ErrorWithCode {
+	return &ErrorWithCode{
+		Error: makeError(code, message),
+		Code:  code,
+	}
 }
 
 // WriteError is returned by kafka.(*Writer).WriteMessages when the writer is
--- a/findcoordinator.go
+++ b/findcoordinator.go
@@ -85,7 +85,7 @@ func (c *Client) FindCoordinator(ctx context.Context, req *FindCoordinatorReques
 		Coordinator: coordinator,
 	}
 
-	return ret, ret.Error
+	return ret, nil
 }
 
 // FindCoordinatorRequestV0 requests the coordinator for the specified group or transaction
--- a/findcoordinator.go
+++ b/findcoordinator.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"strings"
 	"time"
 
 	"github.com/segmentio/kafka-go/protocol/findcoordinator"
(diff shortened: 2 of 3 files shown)


findcoordinator.go
```diff
@@ -2,8 +2,120 @@ package kafka
...
+		Error:       makeError(res.ErrorCode, res.ErrorMessage),
+		ErrorCode:   int(res.ErrorCode),
+		Coordinator: coordinator,
+	}
+
+	return ret, ret.Error
+}
+
+// WaitForCoordinatorIndefinitely is a blocking call till a coordinator is found
+func (c *Client) WaitForCoordinatorIndefinitely(ctx context.Context, req *FindCoordinatorRequest) (*FindCoordinatorResponse, error) {
```
achille-roussel, line 97: I would prefer to not include this method, and keep the `kafka.(*Client)` API a 1:1 mapping to Kafka APIs.

The one-second blocking call may also not be the right fit for many applications, which could need exponential backoff, or respecting context cancellation, etc...

It isn't very complex for a program to re-implement this logic, maybe we could even put it as an example in the `kafka.(*Client).FindCoordinator` documentation?

findcoordinator.go
```diff
@@ -2,8 +2,120 @@ package kafka
...
+// FindCoordinatorResponseCoordinator contains details about the found coordinator
+type FindCoordinatorResponseCoordinator struct {
+	// NodeID holds the broker id.
+	NodeID int32
+
+	// Host of the broker
+	Host string
+
+	// Port on which broker accepts requests
+	Port int32
```
achille-roussel, line 46: We've taken the approach of not exposing implementation details like the size of the integer types in other public APIs, let's stay consistent?
```suggestion
	// NodeID holds the broker id.
	NodeID int

	// Host of the broker
	Host string

	// Port on which broker accepts requests
	Port int
```

findcoordinator.go
```diff
@@ -2,8 +2,120 @@ package kafka
...
+	Throttle time.Duration
+
+	// An error that may have occurred while attempting to retrieve Coordinator
+	//
+	// The error contains both the kafka error code, and an error message
+	// returned by the kafka broker.
+	Error error
+
+	// Error code
+	ErrorCode int
```
achille-roussel, line 64: Since we already inject the error in the `Error` field, this field seems redundant?
```suggestion
```

achille-roussel (commented):
Thanks for your contribution @live-wire !

I left a few comments, let me know what you think about it.

findcoordinator.go
live-wire: Makes sense. This shouldn't be a part of the `kafka.(*Client)` API. Will do the needful.

findcoordinator.go
live-wire: Roger that.

findcoordinator.go
live-wire: The current `makeError` implementation returns an `error` object from the errorCode provided.
Once that is done, the error code information can not be accessed again. There should be a way to access error code information as well to ensure downstream functions that make use of `FindCoordinator` have more visibility. 

I will refactor this to still retain error code information.

live-wire:
Hello @achille-roussel ,

I made the following updates:
- Removed WaitForCoordinator function from the `kafka.(*Client)` API.
- Moved the waitingForCoordinator indefinitely to the test file. (The test needs wait for coordinator initialisation).
- Created a new struct to hold error and error code information.
- Made the other suggested fixes.

Please have a look.

findcoordinator.go
achille-roussel: I don't think that's accurate, the error returned by `makeError` is either a `kafka.Error`, or wraps one.

The underlying `kafka.Error` can always be retrieved with:
```go
var kafkaError kafka.Error
if errors.As(err, &kafkaErr) {
  // use kafkaError
  ...
}
```

error.go
```diff
@@ -9,6 +9,12 @@ import (
 // https://kafka.apache.org/protocol#protocol_error_codes
 type Error int
 
+// ErrorWithCode retains Error code information while creating an error message
+type ErrorWithCode struct {
+	Error error
+	Code  int16
+}
```
achille-roussel, line 16: See my response to this discussion https://github.com/segmentio/kafka-go/pull/596#discussion_r576451053

I don't think we need to introduce this new public API.

findcoordinator.go
live-wire: Interesting. TIL 😮  
Will make the necessary modification.

live-wire:
@achille-roussel I made the suggested changes.

error.go
live-wire: Done.

findcoordinator.go
```diff
@@ -2,8 +2,92 @@ package kafka
...
+		Host:   res.Host,
+		Port:   int(res.Port),
+	}
+	ret := &FindCoordinatorResponse{
+		Throttle:    makeDuration(res.ThrottleTimeMs),
+		Error:       makeError(res.ErrorCode, res.ErrorMessage),
+		Coordinator: coordinator,
+	}
+
+	return ret, ret.Error
```
achille-roussel, line 88: The method should only return an error if no responses were available:
```suggestion
	return ret, nil
```

live-wire: Done. 👍

achille-roussel (approved):
Thanks for the contribution 👍

achille-roussel (approved):

Issue #595, opened 2021-01-21
Reader: consume from multiple topics


alebabai:
**Describe the solution you'd like**
Please, add opportunity to set multiple topics into reader to fetch from.

**Supporting documentation**
For now it supports single topic and pass it into consumerGroup (if GroupID is specified) as a string slice with one element
https://github.com/segmentio/kafka-go/blob/2b3f6824a9578d01c9f065be97459aa179609bdb/reader.go#L656-L663

**Related**
#564

dominicbarnes:
This improvement was recently released with v0.4.12

Issue #564, opened 2020-11-25
Can a consumer consume multiple topics?


zengliming:
**Describe the solution you'd like**

**Supporting documentation**
Please provides links to relevant Kafka protocol docs and/or KIPs.

dominicbarnes:
This is now possible as of v0.4.12

PR #600, opened 2021-02-05
feat(reader): allow consuming from multiple topics with a group

dominicbarnes:
This change allows the `Reader` to consume from multiple topics when being used with a consumer group. To opt into this behavior, you supply `config.GroupTopics` instead of `config.Topic`, the default behavior and config is unchanged.

The following open issues would be resolved by this change:
 - #595
 - #564

I've added test cases for several permutations of this new config:
 - `Topic`
 - `GroupTopics` (one value)
 - `GroupTopics` (multiple values)


achille-roussel (commented):
This change looks great 👍 

We've decided to hold on merging until we have had the opportunity to run a production-scale test internally to build confidence that we are not introducing any regressions.

iAziz786:
@achille-roussel Thanks for your great work!

We have started to use kafka-go in our products and we are in the process of shipping now products and we need to consume from multiple topics by one group. As I can see that this is all set to be release is there anything I can do from my end to close this faster?

If closing is not possible at this point can I you at least guild me with alternatives?

achille-roussel:
@iAziz786 this change should be ready to go. We wanted to do a bit more production testing before we merge it to master to ensure we aren't introducing any regressions, but nothing else is blocking it.

iAziz786:
That's good news, although we have moved to confluent go library for now but will move back to kafka-go as soon as this PR is merged.

xiaokgs:
We have same request to consume from multiple topics, can I ask when this will be merged pls?

dominicbarnes:
We'll proceed with merging this, we would definitely appreciate some feedback after some testing @iAziz786 !

achille-roussel:
@iAziz786 @xiaokgs this has been released as part of the v0.4.12 update 👍

Issue #607, opened 2021-02-25
Empty ReadBatchConfig is treated as valid one


vadzappa:
**Describe the bug**
While validating the ReadBatchConfig:
https://github.com/segmentio/kafka-go/blob/master/conn.go#L755

The following configuration is treated as being valid:
`ReadBatchConfig{}`, while this causes unpredictable results while reading messages from topic (usually only part of the topic got read)

**Kafka Version**
1.1.1

**To Reproduce**
Simple test code:
```go
func TestInvalidReadBatchConfig(t *testing.T) {
	c := &Conn{}
	b := c.ReadBatchWith(ReadBatchConfig{})

	if b.err == nil {
		t.Error("Invalid batch config accepted")
	}
}
```

**Expected behavior**
When empty config provided, error should be returned early, so any further actions will abort early to prevent human mistake

dominicbarnes:
I'm working on addressing this problem with #609 and I've discovered that the failure mode you're describing appears to only impact Kafka 0.10.

The `Reader` sets some sensible defaults for `MinBytes` and `MaxBytes` (1 and 1MB respectively) that you could adopt here, assuming that using `Reader` directly is not an option for you.

PR #608, opened 2021-02-25
avoid invalid empty ReadBatchConfig

vadzappa:
This is simple solution for bug described here:
https://github.com/segmentio/kafka-go/issues/607


dominicbarnes:
Thank you for taking the time to open a PR for this! Please refer to #609 as I've tried my best to address the same issue you are trying to here.

It seems that this only fails for older versions of Kafka (notably 0.10). Rather than change this behavior for all other versions of Kafka, I'm going to recommend that you set explicit configuration when using Kafka 0.10 if upgrading is not feasible.

vadzappa:
Thank you very much for taking care of this! Really appreciate your efforts!
Kudos 👍

PR #609, opened 2021-02-26
Clarify Min/Max Bytes config for consumers

dominicbarnes:
This PR aims to address some concerns raised by #607 and #608 by adding some much-needed clarity on the `MinBytes` and `MaxBytes` configuration used by `Reader` and `Conn.ReadBatch`.

In the case of `ReaderConfig`, default values are documented more clearly, and should be configured explicitly in the event of problems described later on.

In the case of `Conn`, using it with no defined value (ie: 0) for `MinBytes` and `MaxBytes` has undefined behavior. Upon adding a test case for this, it can be observed that in Kafka 0.10, this configuration will result in a batch error, but Kafka 0.11+ are able to handle this without issue. As such, the test case will give an exception to 0.10 and we can revisit if this behavior seems to regress in future versions of Kafka.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -97,7 +97,14 @@ type ConnConfig struct {
 
 // ReadBatchConfig is a configuration object used for reading batches of messages.
 type ReadBatchConfig struct {
+	// MinBytes indicates to the broken the minimum size a batch of messages must
+	// be when consuming. Setting a high minimum on a low-volume topic can result
+	// in messages appearing delayed if they never reach the minimum.
 	MinBytes int
+
+	// MaxBytes indicates to the broker the maximum size a batch of messages must
+	// be when consuming. A broker will truncate a message to respect this config,
+	// so use a value higher than the size of your largest message.
 	MaxBytes int
 
 	// IsolationLevel controls the visibility of transactional records.
--- a/conn.go
+++ b/conn.go
@@ -97,7 +97,7 @@ type ConnConfig struct {
 
 // ReadBatchConfig is a configuration object used for reading batches of messages.
 type ReadBatchConfig struct {
-	// MinBytes indicates to the broken the minimum size a batch of messages must
+	// MinBytes indicates to the broker the minimum size a batch of messages must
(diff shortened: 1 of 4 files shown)


conn.go
```diff
@@ -97,7 +97,14 @@ type ConnConfig struct {
 
 // ReadBatchConfig is a configuration object used for reading batches of messages.
 type ReadBatchConfig struct {
+	// MinBytes indicates to the broker the minimum size a batch of messages must
+	// be when consuming. Setting a high minimum on a low-volume topic can result
+	// in messages appearing delayed if they never reach the minimum.
```
stevevls, line 102: This is a difficult concept to express clearly.  My attempt would be something like:

> MinBytes indicates to the broker the minimum batch size that the consumer will accept.  Setting a high minimum when consuming from a low volume topic may result in delayed delivery because the broker cannot return any messages until it has accumulated enough to satisfy the request.

Though I'd massage that further. 😁

conn.go
```diff
@@ -97,7 +97,14 @@ type ConnConfig struct {
...
 	MinBytes int
+
+	// MaxBytes indicates to the broker the maximum size a batch of messages must
+	// be when consuming. A broker will truncate a message to respect this config,
```
stevevls, line 106: If you edit the first one, I'd also make an update to the first sentence for uniformity, e.g.

> MaxBytes indicates to the broker the maximum batch size that the consumer will accept

reader.go
```diff
@@ -345,8 +345,18 @@ type ReaderConfig struct {
...
+	// in messages appearing delayed if they never reach the minimum.
+	//
+	// Default: 1
 	MinBytes int
+
+	// MaxBytes indicates to the broker the maximum size a batch of messages must
+	// be when consuming. A broker will truncate a message to respect this config,
+	// so use a value higher than the size of your largest message.
+	//
+	// Default: 1MB
```
stevevls, line 359: If you make edits above, then don't forget to propagate them here. 😄

stevevls (approved):

PR #616, opened 2021-03-16
update isTimeout and isTemporary to properly unwrap errors

ianthpun:
the previous implementation does not properly unwrap errors to check if it is a `net.Error`, which was wrapped here:

https://github.com/segmentio/kafka-go/blob/master/produce.go#L139
```
return nil, fmt.Errorf("kafka.(*Client).Produce: %w", err)
```

this leads this issues where the error is not correctly handled when calling `isTemporary` or `isTimeout`

```
error writing messages to TOPIC (partition 12): kafka.(*Client).Produce: dial tcp IP:21948: i/o timeout
```

specifically in the code block here in `writeBatch`
https://github.com/segmentio/kafka-go/blob/master/writer.go#L780-L782

```
		if !isTemporary(err) {
			break
		}
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net"
 )
 
 // Error represents the different error codes that may be returned by kafka.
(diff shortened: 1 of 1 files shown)


error.go
```diff
@@ -491,17 +493,19 @@ func (e Error) Description() string {
 }
 
 func isTimeout(err error) bool {
-	e, ok := err.(interface {
-		Timeout() bool
-	})
-	return ok && e.Timeout()
+	var opError net.Error
```
ianthpun, line 496: the other option that is a _bit_ closer to your previous implementation is to unwrap into an interface that only has timeout

```
func isTimeout(err error) bool {
	var timeoutError interface{ Timeout() bool }
	if errors.As(err, &timeoutError) {
		return timeoutError.Timeout()
	}
	return false
}

func isTemporary(err error) bool {
	var tempError interface{ Temporary() bool }
	if errors.As(err, &tempError) {
		return tempError.Temporary()
	}
	return false
}
```

dominicbarnes: This is what I would prefer, we can add handling for other error types (eg: `net.Error`) as well, but let's not fundamentally change the logic here to start with.

ianthpun: @dominicbarnes updated, lemme know what you think

dominicbarnes (approved):

dominicbarnes:
Thanks for the contribution! I'll post here when this change gets released.

dominicbarnes:
I've tagged this as v0.4.13

PR #634, opened 2021-04-14
Increase the version of pierrec/lz4 module

alexshvid:
Previous version has security issues and this small upgrade can fit them.


achille-roussel:
@alexshvid thanks for reporting the issue. We updated the lz4 package to v2.6.0 and tagged v0.4.14 of kafka-go to release this change.

We also opened #636 to track the further upgrade of the lz4 package to v4 (which has breaking API changes).

I'll close this PR as it seems we addressed the core issue, but feel free to re-open if something is missing 👍

Issue #637, opened 2021-04-17
Basic example producer-random does not work


yngvark:
**Describe the bug**
I'm new to kafka-go and Kafka. I want to do a quick test to get a working producer (and later, a consumer).

```sh
$ go version
go version go1.16.3 linux/amd64
$ uname -r
5.4.0-71-generic

# Using Ubuntu 20.04.2 LTS
```

**Kafka Version**
2.7.0, according to the log (see below).

(I'm using docker image wurstmeister/kafka:2.13-2.7.0.)

**To Reproduce**

```sh
$ git clone https://github.com/segmentio/kafka-go.git
$ cd examples
$ docker-compose up

[ ... ]
kafka_1              | [2021-04-17 06:56:59,246] INFO Kafka version: 2.7.0 (org.apache.kafka.common.utils.AppInfoParser)
kafka_1              | [2021-04-17 06:56:59,246] INFO Kafka commitId: 448719dc99a19793 (org.apache.kafka.common.utils.AppInfoParser)
kafka_1              | [2021-04-17 06:56:59,246] INFO Kafka startTimeMs: 1618642619242 (org.apache.kafka.common.utils.AppInfoParser)
kafka_1              | [2021-04-17 06:56:59,248] INFO [KafkaServer id=1003] started (kafka.server.KafkaServer)
kafka_1              | [2021-04-17 06:56:59,305] INFO [broker-1003-to-controller-send-thread]: Recorded new controller, from now on will use broker 1003 (kafka.server.BrokerToControllerRequestThread)
producer-random_1    | dial tcp 127.0.0.1:9092: connect: connection refused
consumer-mongo-db_1  | 2021/04/17 06:57:00 dial tcp :0: connect: connection refused
mongo-db_1           | 2021-04-17T06:57:00.826+0000 I NETWORK  [conn1] end connection 172.19.0.7:42432 (0 connections now open)
consumer-logger_1    | 2021/04/17 06:57:00 dial tcp :0: connect: connection refused
examples_consumer-mongo-db_1 exited with code 1
examples_consumer-logger_1 exited with code 1
producer-random_1    | dial tcp 127.0.0.1:9092: connect: connection refused
producer-random_1    | dial tcp 127.0.0.1:9092: connect: connection refused
producer-random_1    | dial tcp 127.0.0.1:9092: connect: connection refused
producer-random_1    | dial tcp 127.0.0.1:9092: connect: connection refused
```

Not sure why this happens.

[...]

PR #649, opened 2021-04-30
chore(examples): update kafka-go

dominicbarnes:
Currently, simply running `docker-compose up` in `examples/` doesn't work 100% correctly. When starting up docker-compose containers, it doesn't expose any mechanism to wait on a particular host to respond in a particular way, and kafka has a tendency to boot up pretty slowly.

In the producer-random example, it produces a message every second. One would expect that even if kafka isn't reachable right away, it would eventually succeed once kafka fully boots up. Unfortunately, the older version of kafka-go seems to hold on to the initial error, never able to succeed after failing to start.

By updating kafka-go, this seems to clear up all the problems reported in #637, and thus should fix that issue. It turns out we had pretty old versions of kafka-go hanging around, so they've all been updated to the latest (v0.4.15).


achille-roussel (approved):
🙌

PR #650, opened 2021-04-30
fix(writer): do not use kafka error code for invalid config

### LINKED ISSUE #618: "the message contents does not match its CRC" on invalid topic configuration
**Describe the bug**
When specifying the Topic on both the `kafka.Writer` and `kafka.Message`, the writer returns an `InvalidMessage` error, whose string representation is "the message contents does not match its CRC".

This happens inside `chooseTopic`:

```
func (w *Writer) chooseTopic(msg Message) (string, error) {
	// w.Topic and msg.Topic are mutually exclusive, meaning only 1 must be set
	// otherwise we will return an error.
	if (w.Topic != "" && msg.Topic != "") || (w.Topic == "" && msg.Topic == "") {
		return "", InvalidMessage
	}

	// now we choose the topic, depending on which one is not empty
	if msg.Topic != "" {
		return msg.Topic, nil
	}

	return w.Topic, nil
}
```

**Kafka Version**
Using kafka-go v0.4.11

**To Reproduce**
Set the Topic field on both the Writer and the Message

**Expected behavior**

The returned error should reflect the misconfiguration correctly, i.e. "Invalid topic specified", etc. The current message is incorrect and may lead the developer on a goose chase.

dominicbarnes:
This fixes #618 by using a bespoke error message rather than returning what appears to be a kafka error code, which is confusing. This should make diagnosing this invalid configuration issue much more straightforward.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -975,8 +975,10 @@ func (w *Writer) Stats() WriterStats {
 func (w *Writer) chooseTopic(msg Message) (string, error) {
 	// w.Topic and msg.Topic are mutually exclusive, meaning only 1 must be set
 	// otherwise we will return an error.
-	if (w.Topic != "" && msg.Topic != "") || (w.Topic == "" && msg.Topic == "") {
-		return "", InvalidMessage
+	if w.Topic != "" && msg.Topic != "" {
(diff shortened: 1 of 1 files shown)


achille-roussel (approved):
Thank you @dominicbarnes!

PR #663, opened 2021-05-12
Add ability to get a list of topics matching a regex

benmathews:
The primary use case I envision is subscribing to topics. Example code 

```
re := regexp.MustCompile(`.*go-test.*`)
topics := connection.ListTopicsRegex(*re)
reader := kafka.NewReader(
        kafka.ReaderConfig{
                 Brokers: []string{boostrapServer + ":" + port},
                 GroupID: groupId,
                 GroupTopics: topics,
}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
-localtest: 
-	docker-compose up -d
+test: docker
 	KAFKA_SKIP_NETTEST=1 KAFKA_VERSION=2.3.1 	go test -race -cover ./...
+
+docker: 
+	docker-compose up -d
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 test: docker
-	KAFKA_SKIP_NETTEST=1 KAFKA_VERSION=2.3.1 	go test -race -cover ./...
+	KAFKA_SKIP_NETTEST=1 \
+	KAFKA_VERSION=2.3.1 \
+	go test -race -cover ./...
 
 docker: 
 	docker-compose up -d
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-test: docker
+test: 
 	KAFKA_SKIP_NETTEST=1 \
 	KAFKA_VERSION=2.3.1 \
 	go test -race -cover ./...
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,3 @@
+localtest: 
+	docker-compose up -d
+	KAFKA_SKIP_NETTEST=1 KAFKA_VERSION=2.3.1 	go test -race -cover ./...
--- a/README.md
+++ b/README.md
@@ -666,7 +666,7 @@ docker-compose up -d
 Run tests
 
 ```bash
-KAFKA_VERSION=2.3.1
-KAFKA_SKIP_NETTEST=1
-go test -race ./...
+KAFKA_VERSION=2.3.1 \
+  KAFKA_SKIP_NETTEST=1 \
+  go test -race ./...
 ```
--- a/README.md
+++ b/README.md
@@ -652,3 +652,16 @@ if err := r.Close(); err != nil {
     log.Fatal("failed to close reader:", err)
 }
 ```
+
+## Developer
+
+There are some flaky and non-runable tests. Execute them locally with
+
+```bash
+docker-compose up -d
+export KAFKA_SKIP_NETTEST=1
+export KAFKA_VERSION=2.3.1
+go test -race -cover ./...
+```
+
+This is encapsulated in a Makefile, so optionally call `make`.
--- a/README.md
+++ b/README.md
@@ -670,5 +670,3 @@ KAFKA_VERSION=2.3.1
 KAFKA_SKIP_NETTEST=1
 go test -race ./...
 ```
-
-This is encapsulated in a Makefile, so optionally call `make`.
--- a/README.md
+++ b/README.md
@@ -655,13 +655,20 @@ if err := r.Close(); err != nil {
 
 ## Developer
 
-There are some flaky and non-runable tests. Execute them locally with
+Subtle behavior changes in later Kafka versions have caused some historical tests to break, if you are running against Kafka 2.3.1 or later, exporting the `KAFKA_SKIP_NETTEST=1` environment variables will skip those tests.
+
+Run Kafka locally in docker
 
 ```bash
 docker-compose up -d
-export KAFKA_SKIP_NETTEST=1
-export KAFKA_VERSION=2.3.1
-go test -race -cover ./...
+```
+
+Run tests
+
+```bash
+KAFKA_VERSION=2.3.1
+KAFKA_SKIP_NETTEST=1
+go test -race ./...
 ```
 
 This is encapsulated in a Makefile, so optionally call `make`.
--- a/README.md
+++ b/README.md
@@ -653,7 +653,7 @@ if err := r.Close(); err != nil {
 }
 ```
 
-## Developer
+## Testing
 
 Subtle behavior changes in later Kafka versions have caused some historical tests to break, if you are running against Kafka 2.3.1 or later, exporting the `KAFKA_SKIP_NETTEST=1` environment variables will skip those tests.
 
--- a/topics/list_topics.go
+++ b/topics/list_topics.go
@@ -1,0 +1,35 @@
+package topics
+
+import (
+	"context"
+	"regexp"
+
+	kafkago "github.com/segmentio/kafka-go"
+)
+
+// ListTopics returns a slice of all the Topics
+func ListTopics(ctx context.Context, client *kafkago.Client) (topics []kafkago.Topic, err error) {
+	response, err := client.Metadata(ctx, &kafkago.MetadataRequest{
+		Addr: client.Addr,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return response.Topics, nil
+}
+
+// ListTopicsRegex returns a slice of Topics that match a regex
+func ListTopicsRegex(ctx context.Context, client *kafkago.Client, match regexp.Regexp) (topics []string, err error) {
+	alltopics, err := ListTopics(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, val := range alltopics {
+		if match.Find([]byte(val.Name)) != nil {
+			topics = append(topics, val.Name)
+		}
+	}
+	return topics, nil
+}
--- a/topics/list_topics.go
+++ b/topics/list_topics.go
@@ -27,7 +27,7 @@ func ListTopicsRegex(ctx context.Context, cli *kafka.Client, re *regexp.Regexp) 
 	}
 
 	for _, val := range alltopics {
-		if re.Find([]byte(val.Name)) != nil {
+		if re.MatchString(val.Name) {
 			topics = append(topics, val.Name)
 		}
 	}
--- a/topics/list_topics.go
+++ b/topics/list_topics.go
@@ -7,8 +7,8 @@ import (
 	"github.com/segmentio/kafka-go"
 )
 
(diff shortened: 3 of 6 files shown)


nlsun:
Hi @benmathews, this looks like a similar feature to https://github.com/segmentio/kafka-go/pull/408 except that contributor seems to have gone silent, so we'll be happy to work with you on implementing this.

A couple of thoughts/requests:
- What do you think about creating a `topics` package in this repo, like `github.com/segmentio/kafka-go/topics`? And make a `list_topics.go` for these functions. The purpose of this package would be to provide auxiliary helper functions related to topics
- Note in the package description for `topics` that it's experimental and may have breaking API changes at any time
- Changing the functions to take in the `kafka.Client` as an argument

benmathews:
I'd be happy to make the changes. But:

- Creating a new package means the new functions can't be based on Conn or Client. The new package would have to define it's own local struct.
- I can't find a way to get a Conn from a Client. The implementation in the PR uses Conn's ReadPartition function. Is there a way to get a Conn from a Client? Or are you suggesting I reimplement the ReadPartitions functionality?

mvenable:
@nlsun any update on what the best implementation would be here?  I'd love to keep some momentum here as this is functionality that would be very beneficial to my team.  Thanks

nlsun:
@benmathews @mvenable Sorry for the delay:
- Yup, the function can't be based on Conn or Client, I was thinking it could be a free floating function that takes a `Client` as an argument
- We can use `Client.Metadata` https://pkg.go.dev/github.com/segmentio/kafka-go#Client.Metadata to get the topics, this is what ReadPartitions uses under the hood

benmathews:
@nlsun how does it look now?

topics/list_topics.go
```diff
@@ -0,0 +1,35 @@
+package topics
+
+import (
+	"context"
+	"regexp"
+
+	kafkago "github.com/segmentio/kafka-go"
```
nlsun, line 7: If we don't rename the package it'll default to `kafka` as the package name which I personally think is fine
```suggestion
	"github.com/segmentio/kafka-go"
```

topics/list_topics.go
```diff
@@ -0,0 +1,35 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return response.Topics, nil
+}
+
+// ListTopicsRegex returns a slice of Topics that match a regex
+func ListTopicsRegex(ctx context.Context, client *kafkago.Client, match regexp.Regexp) (topics []string, err error) {
```
nlsun, line 23: Mostly aesthetic comments, but the functional change here is the `*` in the `regexp.Regexp`
```suggestion
func ListTopicsRegex(ctx context.Context, cli *kafkago.Client, re *regexp.Regexp) (topics []string, err error) {
```

nlsun, line 23: Thoughts on adding a test for this?

topics/list_topics.go
```diff
@@ -0,0 +1,35 @@
...
+	alltopics, err := ListTopics(ctx, client)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, val := range alltopics {
+		if match.Find([]byte(val.Name)) != nil {
```
nlsun, line 30:
```suggestion
		if match.MatchString(val.Name) {
```

topics/list_topics.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+	"regexp"
+
+	kafkago "github.com/segmentio/kafka-go"
+)
+
+// ListTopics returns a slice of all the Topics
+func ListTopics(ctx context.Context, client *kafkago.Client) (topics []kafkago.Topic, err error) {
```
nlsun, line 11: Currently the function would be called `topics.ListTopics`, I think it might read better as `topics.List`
```suggestion
// List returns a slice of all the Topics
func List(ctx context.Context, client *kafkago.Client) (topics []kafkago.Topic, err error) {
```

topics/list_topics.go
nlsun, line 23: Same as above about the how it reads to a user of this package
```suggestion
// ListRe returns a slice of Topics that match a regex
func ListRe(ctx context.Context, client *kafkago.Client, match regexp.Regexp) (topics []string, err error) {
```

Makefile
```diff
@@ -0,0 +1,3 @@
+localtest: 
+	docker-compose up -d
+	KAFKA_SKIP_NETTEST=1 KAFKA_VERSION=2.3.1 	go test -race -cover ./...
```
nlsun, line 3: I would personally split up running docker with running the tests, maybe something like
```suggestion
docker: 
	docker-compose up -d
	
test: 
	KAFKA_VERSION=2.3.1 \
		KAFKA_SKIP_NETTEST=1 \
		go test -race ./...
```

nlsun (commented):
Thanks, just some quick comments

topics/list_topics.go
```diff
@@ -0,0 +1,35 @@
...
+	response, err := client.Metadata(ctx, &kafkago.MetadataRequest{
+		Addr: client.Addr,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return response.Topics, nil
```
nlsun, line 19: What do you think about returning a slice of strings instead of `Topic`? If we do so we'll also want to check the `Topic.Error` before returning

README.md
```diff
@@ -652,3 +652,16 @@ if err := r.Close(); err != nil {
     log.Fatal("failed to close reader:", err)
 }
 ```
+
+## Developer
+
+There are some flaky and non-runable tests. Execute them locally with
```
achille-roussel, line 658: Please rephrase:
```suggestion
Subtle behavior changes in later Kafka versions have caused some historical tests to break, if you are running against Kafka 2.3.1 or later, exporting the `KAFKA_SKIP_NETTEST=1` environment variables will skip those tests:
```

topics/list_topics.go
benmathews: I had one with my initial implementation but the testing framework isn't available as this is a new package. I also tried making a mock but there are not interfaces to mock. If you are OK with me adding an interface or two as necessary, I'll go that direction.

topics/list_topics.go
benmathews: I initially had implemented this to return a slice of strings. But it seemed returning Topic fit better with your API. Why would you prefer strings?

README.md
benmathews: I've combined this with some of the other suggestions.

Makefile
benmathews: I've separated the steps. But I want test to be the default and to have a dependency on the docker step so changed your suggestion a little.

README.md
```diff
@@ -652,3 +652,16 @@ if err := r.Close(); err != nil {
...
+
+```bash
+docker-compose up -d
+export KAFKA_SKIP_NETTEST=1
+export KAFKA_VERSION=2.3.1
+go test -race -cover ./...
+```
+
+This is encapsulated in a Makefile, so optionally call `make`.
```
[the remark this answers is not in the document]
benmathews: Done

topics/list_topics.go
benmathews: Done

topics/list_topics.go
benmathews: Done

topics/list_topics.go
benmathews: Done

topics/list_topics.go
benmathews: Done

topics/list_topics.go
benmathews: Done

topics/list_topics.go
nlsun: I was thinking that it might be less to maintain but yeah I agree that it's more general purpose this way, let's keep it `Topic` the way you have it

topics/list_topics.go
nlsun: Hmm yeah it's not a great setup for this as is, if it's not too much extra work we could try moving functions as needed to the `testing` package? https://github.com/segmentio/kafka-go/tree/master/testing

Makefile
nlsun: I don't think we need to be running docker every time, I was thinking it is more common to run `make docker` once, and run `make test` repeatedly?

Also it might've been missed, but I had a suggestion for the formatting of the test as well
```
	KAFKA_VERSION=2.3.1 \
			KAFKA_SKIP_NETTEST=1 \
			go test -race ./...
```

README.md
```diff
@@ -652,3 +652,23 @@ if err := r.Close(); err != nil {
...
+```bash
+docker-compose up -d
+```
+
+Run tests
+
+```bash
+KAFKA_VERSION=2.3.1
+KAFKA_SKIP_NETTEST=1
+go test -race ./...
```
nlsun, line 671: I made a mistake in the formatting earlier
```suggestion
KAFKA_VERSION=2.3.1 \
  KAFKA_SKIP_NETTEST=1 \
  go test -race ./...
```

README.md
```diff
@@ -652,3 +652,23 @@ if err := r.Close(); err != nil {
     log.Fatal("failed to close reader:", err)
 }
 ```
+
+## Developer
```
nlsun, line 656:
```suggestion
## Testing
```

README.md
```diff
@@ -652,3 +652,23 @@ if err := r.Close(); err != nil {
...
+
+Run tests
+
+```bash
+KAFKA_VERSION=2.3.1
+KAFKA_SKIP_NETTEST=1
+go test -race ./...
+```
+
+This is encapsulated in a Makefile, so optionally call `make`.
```
nlsun, line 674: I think we can let people find the makefile on their own
```suggestion
```

topics/list_topics.go
```diff
@@ -0,0 +1,35 @@
+package topics
```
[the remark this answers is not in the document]
benmathews: Done

Makefile
benmathews: I've made the formatting change. 

Running docker compose takes a half second once the kafka cluster is already running. I'd like new users to have a low barrier to entry and consider the time worth it.

topics/list_topics.go
benmathews: I'd rather not extend this PR to include rewriting the test framework.

Makefile
nlsun: I know this nitpicky but I think `make docker test` is as low barrier to entry as `make test`

I think the distinction provides value because it may be confusing to someone who is not familiar with docker or docker compose when kafka containers suddenly start running on their laptop or if a `docker-compose` command fails

topics/list_topics.go
nlsun: If you would rather not copy functions into the `testing` package, what do you think about copying the functions you need into the `list_topics_test.go` file itself?

benmathews: Done

Makefile
benmathews: Done

benmathews:
I believe I've addressed all your concerns @nlsun.

topics/list_topics_test.go
```diff
@@ -0,0 +1,120 @@
+// Package topics is an experimental package that provides additional tooling
+// around Kafka Topics. This package does not make any promises around
+// backwards compatibility.
```
nlsun, line 3: We don't need to repeat the package comment

nlsun (commented):
Thank you! Just 1 thing and we're good to go!

topics/list_topics_test.go
benmathews: Done.

chenzhiwei:
Can someone tell me why not allowing a single consumer to consume multiple topics?

Is this a limitation of Kafka or just because kafka-go does not have such feature?

PR #665, opened 2021-05-14
allow resetting offsets to either earliest or latest

erikdw:
Kafka has a notion of magic sentinel numbers (-1, -2) for consumer offset values [1],
which are meant to signal that the consumer should start from earliest (first) or
latest (last) available offset for a given partition.

Instead of obeying that signal, we would unconditionally respect the default
`StartOffset` value from the configuration when reading a negative offset upon
synchronizing consumergroup state (such as at initialization).

Hence you could *only* have either "read from earliest" or "read from latest".
behavior from your consumers when you reset the offsets to one of these
sentinel values.

With this change, the consumer will respect the returned value if it is -1
or -2, allowing an operator to start a consumer at earliest or latest irrespective
of the configured value of `StartOffset`.

## Example of consumer offsets working

```
% kafka-console-consumer.sh \
  --bootstrap-server localhost:9092 \
  --consumer-property exclude.internal.topics=false \
  --formatter 'kafka.coordinator.group.GroupMetadataManager$OffsetsMessageFormatter' \
  --topic __consumer_offsets | grep -v __consumer_offset 
...
[test-group,topic-default,0]::[OffsetMetadata[5433,NO_METADATA],CommitTime 1620981162537,ExpirationTime 1621067562537]
[test-group,topic-default,0]::[OffsetMetadata[5434,NO_METADATA],CommitTime 1620981162540,ExpirationTime 1621067562540]
[test-group,topic-default,0]::[OffsetMetadata[5435,NO_METADATA],CommitTime 1620981162543,ExpirationTime 1621067562543]
[test-group,topic-default,0]::[OffsetMetadata[5436,NO_METADATA],CommitTime 1620981162545,ExpirationTime 1621067562545]
# Stop the consumer, while letting the producer continue, then reset the offset to -1 (latest)
[test-group,topic-default,0]::[OffsetMetadata[-1,NO_METADATA],CommitTime 1620981192527,ExpirationTime 1621067592527]
# Restart the consumer.
[test-group,topic-default,0]::[OffsetMetadata[6040,NO_METADATA],CommitTime 1620981518165,ExpirationTime 1621067918165]
# Note that the consumer started at latest, instead of the earliest offset.
[test-group,topic-default,0]::[OffsetMetadata[6041,NO_METADATA],CommitTime 1620981518170,ExpirationTime 1621067918170]
# Stop the consumer again, and reset the offset to -2 (earliest)
[test-group,topic-default,0]::[OffsetMetadata[-2,NO_METADATA],CommitTime 1620981551179,ExpirationTime 1621067951179]
[test-group,topic-default,0]::[OffsetMetadata[1,NO_METADATA],CommitTime 1620981631351,ExpirationTime 1621068031351]
# We see the consumer started at the earliest offset instead of the latest.
[test-group,topic-default,0]::[OffsetMetadata[2,NO_METADATA],CommitTime 1620981631354,ExpirationTime 1621068031354]
[test-group,topic-default,0]::[OffsetMetadata[3,NO_METADATA],CommitTime 1620981631358,ExpirationTime 1621068031358]
[test-group,topic-default,0]::[OffsetMetadata[4,NO_METADATA],CommitTime 1620981631361,ExpirationTime 1621068031361]
[test-group,topic-default,0]::[OffsetMetadata[5,NO_METADATA],CommitTime 1620981631364,ExpirationTime 1621068031364]
```

[1] From the kafka wiki for the [ListOffset API](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-OffsetAPI(AKAListOffset)):

> Field: Time
>
> Description:
> Used to ask for all messages before a certain time (ms). There are two special values
> Specify -1 to receive the latest offset (i.e. the offset of the next coming message)
> and -2 to receive the earliest available offset. This applies to all versions of the API.
> Note that because offsets are pulled in descending order, asking for the earliest offset
> will always return you a single element.

### DIFF
--- a/consumergroup.go
+++ b/consumergroup.go
@@ -1113,7 +1113,14 @@ func (cg *ConsumerGroup) fetchOffsets(conn coordinator, subs map[string][]int32)
 				if partition == pr.Partition {
 					offset := pr.Offset
 					if offset < 0 {
-						offset = cg.config.StartOffset
+						// If offset is one of the magic sentinel numbers (-1, -2), respect it.
+						// Whereas if offset is otherwise negative, set it to the default StartOffset.
+						switch offset {
+						case FirstOffset:
+						case LastOffset:
+						default:
+							offset = cg.config.StartOffset
+						}
 					}
 					offsetsByPartition[int(partition)] = offset
 				}


stevevls (commented):
Nice find!  Have you looked at how the Java or librdkafka clients handle this?  We use those as the reference implementations, so it would be best to mimic their behavior.

Also, would it be possible to add a unit test?

erikdw:
As discussed offline with @stevevls , this change is not viable.

Unfortunately, the Kafka protocol does not provide a direct mechanism for an operator to specifically guide a consumer to reset to reading the earliest or latest offsets.  Instead the `-1` sentinel value is intended only to indicate to the consumer that they should reset to whatever their default behavior is:

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-OffsetRequest

> Note that if there is no offset associated with a topic-partition under that consumer group the broker does not set an error code (since it is not really an error), but returns empty metadata and sets the offset field to -1.

We propose to instead provide some tooling (in [`topicctl`](https://github.com/segmentio/topicctl/)) to reset each partition's offset to "earliest" or "latest", based on the current value when the tool is executed.  One issue with such a tool would be that earliest offsets could quickly become invalid if kafka expires data before the consumer starts.  An idea to alleviate this is to add a configurable "fudge factor" of say N% of the available offsets.

abraithwaite:
I'm  fine with us implementing something in `topicctl`, but I don't think we should write off this solution either.

In particular, I think it's typical for the sentinel values `-1` and `-2` to represent earliest and latest as you say and reference.

Digging into the kafka docs, canonical implementation of the java API, the fallback to default behavior you describe was specifically intended to be used in situations where there was no initial offset (i.e. just starting up) or the offset is out of range of the currently tracked offsets (i.e. already expired or somehow in the future).  I don't believe that the fallback to default behavior was intended to be applied in the cases of the sentinel values of `-1` or `-2`, since those are not "naturally" occurring during the course of typical consumer operations.  That is, it has to be placed there on purpose by an operator or a consumer.

https://github.com/apache/kafka/blob/e4f2f6f6e82cafbdea785d53521b96fe062e172d/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L721-L722
http://people.apache.org/~nehanarkhede/kafka-0.9-producer-javadoc/doc/org/apache/kafka/clients/consumer/ConsumerConfig.html#AUTO_OFFSET_RESET_CONFIG

In this code here:

https://github.com/apache/kafka/blob/31bb4c368dd8723ac90a89c5b255506fba43c633/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L477-L479

We can see that it first is checking the strategy for the reset, but then sending the given sentinel values (`-1` or `-2`) to indicate earliest or latest offsets to the server.  Granted, this is a separate codepath.

Finally, checking the reset offsets command of the kafka-consumer-groups tool, it fetches the latest offsets to set them manually from the command / client doing the call (that is, it's not done in the consumer).

https://github.com/apache/kafka/blob/c3475081c5e8228e9bd3a45022a93d61e542f72e/core/src/main/scala/kafka/admin/ConsumerGroupCommand.scala#L795-L818

In summary, this may not mirror the behavior of the Java client, but I believe the value is high enough and risk is low enough that it's worth implementing with good documentation.  The caveats you mention may be large, especially in low-memory or high-multitenancy environments where events may not stay in the pagecache very long.  

Other relevant code (for reference):

https://github.com/apache/kafka/blob/31bb4c368dd8723ac90a89c5b255506fba43c633/clients/src/main/java/org/apache/kafka/common/requests/ListOffsetsRequest.java#L40-L42
https://github.com/apache/kafka/blob/31bb4c368dd8723ac90a89c5b255506fba43c633/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java

PR #666, opened 2021-05-14
Document default value for ReaderConfig.MaxWait

jan-carreras:
**What**: Document the default value for `ReaderConfig.MaxWait`
**Why**: Makes it clearer — the reader of the configuration does not need to look into the implementation to know the default
**How**: Add a comment on top of the MaxWait definition

Source where the default is being set:
https://github.com/segmentio/kafka-go/blob/master/reader.go#L629 :

```
	if config.MaxWait == 0 {
		config.MaxWait = 10 * time.Second
	}
```

Disclaimer: Not used to contribute to OS projects .
Thanks: For the time dedicated creating/maintaining the project

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -381,6 +381,8 @@ type ReaderConfig struct {
 
 	// Maximum amount of time to wait for new data to come when fetching batches
 	// of messages from kafka.
+	//
+	// Default: 10s
 	MaxWait time.Duration
 
 	// ReadLagInterval sets the frequency at which the reader lag is updated.


jan-carreras:
Pipeline failed with:

```
ok  	github.com/segmentio/kafka-go	54.020s	coverage: 77.7% of statements
--- FAIL: TestCompressedMessages (14.89s)
    --- FAIL: TestCompressedMessages/zstd (9.70s)
        compress_test.go:152: error sending batch 3, reason: context deadline exceeded
FAIL
coverage: 33.3% of statements
```

It is unrelated to the change. How should I proceed?

achille-roussel (approved):
Thanks for the contribution!

Issue #672, opened 2021-05-24
kafka.(*Client).CreateTopics never returns when kafka responds with an error - includes possible fixes


omaskery:
**Describe the bug**
If you pass an invalid `kafka.CreateTopicsRequest` to `kafka.(*Client).CreateTopics` the function never returns, even though internally the library receives a valid response from the broker containing a valid error message.

I have debugged the issue and think I know the cause. See end of this issue for details and possible solutions.
(Caveat: I am not familiar with this library's innards, nor Kafka's protocol).

**Kafka Version**
This docker image: https://hub.docker.com/r/bitnami/kafka/
Seems to be version 2.8.0 of Kafka according to its docs.

**To Reproduce**

If you attempt to create an invalid topic via `kafka.(*Client).CreateTopics` as shown below, the function will never return:

```go
func example(client *kafka.Client) error {
	illegalPartitionCount := 0
	_, _ := client.CreateTopics(ctx, &kafka.CreateTopicsRequest{
		Topics: []kafka.TopicConfig{
			{
				Topic:             "some.topic.name",
				NumPartitions:     illegalPartitionCount,
				ReplicationFactor: 1,
			},
		},
	})
	panic("the code will never get here")
}
```

**Expected behavior**
I would expect to receive an error value from `CreateTopics` containing the error from the broker describing why the request was invalid.

**Additional context**
I have looked through the code and I believe the issue is here due to a `defer` in the `kafka.(*connPool).roundTrip` function. There is special handling for `CreateTopics` that is designed to force a metadata refresh before returning to the caller, in order to ensure the topic has been created - since it is an asynchronous operation. Unfortunately, in the event that the request was invalid this `defer` causes it to infinitely poll for a topic that will never exist:

* https://github.com/segmentio/kafka-go/blob/master/transport.go#L343-L347 <- the defer in question
* https://github.com/segmentio/kafka-go/blob/master/transport.go#L366 <- where the defer is "configured" to expect the newly created topic
* https://github.com/segmentio/kafka-go/blob/master/transport.go#L390 <- perhaps where the fix would go? if the request was to create topics, and it succeeded, then run the metadata polling explicitly instead of a defer?
* https://github.com/segmentio/kafka-go/blob/master/transport.go#L398 <- the function that never returns

Would something like this work?

```diff
 func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error) {
 	// This first select should never block after the first metadata response
 	// that would mark the pool as `ready`.
[...]

PR #673, opened 2021-05-29
Only refresh metadata on success (#672)

omaskery:
Whilst I've made the changes that were agreed in the [GitHub issue](https://github.com/segmentio/kafka-go/issues/672), I haven't done any testing. Please advise on any testing you'd like me to do, above and beyond the unit tests that I assume will be run automatically on the PR after I've submitted it.

Note: I will endeavor to test that this change _actually fixes_ the issue on Monday when I'm back at work (and have access to the codebase where the issue presents).

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -339,8 +339,6 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
 		return nil, ctx.Err()
 	}
 
-	var expectTopics []string
-
 	state := p.grabState()
 	var response promise
 
@@ -354,14 +352,6 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
 			return nil, state.err
 		}
 		return filterMetadataResponse(m, state.metadata), nil
-
-	case *createtopics.Request:
-		// Force an update of the metadata when adding topics,
-		// otherwise the cached state would get out of sync.
-		expectTopics = make([]string, len(m.Topics))
-		for i := range m.Topics {
-			expectTopics[i] = m.Topics[i].Name
-		}
 
 	case protocol.Splitter:
 		// Messages that implement the Splitter interface trigger the creation of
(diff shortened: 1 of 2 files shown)


achille-roussel:
Thanks for opening the pull request @omaskery!

Unless it turns out to be too much efforts (these code paths are well exercised), having a test would be a great addition. We'd want to be able to trigger an error, which could be achieved with a construct like this maybe?

```go
transport := &kafka.Transport{
    Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
        return nil, errSomethingIsBroken
    },
}
```

Using this transport in the test will cause all requests to fail, then the test would perform a `kafka.(*Client).CreateTopics` and verify that the method call returns an error.

omaskery:
Hi, sorry for the delay, I am still alive and will try and add the tests you mention - and to test it in my work. I'm afraid I've been particularly busy recently and haven't gotten back to this work.

omaskery:
Hi, I've finally had a chance to test this fix, and it looks like my initial solution wasn't good enough.

It turns out that when Kafka returns an error that the topic has an invalid configuration, the lower level communication code returns a `createtopics.Response` object which internally contains the error information. The `response.await` itself returns success (error is `nil`), so my naive check for an error and exiting early doesn't work.

Unfortunately the response type is `Response` which only has the method `ApiKey`, no way of inspecting it generically for errors, allowing us to exit early.

Without understanding `kafka-go` much better than I already do, I'm afraid that the only proposal I can offer is the following:

If I move the code that checks the request type from being before the request to afterwards, then it can build the correct list of topics to refresh the metadata for whilst excluding topics which it failed to recreate.

I have tested this approach locally and have found it works, so I will push a commit to this branch demonstrating this approach and you can consider whether you think it's a suitable solution :)

omaskery:
I've also had a chance to add unit tests, it was a bit more involved than you and I had hoped, but I managed to find a solution that detected the issue in the original implementation and passes with the new implementation.

I don't know to what degree this test is a total hack, as I say I'm completely unfamiliar with this code-base, so I just kind of funbled through until I had a working test. Let me know what you think :)

achille-roussel (approved):
Thanks for the contribution @omaskery, the test looks good, I'll go ahead and merge 👍

PR #678, opened 2021-06-05
Explicitly document defaults to RequiredAcks and Async

kishaningithub:
Background
Was wondering what the default value of RequiredAcks was and was looking across writer.go and writer_test.go files. 
This PR is to help people like me know the same without wondering if there is an override of the same, down the call hierarchy


achille-roussel (approved):
Thanks for the contribution!

PR #695, opened 2021-07-03
add unit tests for protocol/

kenju:
Document: https://kafka.apache.org/protocol.html

This PR includes the following commits:

- https://github.com/segmentio/kafka-go/commit/b4e6be33b704cd8ac268f30adfa2600ad834130a: protocol/alterconfigs
- https://github.com/segmentio/kafka-go/pull/695/commits/c859f72577b5e59db9345335bc2cb46c09b1de6a: protocol/createpartitions
- https://github.com/segmentio/kafka-go/pull/695/commits/4fac6aa9e89f3ffb9d8a8b40ac4b5c0209bdd3c1: protocol/deletetopics
- https://github.com/segmentio/kafka-go/pull/695/commits/2c3018278fbf51aca622d0e35b6cd1c9c974098d: protocol/electleaders

### DIFF
--- a/protocol/alterconfigs/alterconfigs_test.go
+++ b/protocol/alterconfigs/alterconfigs_test.go
@@ -17,7 +17,7 @@ func TestAlterConfigsRequest(t *testing.T) {
 		ValidateOnly: true,
 		Resources: []alterconfigs.RequestResources{
 			{
-				ResourceType: 0,
+				ResourceType: 1,
 				ResourceName: "foo",
 				Configs: []alterconfigs.RequestConfig{
 					{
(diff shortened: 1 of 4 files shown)


protocol/alterconfigs/alterconfigs_test.go
```diff
@@ -0,0 +1,73 @@
...
+	v0 = 0
+	v1 = 1
+)
+
+func TestAlterConfigsRequest(t *testing.T) {
+	prototest.TestRequest(t, v0, &alterconfigs.Request{
+		ValidateOnly: true,
+		Resources: []alterconfigs.RequestResources{
+			{
+				ResourceType: 0,
```
achille-roussel, line 20: How about using non-zero values for integer fields? If there was a bug in the encoding or decoding logic it would likely result in the integers having their value, which we can't differentiate if the test expects zero as well.

kenju: @achille-roussel Yes good point :+1: I agree with your idea here.

:heavy_check_mark: https://github.com/segmentio/kafka-go/pull/695/commits/ac63e80980f8fa525d9f92eb3655d1897e19c569

achille-roussel (approved):
Thanks for the contribution!

Issue #621, opened 2021-03-30
Reading topic partitions stopped at specific offsets


OhohLeo:
We use kafka 2.6 & segmentio v0.4.12 (but we reproduce this issue on master branch too).

We are consuming a topic having 10 partitions with a simple program called avro-reader : 
```
GROUP           TOPIC                        PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG            
avro-reader     qv3sync.uuids.to.ids.changes 0          19820831        19820856        25             
avro-reader     qv3sync.uuids.to.ids.changes 1          5719209         19824901        14105692       
avro-reader     qv3sync.uuids.to.ids.changes 2          5735913         19825060        14089147       
avro-reader     qv3sync.uuids.to.ids.changes 3          19826131        19826170        39             
avro-reader     qv3sync.uuids.to.ids.changes 4          8141344         19821910        11680566       
avro-reader     qv3sync.uuids.to.ids.changes 5          8141618         19831628        11690010       
avro-reader     qv3sync.uuids.to.ids.changes 6          19830374        19830407        33             
avro-reader     qv3sync.uuids.to.ids.changes 7          5723818         19824549        14100731       
avro-reader     qv3sync.uuids.to.ids.changes 8          8926384         19824663        10898279       
avro-reader     qv3sync.uuids.to.ids.changes 9          4286435         19828306        15541871       
```

The main issue is that partitions 0, 3 & 6 are fully synced but other partitions stopped to be read.

We use kafkacat to see what is happening and we observe that there are some offset jumps but kafkacat can manage to read the next offset. 

Here are the previous and next message around the offset that where segmentio could not decode : 
```
# Partition 1 offset 5719209
1606765019807 qv3sync.uuids.to.ids.changes(1/5719208) 905bd1cb-8499-4092-b5ef-c38d07de2cd9 => {"instance": "QOS", "type": "Event", "id": 257462080}
1606765019821 qv3sync.uuids.to.ids.changes(1/5719224) 58efef04-e21c-4768-9902-e4381ee38189 => {"instance": "QOS", "type": "Event", "id": 257462551}

# Partition 2 offset 5735913
1606765015761 qv3sync.uuids.to.ids.changes(2/5735912) a05d68dc-712f-4a81-96bf-587cf00a5060 => {"instance": "QOS", "type": "Event", "id": 257376813}
[...]

PR #654, opened 2021-05-04
Handle message sets with zero messages

Nilix007:
This is a hotfix for #653.

PR #709, opened 2021-07-16
advance batch offset where empty message resides at the end

leedavis81:
We've seen issues in being able to advance grouped consumers where the last record in a batch has no bytes. More details on this issue can be seen here https://github.com/segmentio/kafka-go/issues/621

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -243,16 +243,6 @@ func (batch *Batch) readMessage(
 			// consumed or a batch whose connection is in an error state.
 			batch.err = dontExpectEOF(err)
 		case batch.msgs.remaining() == 0:
-
-			// our batch is reporting as not empty despite there being no more messages remaining.
-			// occasionally you could receive a batch with an empty message at the end.
-			// This will  attempt to be discarded, but as the conn expects one more read
-			// the messageSetReader is never emptied. In this scenario we should advance the
-			// batch offset and move on
-			if !batch.msgs.empty {
-				batch.offset++
-			}
-
 			// Because we use the adjusted deadline we could end up returning
 			// before the actual deadline occurred. This is necessary otherwise
 			// timing out the connection for real could end up leaving it in an
@@ -262,6 +252,17 @@ func (batch *Batch) readMessage(
 			// read deadline management.
 			err = checkTimeoutErr(batch.deadline)
 			batch.err = err
+
+			// our batch is reporting as not empty despite there being no more messages remaining.
+			// occasionally you could receive a batch with an empty message at the end.
+			// This will attempt to be discarded, but as the conn expects one more read
+			// the messageSetReader is never emptied. In the scenario where our batch is reporting as non-empty
+			// but our batch offset equals the connection offset we should advance the batch.
+			if !batch.msgs.empty &&
+				batch.offset == batch.conn.offset &&
+				batch.err == io.EOF {
+				batch.offset++
+			}
 		}
 	default:
 		// Since io.EOF is used by the batch to indicate that there is are
--- a/batch.go
+++ b/batch.go
@@ -264,6 +264,26 @@ func (batch *Batch) readMessage(
 				batch.offset++
 			}
 		}
+	// in an instance where the last message of a batch doesn't exist we need to advance past it to
+	// prevent falling into a state where the consumer cannot advance. errEndOfBatch errors are returned
+	// under a scenario where no bytes remain on a reader but a read attempt is being made because our
+	// conn struct is expecting an additional message (i.e it's offset is set to the deleted / end of batch offset)
+	case errEndOfBatch:
+		err = batch.msgs.discard()
+		switch {
+		case err != nil:
+			batch.err = dontExpectEOF(err)
+		case batch.msgs.remaining() == 0:
+			err = checkTimeoutErr(batch.deadline)
+			batch.err = err
+			// Where our checkTimeoutErr results in an io.EOF and our batch offset is equal to our connection offset
+			// we advance the batch to move past the deleted message.
+			if !batch.msgs.empty &&
+				batch.offset == batch.conn.offset &&
+				batch.err == io.EOF {
+				batch.offset++
+			}
+		}
 	default:
 		// Since io.EOF is used by the batch to indicate that there is are
 		// no more messages to consume, it is crucial that any io.EOF errors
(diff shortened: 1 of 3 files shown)


leedavis81:
upon testing this is appears that it occasionally skips over records it shouldn't, prematurely advancing the offset. I'm fairly certain the issue resides around there being an empty byte message at the end of a batch, but i'm having trouble isolating that.

keen to get any feedback on how we can resolve this.

nlsun:
Hi @leedavis81 do we have a way of testing that this issue is fixed in a unit test?

Also, if I understand correctly, the issue isn't completely solved? I'll need to look into the issue as well, so I don't have any suggestions at this time on how to fix/debug yet

leedavis81:
Yep, on testing it appears to be advancing under scenarios where it shouldn't. I'm struggling to see where the best place is to hook into the code, but i'm fairly certain the issue surrounds an empty byte message at the end of a batch. I'll try to get a test down as soon as I have time.

achille-roussel:
Is this related to https://github.com/segmentio/kafka-go/pull/654 ?

leedavis81:
No, I tried that patch and it didn't solve the issue of not being able to advance past our problem batch

leedavis81:
Ok, so i've added in a few additional conditions to stop the batch advancing in other genuine errShortRead situations. They issue here is that `batch.msgs.empty` is reporting false where connection highwatermark is always +1 on the reader offset. The reader continues to process the last message of the batch but can never do so successfully because it has no bytes, meaning the reader is never reset as empty under this condition in conn.go...

```		if highWaterMark == offset {
			msgs = &messageSetReader{empty: true}
		} else {
			msgs, err = newMessageSetReader(&c.rbuf, remain)
		}
```

This fix checks for a condition where the batch is reporting as non-empty however the batch offset and the connection offset are equal. Additionally to ensure we don't fall into this condition under a genuine timeout a check for `io.EOF` is added.

I've tested this manually on topics which present the issue and we managed to consume everything correction without any message loss. It's possible that this issue could be resolved at the reader level (where no bytes remaining) but I found it difficult to define that condition accurately. This implementation would also solve this issue across both v1/v2 kafka message types (though I can only confirm the issue is present in v2).

leedavis81:
After further tests it looks like this change does solve the problem, but introduces a `OffsetOutOfRange` error where you're caught-up and tailing a partition. Maybe a better approach would be to try and detect this at the reader level and try to increase the offset count there.

nlsun:
@leedavis81 it sounds like you guys have a scenario on hand that reproduces the issue, have you had a chance to try and identify the conditions required to minimally reproduce the issue?

leedavis81:
Hi @nlsun, thanks for following up.

So to recreate the issue we need to have a zero byte record at the end of a batch. So for example, let's say i'm using a grouped consumer and the broker has me marked at offset `1270848` on partition `0` (which happens to be a zero byte message).

A batch read will occur requesting offset `1270848` and the broker returns a batch, let's say from `1270842`. So we have `1270842, 1270843, 1270844, 1270845, 1270846, 1270847 and 1270848`

when our `Reader` is initialised the `Seek()` call on the underlying `*Conn` object is set with an offset of `1270848`

Our reader then begins to parse the header and starts reading messages. A `messageCount` value is taken from the header which correctly states `6` messages. 

The reader loops through the messages, reading all 6 up to `1270847`. At this point despite `remain` being 0 the reader continues to try to read the following message as it's not met the conn offset yet of `1270848`. This can be seen with the following line:

`for batch.conn != nil && offset < batch.conn.offset { // do another read`

an attempt to read another message is made with both a message count of 0 and no remaining bytes in the reader (`*bufio.Reader`). under `*messageSetReaderV2.readMessage` this falls into the condition of trying to read the batch header (which make sense as it appears like a fresh batch)..

```
		if err = r.readHeader(part); err != nil {
			return
		}
```
this then produces an `errShortRead` and we fall into the `case batch.msgs.remaining() == 0:` scenario on `*Batch.readMessage()`. An `io.EOF` error is set from the checkTimeoutErr() function and the for loop described above is exited with a `break`. That err is then passed up and the batch is closed..

```
		if msg, err = batch.ReadMessage(); err != nil {
			batch.Close()
			break
		}

the offset returned from `(r *reader) read` is set as the batch offset + 1 `offset = msg.Offset + 1` (which in this case is `1270847+1`) and that's then returned and subsequently used to be marked on the broker. The reader is then initialised with the offset pulled from the broker, and the cycle repeats, meaning we never progress beyond `1270848`.

As you can see the issue spans across both what the conn object holds, and the message reader attempting a read with 0 bytes remaining. Given the way the library is layered it's challenging to setup a test to reflect this issue. I'm also not sure where the best place is to hook in a fix. I think the right approach is to somehow skip over this message, increasing the `batch.offset` by one (rather than trying to decrease the `conn.offset`) and is reflected in my PR. But it would be cleaner if we could do this at the message level (where we know no bytes remain).

The PR will work, but when the consumer is caught up and reads rightly start producing `io.EOF`'s we're incorrectly advancing the batch offset which begins to generate `OffsetOutOfRange` errors from our `conn` struct. To clean this up I could instead introduce another error type. Something like ...

`var ErrNoMsgsRemain = errors.New("no remaining messges in batch")`

which I can return from the message structs in the case of ...

`r.messageCount == 0 && r.remain == 0 && r.readHeader() produces an errShortRead`

New batches will always have something in their reader buffer, so I think this describes the scenario quite well of running into a zero byte message at the end of a batch. 

Keen to hear your thoughts

leedavis81:
I've updated the PR to reflect what I suggested above. This passes an error in the scenario of trying to read with an empty buffer back up to the batcher, which will then advance the offset forward. It's not the cleanest code but it does solve the issue without incorrectly advancing the connection offset in genuine io.EOF scenarios

leedavis81:
Hi @nlsun, do you have any thoughts on the above? This PR does resolve the issue we're seeing however happy to take another approach.

Lee

nlsun:
@leedavis81 I haven't had much time to look yet but I've been looking into reproducing the issue. Thanks for the detailed report, I think there should be enough information, now I just need to isolate the issue

nlsun:
The ideas in this PR have been incorporated into https://github.com/segmentio/kafka-go/pull/813

PR #715, opened 2021-07-21
Do a final commit on end consumer group generation for immediate commits

### LINKED ISSUE #711: Question: starting offsets - do they work with consumer groups?
One of my automated FMEA test failed. The test simulates a deletion of a pod and verifies that it re-connects to Kafka and correctly continues to process messages. It lost around 50% of the total messages - this is, not to mince my words, catastrophic. 

The diagnostics hinted to me that when the consumer group re-connected, it started at the latest offset. 

I recently (maybe coincidentally, maybe causally) added a configuration items to the `ReaderConfig` as follows:

```golang
kafka.ReaderConfig{
		Brokers:     x.Brokers,
		Dialer:      dialer,
		GroupID:     groupName,
		Topic:       x.Topic,
		StartOffset: startOffset,
	}
```

My expectation would be that, on first connect it'd use the `StartOffset`, but on re-connect, the consumer group would just continue from the last committed offset. 

Am I used this incorrectly?

stevevls:
This is a WIP and isn't ready to commit until I put a unit test together. 😄 

This PR adds a final commit upon exiting the consumer group when using the synchronous commit mode.  Otherwise, it's possible that we lose commits on generation end (or other events that initiate the end of a generation such as closing the reader).

Fixes #711

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -212,6 +212,25 @@ func (r *Reader) commitLoopImmediate(ctx context.Context, gen *Generation) {
 	for {
 		select {
 		case <-ctx.Done():
+			// drain the commit channel and prepare a single, final commit.
+			// the commit will combine any outstanding requests and the result
+			// will be sent back to all the callers of CommitMessages so that
+			// they can return.
+			var errchs []chan<- error
+			for hasCommits := true; hasCommits; {
+				select {
+				case req := <-r.commits:
+					offsets.merge(req.commits)
+					errchs = append(errchs, req.errch)
+				default:
+					hasCommits = false
+				}
+			}
+			err := r.commitOffsetsWithRetry(gen, offsets, defaultCommitRetries)
+			for _, errch := range errchs {
+				// NOTE : this will be a buffered channel and will not block.
+				errch <- err
+			}
 			return
 
 		case req := <-r.commits:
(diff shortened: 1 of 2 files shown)


rhansen2:
Possible related to https://github.com/segmentio/kafka-go/issues/772

stevevls:
Good call.  This definitely could be related since they are not setting the `CommitInterval`.  I think most users (myself included) don't realize that the commit loop has to be explicitly enabled.

rhansen2:
From https://docs.confluent.io/platform/current/clients/consumer.html

> Offset commit failures are merely annoying if the following commits succeed since they won’t actually result in duplicate reads. However, if the last commit fails before a rebalance occurs or before the consumer is shut down, then offsets will be reset to the last commit and you will likely see duplicates. A common pattern is therefore to combine async commits in the poll loop with sync commits on rebalances or shut down. Committing on close is straightforward, but you need a way to hook into rebalances.

derekbassett:
Bump

achille-roussel (approved):
🚢  it!

stevevls:
Thanks for landing this team! 🙌   I'm sorry I didn't see the mention for whether I wanted to keep working on it.

PR #716, opened 2021-07-22
Preserve message order across batches.

### LINKED ISSUE #686: v0.4 Writer does not preserve order of messages
**Describe the bug**
When writing multiple messages using kafka.Writer, multiple batches will be created, and those batches will be processed asynchronously, causing messages to arrive Kafka in random order. e.g. If I write message 1, 2, 3 in one WriteMessages call, then we may create three batches for message 1, 2, 3 and they will reach Kafka in the order of 2, 3, 1.

**Kafka Version**
2.6.0

**To Reproduce**
Start Kafka locally using default configs, create a topic "test-topic", then use this code snippet:
```
package bugreport_test

import (
	"context"
	"math/rand"
	"strconv"
	"testing"
	"time"

	"github.com/golang/glog"
)

type balancer int

func (b balancer) Balance(msg kafka.Message, partitions ...int) (partition int) {
	return 0
}

func NewMessage(size int, key, topic string) kafka.Message {
	return kafka.Message{
		Key:     []byte(key),
		Value:   messageValue(size),
		Headers: []kafka.Header{},
		Time:    time.Now(),
	}
}

func messageValue(size int) []byte {
	val := make([]byte, size)
	for i := 0; i < size; i++ {
		val[i] = byte(random(0, 9))
	}
	return val
}

func random(min int, max int) int {
	return rand.Intn(max-min) + min
}

func TestWriter(t *testing.T) {
	wr := kafka.NewWriter(kafka.WriterConfig{
		Brokers:      []string{"localhost:9092"},
		Topic:        "test-topic",
		Dialer:       kafka.DefaultDialer,
		BatchSize:    1000,
		BatchTimeout: 500 * time.Millisecond,
		Balancer:     balancer(0),
	})
	var msgs = []kafka.Message{
		NewMessage(950000, "1", "test-topic"),
		NewMessage(950000, "2", "test-topic"),
		NewMessage(950000, "3", "test-topic"),
		NewMessage(950000, "4", "test-topic"),
		NewMessage(950000, "5", "test-topic"),
		NewMessage(950000, "6", "test-topic"),
		NewMessage(950000, "7", "test-topic"),
		NewMessage(950000, "8", "test-topic"),
		NewMessage(950000, "9", "test-topic"),
		NewMessage(950000, "10", "test-topic"),
	}
	rdr := kafka.NewReader(kafka.ReaderConfig{
		Brokers:     []string{"localhost:9092"},
		Topic:       "test-topic",
		Dialer:      kafka.DefaultDialer,
		GroupID:     strconv.FormatInt(time.Now().Unix(), 10),
		StartOffset: kafka.LastOffset,
	})
	ctx, _ := context.WithTimeout(context.Background(), 100*time.Second)
	go func() {
		for {
			msg, err := rdr.ReadMessage(ctx)
			if err != nil {
				glog.Errorf("err read msg: %s", err.Error())
			} else {
				glog.Infof("receive msg: key: %s, offset: %d", string(msg.Key), msg.Offset)
			}
		}
	}()
	time.Sleep(3 * time.Second)
	err := wr.WriteMessages(ctx, msgs...)
	if err != nil {
		glog.Errorf("err write msg: %s", err.Error())
	} else {
		glog.Info("write complete")
	}
	time.Sleep(10 * time.Second)
}
```
Run the test using ```go test -v './...' --logtostderr --v=1```, you will see something like:

I0622 14:36:34.674736 kafka-go_test.go:69] receive msg: key: 2, offset: 36
I0622 14:36:34.780102 kafka-go_test.go:69] receive msg: key: 7, offset: 37
I0622 14:36:34.885081 kafka-go_test.go:69] receive msg: key: 3, offset: 38
I0622 14:36:34.990090 kafka-go_test.go:69] receive msg: key: 5, offset: 39
I0622 14:36:35.095683 kafka-go_test.go:69] receive msg: key: 8, offset: 40
I0622 14:36:35.200772 kafka-go_test.go:69] receive msg: key: 6, offset: 41
I0622 14:36:35.303842 kafka-go_test.go:69] receive msg: key: 9, offset: 42
I0622 14:36:35.409054 kafka-go_test.go:69] receive msg: key: 4, offset: 43
I0622 14:36:35.530030 kafka-go_test.go:69] receive msg: key: 1, offset: 44
I0622 14:36:35.634247 kafka-go_test.go:69] receive msg: key: 10, offset: 45

**Expected behavior**
Since we call WriteMessages() in the order of 1 to 10, we expect to see
I0622 14:36:34.674736 kafka-go_test.go:69] receive msg: key: 1, offset: 36
I0622 14:36:34.780102 kafka-go_test.go:69] receive msg: key: 2, offset: 37
I0622 14:36:34.885081 kafka-go_test.go:69] receive msg: key: 3, offset: 38
I0622 14:36:34.990090 kafka-go_test.go:69] receive msg: key: 4, offset: 39
I0622 14:36:35.095683 kafka-go_test.go:69] receive msg: key: 5, offset: 40
I0622 14:36:35.200772 kafka-go_test.go:69] receive msg: key: 6, offset: 41
I0622 14:36:35.303842 kafka-go_test.go:69] receive msg: key: 7, offset: 42
I0622 14:36:35.409054 kafka-go_test.go:69] receive msg: key: 8, offset: 43
I0622 14:36:35.530030 kafka-go_test.go:69] receive msg: key: 9, offset: 44
I0622 14:36:35.634247 kafka-go_test.go:69] receive msg: key: 10, offset: 45

**Additional context**
We are using kafka-go v0.4.16

rhansen2:
Attempt to preserve message order across batches.

Fixes #686

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -884,7 +884,7 @@ func (b *batchQueue) Put(batch *writeBatch) bool {
 	return true
 }
 
-func (b *batchQueue) Get(batches []*writeBatch) []*writeBatch {
+func (b *batchQueue) Get() *writeBatch {
 	b.cond.L.Lock()
 	defer b.cond.L.Unlock()
 
@@ -892,21 +892,14 @@ func (b *batchQueue) Get(batches []*writeBatch) []*writeBatch {
 		b.cond.Wait()
 	}
 
-	limit := cap(batches) - len(batches)
-	if limit == 0 {
-		// ensure some progress is made at the cost of a reallocation.
-		limit = 1
-	}
-	queueMax := len(b.queue)
-	var i int
-	for i = 0; i < limit && i < queueMax; i++ {
-		batches = append(batches, b.queue[i])
-		b.queue[i] = nil
+	if len(b.queue) == 0 {
+		return nil
 	}
 
-	b.queue = b.queue[i:]
+	var batch *writeBatch
+	batch, b.queue = b.queue[0], b.queue[1:]
 
-	return batches
+	return batch
 }
 
 func (b *batchQueue) Close() {
@@ -959,22 +952,18 @@ func newPartitionWriter(w *Writer, key topicPartition) *partitionWriter {
 }
 
 func (ptw *partitionWriter) writeBatches() {
-	batches := make([]*writeBatch, 0, 10)
-
 	for {
-		batches = ptw.queue.Get(batches)
+		batch := ptw.queue.Get()
 
-		// The only time we can return 0 batches is when the queue is closed
+		// The only time we can return nil is when the queue is closed
 		// and empty. If the queue is closed that means
 		// the Writer is closed so once we're here it's time to exit.
-		if len(batches) == 0 {
+		if batch == nil {
 			return
 		}
 
-		for _, batch := range batches {
-			ptw.writeBatch(batch)
-		}
-		batches = batches[:0]
+		ptw.writeBatch(batch)
+
 	}
 }
 
--- a/writer.go
+++ b/writer.go
@@ -969,15 +969,16 @@ func (ptw *ptWriter) writeMessages(msgs []Message, indexes []int32) map[*writeBa
 	ptw.group.Add(1)
 	defer ptw.group.Done()
 
+	ptw.mutex.Lock()
+	defer ptw.mutex.Unlock()
+
+	batchSize := ptw.w.batchSize()
+	batchBytes := ptw.w.batchBytes()
+
 	var batches map[*writeBatch][]int32
 	if !ptw.w.Async {
 		batches = make(map[*writeBatch][]int32, 1)
 	}
-	batchSize := ptw.w.batchSize()
-	batchBytes := ptw.w.batchBytes()
-
-	ptw.mutex.Lock()
-	defer ptw.mutex.Unlock()
 
 	for _, i := range indexes {
 	assignMessage:
@@ -1006,6 +1007,7 @@ func (ptw *ptWriter) writeMessages(msgs []Message, indexes []int32) map[*writeBa
 	return batches
 }
 
+// ptw.w can be accessed here because this is called with the lock ptw.mutex already held.
 func (ptw *ptWriter) newWriteBatch() *writeBatch {
 	batch := newWriteBatch(time.Now(), ptw.w.batchTimeout())
 	ptw.group.Add(1)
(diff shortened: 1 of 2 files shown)


writer.go
```diff
@@ -189,9 +189,9 @@ type Writer struct {
 	closed uint32
 	group  sync.WaitGroup
 
-	// Manages the current batch being aggregated on the writer.
+	// Manages the curret set of partition-topic writers.
```
MaerF0x0, line 192:
```suggestion
	// Manages the current set of partition-topic writers.
```

writer.go
```diff
@@ -992,6 +859,301 @@ func (w *Writer) chooseTopic(msg Message) (string, error) {
...
+	ptw.mutex.Lock()
+	defer ptw.mutex.Unlock()
+
+	if ptw.currBatch != nil {
+		batch := ptw.currBatch
+		ptw.queue.Put(batch)
+		ptw.currBatch = nil
+		batch.trigger()
+	}
+	ptw.w = nil
```
achille-roussel, line 1145: Could we make `ptw.w` an immutable property so we don't need synchronization to access the field?
```suggestion
```
This means we wouldn't need the `writer` method either.

rhansen2: Yes, I think that's preferred. I was worried about leaving circular references around but those can be broken when closing the writer.

writer.go
```diff
@@ -992,6 +859,301 @@ func (w *Writer) chooseTopic(msg Message) (string, error) {
...
+		queue: make([]*writeBatch, 0, initialSize),
+		cond: sync.Cond{
+			L: &sync.Mutex{},
+		},
+	}
+}
+
+// ptWriter is a writer for a topic-partion pair. It maintains messaging order
+// across batches of messages.
+type ptWriter struct {
```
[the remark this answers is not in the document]
rhansen2: I was torn about including topic in the name since these writers are tied to a particular topic-partition pair but I prefer `partitionWriter` as well.

writer.go
```diff
@@ -992,6 +859,301 @@ func (w *Writer) chooseTopic(msg Message) (string, error) {
...
+
+func (ptw *ptWriter) writeBatches() {
+	ptw.group.Add(1)
+	defer ptw.group.Done()
+
+	// Should we make this configurable?
+	batches := make([]*writeBatch, 0, 10)
+
+	for {
+		batches = ptw.queue.Get(batches)
```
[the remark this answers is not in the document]
rhansen2: My thought was if we grab everything in the queue at once, we'd have to contend for the lock less often but I'm not opposed to changing this to grab a single item at a time.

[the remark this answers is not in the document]
achille-roussel: I doubt we would see any contention in this case, we are pulling batches of messages, so the synchronization gets amortized already. The synchronization time is also minimal in comparison with what is happening in `writeBatch`.

I'd be supportive of keeping the code simpler unless we are able to measure that it is an issue.

writer.go
achille-roussel: I believe the Go GC also knows how to deal with circular references 👍  we can let it take care of it.

achille-roussel (approved):
This is looking great 👍 

I wonder if we could have someone from #686 give it a try to make sure it solves the issue before merging.

Thanks for the contribution!

PR #700, opened 2021-07-09
WIP: Allow Writer to auto-create topics (Fixes #683)

dtjm:
Addresses #683 by enabling auto-topic creation from `kafka.Writer.WriteMessages` 

Auto-topic creation is still asynchronous, so you are still likely to run into a "in the middle of a leader election" error if the topic has not yet been created, and you will need to keep retrying until Kafka creates the topic. 

This is still a WIP, just putting the idea out there for initial feedback.

PR #719, opened 2021-07-30
Refactor Writer tests 

dtjm:
This is split out as a prerequisite to #700 since that PR is breaking the Writer tests, so I'm trying to isolate the orthogonal changes as much as possible.

- Add test coverage for `Writer{}` literal instantiation 
- Keep test coverage for deprecated `NewWriter()` constructor since it's still in use


achille-roussel (commented):
@rhansen2 said we covered this in a different PR :)

PR #722, opened 2021-08-05
Make Writer tests more resilient

dtjm:
- Increase some timeouts
- When creating a test topic, block until it's ready
- Wrap some errors to help debug tests 

Tests do seem to run more reliably (ran with `-count=20` and they don't fail as often now), but the downside is that they take longer to run (~2-3 min instead of ~1m) 

I broke this out of #719 because I kept getting intermittent failures there, and I'm trying to isolate them away from the refactoring changes.

### DIFF
--- a/compress/compress_test.go
+++ b/compress/compress_test.go
@@ -16,7 +16,7 @@ import (
 	"text/tabwriter"
 	"time"
 
-	kafka "github.com/segmentio/kafka-go"
+	"github.com/segmentio/kafka-go"
 	pkg "github.com/segmentio/kafka-go/compress"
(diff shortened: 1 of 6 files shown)


rhansen2:
LGTM

achille-roussel:
Is some of the logic we introduced here a duplicate of what we had in https://github.com/segmentio/kafka-go/blob/master/client_test.go#L36-L68 ?

dtjm:
@achille-roussel yes it looks to be the same intention.  I can refactor to use the existing stuff in my next PR, as I'm still going to be dealing with writer tests.

PR #725, opened 2021-08-19
feat(sasl): enrich authentication context with metadata

mrolla:
Picking up from https://github.com/segmentio/kafka-go/pull/563 and https://github.com/segmentio/kafka-go/pull/598, this PR only does the necessary changes to have additional metadata propagated to the authentication functions so it can be used for authenticating. Currently it's only the address (needed by GSSAPI).

I chose the context approach because it requires the least amount of changes to existing code.

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -282,7 +282,6 @@ func (d *Dialer) connect(ctx context.Context, network, address string, connCfg C
 		metadata := &sasl.Metadata{
 			BrokerAddress: address,
 		}
-
 		if err := d.authenticateSASL(sasl.WithMetadata(ctx, metadata), conn); err != nil {
 			_ = conn.Close()
 			return nil, err
--- a/dialer.go
+++ b/dialer.go
@@ -280,7 +280,7 @@ func (d *Dialer) connect(ctx context.Context, network, address string, connCfg C
 
 	if d.SASLMechanism != nil {
 		metadata := &sasl.Metadata{
-			BrokerAddress: address,
+			Host: address,
 		}
 		if err := d.authenticateSASL(sasl.WithMetadata(ctx, metadata), conn); err != nil {
 			_ = conn.Close()
--- a/dialer.go
+++ b/dialer.go
@@ -279,7 +279,11 @@ func (d *Dialer) connect(ctx context.Context, network, address string, connCfg C
 	conn := NewConnWith(c, connCfg)
 
 	if d.SASLMechanism != nil {
-		if err := d.authenticateSASL(ctx, conn); err != nil {
+		metadata := sasl.Metadata{
+			Address: address,
+		}
+
+		if err := d.authenticateSASL(metadata.WithContext(ctx), conn); err != nil {
 			_ = conn.Close()
 			return nil, err
 		}
--- a/dialer.go
+++ b/dialer.go
@@ -280,7 +280,7 @@ func (d *Dialer) connect(ctx context.Context, network, address string, connCfg C
 
 	if d.SASLMechanism != nil {
 		metadata := sasl.Metadata{
-			Address: address,
+			BrokerAddress: address,
 		}
 
 		if err := d.authenticateSASL(metadata.WithContext(ctx), conn); err != nil {
--- a/dialer.go
+++ b/dialer.go
@@ -279,11 +279,11 @@ func (d *Dialer) connect(ctx context.Context, network, address string, connCfg C
 	conn := NewConnWith(c, connCfg)
 
 	if d.SASLMechanism != nil {
-		metadata := sasl.Metadata{
+		metadata := &sasl.Metadata{
 			BrokerAddress: address,
 		}
 
-		if err := d.authenticateSASL(metadata.WithContext(ctx), conn); err != nil {
+		if err := d.authenticateSASL(sasl.WithMetadata(ctx, metadata), conn); err != nil {
 			_ = conn.Close()
 			return nil, err
 		}
--- a/sasl/sasl.go
+++ b/sasl/sasl.go
@@ -47,7 +47,9 @@ type StateMachine interface {
 
 // Metadata contains additional data for performing SASL authentication.
 type Metadata struct {
-	Address string
+	// BrokerAddress is the address of the broker the authentication will be
+	// performed on.
+	BrokerAddress string
 }
 
 // WithContext returns a copy of the context with associated Metadata.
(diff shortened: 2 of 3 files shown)


sasl/sasl.go
```diff
@@ -42,3 +44,21 @@ type StateMachine interface {
 	// value will be true.
 	Next(ctx context.Context, challenge []byte) (done bool, response []byte, err error)
 }
+
+// Metadata contains additional data for performing SASL authentication.
+type Metadata struct {
+	Address string
```
achille-roussel, line 50: Could we document or rename this field? "Address" is a very generic term, it does not do a good job at explaining what value is expected here.

sasl/sasl.go
```diff
@@ -42,3 +44,21 @@ type StateMachine interface {
...
+}
+
+// WithContext returns a copy of the context with associated Metadata.
+func (m *Metadata) WithContext(ctx context.Context) context.Context {
```
achille-roussel, line 54: What made you choose a method over a top-level function here?

sasl/sasl.go
```diff
@@ -2,6 +2,8 @@ package sasl
 
 import "context"
 
+type ctxKey struct{}
```
achille-roussel, line 5: How about renaming to be more explicit what value this key is intended to map to?
```suggestion
type metadataKey struct{}
```

sasl/sasl.go
```diff
@@ -42,3 +44,21 @@ type StateMachine interface {
...
+func (m *Metadata) WithContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, ctxKey{}, m)
+}
+
+// MetadataFromContext retrieves the Metadata from the context.
+func MetadataFromContext(ctx context.Context) *Metadata {
+	if m, ok := ctx.Value(ctxKey{}).(*Metadata); ok {
+		return m
+	}
+	return nil
```
achille-roussel, line 63: This can be simplified:
```suggestion
	m, _ := ctx.Value(ctxKey{}).(*Metadata)
	return m
```

sasl/sasl.go
mrolla: I will both rename and document the field. I think a better name would be `BrokerAddress`

sasl/sasl.go
mrolla: I like the direct approach of embedding itself into a context given a parent (examples of the same approach out in the wild https://github.com/rs/zerolog/blob/master/ctx.go#L28)

sasl/sasl.go
mrolla: I'm fine with renaming this, but while `ctxKey` might sound obscure, `metadataKey` is a little bit ambiguous imho. `ctxKey` makes it clear it's a type used to hold stuff into the context. What do you think?

mrolla:
@achille-roussel thanks for the review! I incorporated your feedback in the PR. The remaining points I think boil down to personal preference, let me know.

sasl/sasl.go
achille-roussel: I guess I was thinking it may be a bit generic, and not leave a lot of room for adding more keys for other values we'd keep in the context.

It's unexported so easy to change, no real concerns about leaving it as-is for now, it was just a suggestion.

sasl/sasl.go
achille-roussel: I think `WithContext` is a bit weird as a method name. For example, `http.(*Request).WithContext` returns a `*http.Request` which embeds the context, `context.WithValue` returns a context which embeds a value, etc...

Here we say "metadata with context" but don't return a metadata embedding a context, we return a context which embeds the metadata that it was called on... effectively inverting the pattern established by the standard library.

In my opinion, the top-level function seems to be the most in line with the way the standard library is constructed:
```go
func WithMetadata(ctx context.Context, m *Metadata) context.Context { ... }
```

sasl/sasl.go
```diff
@@ -42,3 +44,21 @@ type StateMachine interface {
 	// value will be true.
 	Next(ctx context.Context, challenge []byte) (done bool, response []byte, err error)
 }
+
+// Metadata contains additional data for performing SASL authentication.
+type Metadata struct {
+	// BrokerAddress is the address of the broker the authentication will be
+	// performed on.
```
achille-roussel, line 51: Is this address meant to be the logical hostname or the resolved network address (IP+port)?

achille-roussel (commented):
Thanks for following up, I left a couple more questions on the updates.

sasl/sasl.go
mrolla: It's meant to be the logical hostname. Its host part is then used to generate the SPN (Service Principal Name) in the context of the GSSAPI authentication.

mrolla:
@achille-roussel, feedback applied!

dialer.go
```diff
@@ -279,7 +279,10 @@ func (d *Dialer) connect(ctx context.Context, network, address string, connCfg C
 	conn := NewConnWith(c, connCfg)
 
 	if d.SASLMechanism != nil {
-		if err := d.authenticateSASL(ctx, conn); err != nil {
+		metadata := &sasl.Metadata{
+			BrokerAddress: address,
```
achille-roussel, line 283:
```suggestion
			Host: address,
```

sasl/sasl.go
```diff
@@ -42,3 +44,21 @@ type StateMachine interface {
 	// value will be true.
 	Next(ctx context.Context, challenge []byte) (done bool, response []byte, err error)
 }
+
+// Metadata contains additional data for performing SASL authentication.
+type Metadata struct {
+	// BrokerAddress is the address of the broker the authentication will be
```
achille-roussel, line 50:
```suggestion
	// Host is the address of the broker the authentication will be
```

sasl/sasl.go
```diff
@@ -42,3 +44,21 @@ type StateMachine interface {
...
+	// performed on.
+	BrokerAddress string
```
achille-roussel, line 52:
```suggestion
	Host string
```

transport.go
```diff
@@ -1161,7 +1161,10 @@ func (g *connGroup) connect(ctx context.Context, addr net.Addr) (*conn, error) {
 	pc.SetDeadline(time.Time{})
 
 	if g.pool.sasl != nil {
-		if err := authenticateSASL(ctx, pc, g.pool.sasl); err != nil {
+		metadata := &sasl.Metadata{
+			BrokerAddress: netAddr.String(),
```
achille-roussel, line 1165:
```suggestion
			Host: netAddr.String(),
```

mrolla:
Better naming. Thanks @achille-roussel. Sorry, if I let it pending, I thought it was just waiting to be merged.

achille-roussel:
All good! Thanks again for the contribution!

PR #735, opened 2021-09-13
Add Offsetcommit to Client

rhansen2:


achille-roussel (approved):
Thanks for picking this up and adding tests 🙇

PR #737, opened 2021-09-14
Support more of the InitProducerID api.

rhansen2:
Ran into a case testing vs Kafka 2.6.0 where the additional fields were needed.


achille-roussel (approved):
Looking good 👍 thanks for the contribution!

PR #744, opened 2021-09-20
Add Transactionalmessage interface for request routing

rhansen2:


achille-roussel (approved):
Nice!

protocol/protocol.go
```diff
@@ -438,6 +438,14 @@ type GroupMessage interface {
 	Group() string
 }
 
+// TransactionalMessage is an extension of the Message interface implemented by some
+// request types to inform the program that they should be reouted to a transaction
```
erikdw, line 442: typo: reouted  -> routed (I think!)

rhansen2: Sorry, I think I merged just before your comment. I'll get that typo fixed up in another PR. Thanks!

PR #747, opened 2021-09-23
fix: set the offset does not take effect

maweikee:
when I use consumer groups, I want to start reading from the current message with new pod， so I set last offset，It seems that setting LastOffset always does not work， so i tried this。

version :v0.4.17

### DIFF
--- a/consumergroup.go
+++ b/consumergroup.go
@@ -1137,7 +1137,7 @@ func (cg *ConsumerGroup) makeAssignments(assignments map[string][]int32, offsets
 			if ok {
 				offset, ok = partitionOffsets[int(partition)]
 			}
-			if !ok {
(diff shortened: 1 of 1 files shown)


dominicbarnes:
It looks like you are trying to ensure that the `config.StartOffset` is used when no prior offset is found for the that consumer group. The special-case for only `LastOffset` strikes me as unusual, as there are other configurations that I think could make sense as well. (eg: `FirstOffset`)

Before anything else, could you add a test case to cover the new behavior you are attempting to add here? It seems like you'll want to create a topic, write at least 2 messages and then verify that consumption starts with the last message, rather than the first. (in the case of `LastOffset`)

Let me know if I can assist in getting a test case set up.

PR #752, opened 2021-10-01
Fix member metadata decoding and remove deprecated describeGroups API

yolken-segment:
## Description
This change fixes how member metadata is decoded in kafka-go's `DescribeGroups` implementation. Although the [top-level protocol spec](https://kafka.apache.org/protocol#The_Messages_DescribeGroups) just says that this metadata is "bytes", it's actually encoded using the consumer protocol spec described [here](https://github.com/apache/kafka/blob/2.4/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerProtocol.java#L37). 

The latest version of this spec allows consumers to set an `OwnedPartitions` field that may be used for cooperative rebalancing. The current implementation does not handle this extra field, which leads to a "Got non-zero number of bytes remaining ..." errors when decoding the response. Users have reported this error when running `get lags` in `topicctl` against newer Java-based consumers (see https://github.com/segmentio/topicctl/issues/15).

This change also removes the old, lower-level `describeGroups` API. The latter is no longer needed now that we have `DescribeGroups` in the higher-level client.

## Testing
Testing completed successfully via building `topicctl` with the updated `kafka-go` client, verifying that `get lags` works correctly for groups using both newer and older versions of the protocol.

### DIFF
--- a/describegroups.go
+++ b/describegroups.go
@@ -207,11 +207,12 @@ func decodeMemberMetadata(rawMetadata []byte) (DescribeGroupsResponseMemberMetad
 			if fnRemain, fnErr = readString(r, size, &op.Topic); fnErr != nil {
 				return
 			}
-			ps := []int32{}
 
-			if fnRemain, fnErr = readInt32Array(r, remain, &ps); fnErr != nil {
+			ps := []int32{}
+			if fnRemain, fnErr = readInt32Array(r, fnRemain, &ps); fnErr != nil {
 				return
 			}
+
 			for _, p := range ps {
 				op.Partitions = append(op.Partitions, int(p))
 			}
--- a/describegroups.go
+++ b/describegroups.go
@@ -156,7 +156,7 @@ func (c *Client) DescribeGroups(
 	return resp, nil
 }
 
-// readFrom
+// readFrom decodes an owned partition item from the member metadata.
 func (t *DescribeGroupsResponseMemberMetadataOwnedPartition) readFrom(r *bufio.Reader, size int) (remain int, err error) {
 	if remain, err = readString(r, size, &t.Topic); err != nil {
 		return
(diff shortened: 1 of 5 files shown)


describegroups.go
```diff
@@ -69,6 +69,18 @@ type DescribeGroupsResponseMemberMetadata struct {
 
 	// UserData is the user data for the member.
 	UserData []byte
+
+	// OwnedPartitions contains the partitions owned by this group member; only set if
+	// consumers are using a cooperative rebalancing assignor protocol.
+	OwnedPartitions []DescribeGroupsResponseMemberMetadataOwnedPartition
+}
+
+type DescribeGroupsResponseMemberMetadataOwnedPartition struct {
```
achille-roussel, line 78: Maximum Javaname intensity achieved :trollface:

reader_test.go
```diff
@@ -571,6 +572,14 @@ func BenchmarkReader(b *testing.B) {
 }
 
 func TestCloseLeavesGroup(t *testing.T) {
+	if os.Getenv("KAFKA_VERSION") == "2.3.1" {
+		// There's a bug in 2.3.1 that causes the MemberMetadata to be in the wrong format and thus
+		// leads to an error when decoding the DescribeGroupsResponse.
+		//
+		// See https://issues.apache.org/jira/browse/KAFKA-9150 for details.
+		t.Skip("Skipping because kafka version is 2.3.1")
```
achille-roussel, line 580: Oh Kafka...

achille-roussel (approved):
Nice change, thanks for submitting the fix :)

describegroups.go
yolken-segment: Haha, yes! Let's hope that the protocol isn't modified to add another layer of nesting, otherwise we're really going to be in trouble.

stevevls (approved):
Nice!

erikdw (approved):
🚀  🪨 ⭐

PR #753, opened 2021-10-05
Message set reader allows different versions of batches in fetch response.

collinvandyck:
Typically, batches returned from a broker will all be the same version; batches in a response will usually be Messages (legacy) or Record Batches. Sometimes, depending on the client used, or when a broker is upgraded, this will not be the case.

For example, this is a fetch response we saw come back from a broker in a staging environment:

image

Before this change, the reader would inspect the magic byte of the first batch in the response, and then attempt to read each batch using the format indicated by this magic byte (0 and 1 are older formats, while 2 is the most recent format). In a fetch response in which there are different kinds of batches, this causes an error as the reader parses the data incorrectly. The behavior at this point is undetermined -- the reader can either panic, return an error, etc.

This PR replaces the old message set reader with a new one that parses each batch using a format indicated by the magic byte in the batch.

To ensure coverage, new unit tests and fixture tests (using responses from an actual broker) are added. Additionally, a new response builder type was added to enable the unit tests to create a wide variety of mixed batch types and compression formats.


message_test.go
```diff
@@ -37,6 +347,117 @@ func TestMessageSetReaderEmpty(t *testing.T) {
...
+			_, err = io.Copy(buf, hex.NewDecoder(bytes.NewReader(bs)))
+			require.NoError(t, err)
+
+			// discard 4 byte len and 4 byte correlation id
+			bs = make([]byte, 8)
+			buf.Read(bs)
+
+			rh, err := newReaderHelper(t, buf.Bytes())
+			require.NoError(t, err)
+			messagesCount := 0
```
erikdw, line 445: nit (on myself since I wrote this part!):  `messagesCount` -> `messageCount` to be consistent pluralization-wise with `expectedMessageCount`.

collinvandyck:
Ready for review :tada:

PR #754, opened 2021-10-05
add encoding.TextMarshaler and encoding.TextUnmarshaler implementations

achille-roussel:
This PR implements the standard `encoding.TextMarshaler` and `encoding.TextUnmarshaler` interfaces on the `kafka.RequiredAcks` and `kafka.Compression` types.

The intent is to facilitate loading and saving of these configuration values via textual representations; a common use case would be to have them provided in a configuration file, or on the command line of a kafka producer.

### DIFF
--- a/compress/compress.go
+++ b/compress/compress.go
@@ -4,6 +4,7 @@ import (
 	"encoding"
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 
 	"github.com/segmentio/kafka-go/compress/gzip"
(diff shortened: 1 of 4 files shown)


compress/compress.go
```diff
@@ -33,6 +38,50 @@ func (c Compression) String() string {
...
+	}
+
+	i, err := strconv.ParseInt(string(b), 10, 64)
+	if err == nil && i >= 0 && i < int64(len(Codecs)) {
+		*c = Compression(i)
+		return nil
+	}
+
+	s := &strings.Builder{}
+	s.WriteString("none")
```
rhansen2, line 66: Should uncompressed be include here since it's accepted above?

rhansen2 (approved):
LGTM

PR #755, opened 2021-10-06
Sketch of API

rhansen2:


theflyingcodr:
Hi, this api looks pretty good and like its similarity to sql transactions. Just wondering the status of this and ETA, i have an annoying architect wanting exactly once semantics and I'd prefer to use this lib than the confluent one if possible.

transaction.go
```diff
@@ -0,0 +1,124 @@
...
+		ProducerEpoch:   t.producerEpoch,
+		Committed:       true,
+	})
+	if err != nil {
+		return err
+	}
+	return resp.Error
+}
+
+func (t *Transaction) SendOffsets(ctx context.Context, offsets map[string][]TxnOffsetCommit, generation *Generation) error {
```
rhansen2, line 86: Using passing in the generation here is a little jenky but we need access to some of its fields for the new version of the kafka api. Readers don't expose this so we'd potentially have to add a way to retrieve it or somehow link a reader to the writer. But that seems less flexible.

writer.go
```diff
@@ -615,36 +668,82 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
...
+	ErrTxnNotInitialized = errors.New("transactions not initialized")
+	ErrTxnAlreadyStarted = errors.New("there is on ongoing transaction")
+)
+
+func (w *Writer) BeginTxn(ctx context.Context) (*Transaction, error) {
+	// Need to enforce only having a single txn active at a time.
+	w.txMutex.Lock()
+	defer w.txMutex.Unlock()
+
+	if w.currentTx != nil {
```
rhansen2, line 721: Still need to set this somewhere.

changkun:
What's the status of this PR? How could the community provide help for landing this feature?

githubsands:
Any status on this PR?

Issue #661, opened 2021-05-11
SASL: Support AWS_MSK_IAM mechanism


jgiles:
**Describe the solution you'd like**
 * Add a package with a [sasl.Mechanism](https://github.com/segmentio/kafka-go/blob/84623745d7b8ae64830d78bac17afc0b839def5a/sasl/sasl.go#L11) implementation supporting SASL mechanism AWS_MSK_IAM for IAM-based authentication to AWS MSK managed Kafka.
 * Alternatively (if there are concerns about including AWS-specific code in this repo), establish and document a mechanism for registering + using externally-implemented SASL mechanisms (probably this is just a documentation update pointing at sasl.Mechanism and then listing known OSS impls)

**Supporting documentation**
* MSK is a popular managed Kafka solution on AWS
* AWS announces IAM auth support: https://aws.amazon.com/about-aws/whats-new/2021/05/introducing-iam-access-control-amazon-msk/
* Reference implementation in Java: https://github.com/aws/aws-msk-iam-auth
* Go library for underlying sigv4 signing: https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.38.37/aws/signer/v4
* Related issues
   * https://github.com/segmentio/kafka-go/issues/539
   * https://github.com/segmentio/kafka-go/issues/238
   * https://github.com/segmentio/kafka-go/issues/237

twmb:
For potential implementation reference, I've added support for AWS_MSK_IAM in my [franz-go](https://github.com/twmb/franz-go/blob/master/CHANGELOG.md#v073) client.

achille-roussel:
Hello @jgiles, thanks for opening the discussion.

We've historically refrained from bringing in more SASL mechanisms in kafka-go due to the external dependencies that they tend to require, and the often unfortunate consequences of forcing external dependencies on programs that depend on kafka-go (too many times we have had builds broken by external dependencies disappearing or not respecting versioning rules, etc...).

We do recognize that there is a need to integrate kafka-go with these SASL mechanisms, so we are thinking of introducing them as a separate modules within the repository, probably with a structure like this:
```
sasl
├── msk
│   ├── go.mod
│   ├── go.sum
│   └── msk.go
├── plain
│   └── plain.go
├── scram
│   └── scram.go
├── sasl.go
└── sasl_test.go
```

The `sasl/msk` module would depend on kafka-go and other dependencies (like the AWS SDK), but it won't force the external dependencies on clients that don't import it.

[...]

PR #763, opened 2021-10-21
Add AWS_MSK_IAM SASL Mechanism

cmaher:
Addresses #661 

I've verified this against a live MSK cluster and confirmed the behavior relative to other implementations. Unfortunately, live connections can't be unit-tested, so I'm testing the signature generation, which is the heart of this mechanism. Also, as discussed in the original ticket, I've made this a submodule, with the intent that it won't cause conflicts with general user code. Please let me know if I need to change/add anything.


dialer.go
```diff
@@ -258,6 +258,8 @@ func (d *Dialer) connectTLS(ctx context.Context, conn net.Conn, config *tls.Conf
 	return
 }
 
+const ContextKeyBrokerAddr = "brokerAddr"
```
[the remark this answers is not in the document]
dominicbarnes: It looks like this is actually unnecessary now that `sasl.MetadataFromContext` is being used. Would you mind removing this const and updating the tests that still reference it?

sasl/aws_msk_iam/go.mod
```diff
@@ -6,3 +6,5 @@ require (
 	github.com/aws/aws-sdk-go v1.41.3
 	github.com/segmentio/kafka-go v0.4.21
 )
+
+replace github.com/segmentio/kafka-go => ../..
```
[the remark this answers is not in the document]
dominicbarnes: Ah yeah, in that case I would think we can just remove this line before merging, once we've gotten tests to pass. I would rather avoid 2 PRs.

sasl/aws_msk_iam/msk_iam.go
```diff
@@ -0,0 +1,120 @@
...
+// 	  "x-amz-signature" : "<AWS SigV4 signature computed by the client>"
+// 	}
+func (m *Mechanism) Start(ctx context.Context) (sess sasl.StateMachine, ir []byte, err error) {
+	saslMeta := sasl.MetadataFromContext(ctx)
+	query := url.Values{
+		queryActionKey: {signAction},
+	}
+	signUrl := url.URL{
+		Scheme:   "kafka",
+		Host:     saslMeta.Host,
```
dominicbarnes, line 71: Locally, my tests panic because `saslMeta` does not exist, which I think will be corrected by updating your test cases.

That being said, should we guard against this and return an explicit error message if we detect this? (rather than panicking)

sasl/aws_msk_iam/msk_iam_test.go
```diff
@@ -0,0 +1,88 @@
...
+// using a fixed time allows the signature to be verifiable in a test
+var signTime = time.Date(2021, 10, 14, 13, 5, 0, 0, time.UTC)
+
+func TestAwsMskIamMechanism(t *testing.T) {
+	tests := []struct {
+		ctx func() context.Context
+	}{
+		{
+			ctx: func() context.Context {
+				return context.WithValue(context.Background(),
```
dominicbarnes, line 29: I think these need to be updated to use `sasl.WithMetadata`.

dominicbarnes (changes requested):
Looks like we're getting closer!

dominicbarnes:
Howdy again! Sorry about the delay, I've been trying to get this code to work with a manually-provisioned MSK cluster just to make sure I've at least seen it in action for myself (since as you pointed out, unit-testing this is hard/unfeasible). Thankfully, after some trial and error of my own, I was able to get it to work! That makes me more comfortable moving forward with merging even without some added test coverage beyond what we have here.

I'll follow up on the existing threads, as I think we have a few outstanding ones to finish addressing. That being said, this is looking to be really close.

.circleci/config.yml
```diff
@@ -40,6 +40,7 @@ jobs:
         paths:
         - /go/pkg/mod
     - run: go test -race -cover ./...
+    - run: find ./sasl -name go.mod -execdir go test -race -cover ./... \;
```
dominicbarnes, line 43: I know this was my idea, but it looks like it was a bad one, sorry about that! It looks like the exit code from the `-execdir` is being lost, so CircleCI is treating these as passing even though they panic. :(

If we don't have any other ideas, feel free to revert back to what you had before.

cmaher:
Hey, thanks for getting back to this. I probably don't have time this week, but I'll be able to get to it within the next two weeks after the holiday.

dominicbarnes:
This PR has been addressed by #798 so I will close this. Thank you so much for your contribution!

PR #767, opened 2021-10-25
Expose metadata request parameters for cluster discovery routine

e-chip:
I have received complaints from our kafka cluster admins that service my team was developing produces massive amount of logs on kafka side: 
```
[2021-10-25 13:24:00,836] INFO Principal = User:user_rw is Denied Operation = Alter from host = 1.2.3.4 on resource = Topic:LITERAL:the_topic (kafka.authorizer.logger)
[2021-10-25 13:24:00,836] INFO Principal = User:user_rw is Denied Operation = AlterConfigs from host = 1.2.3.4 on resource = Topic:LITERAL:the_topic (kafka.authorizer.logger)
[2021-10-25 13:24:00,837] INFO Principal = User:user_rw is Denied Operation = Delete from host = 1.2.3.4 on resource = Topic:LITERAL:the_topic (kafka.authorizer.logger)
[2021-10-25 13:24:00,837] INFO Principal = User:user_rw is Denied Operation = DescribeConfigs from host = 1.2.3.4 on resource = Topic:LITERAL:the_topic (kafka.authorizer.logger)
```

After short investigation i have found the reason: `connPool.discover()` routine was sending `Metadata` request with `include_cluster_authorized_operations` and `include_topic_authorized_operations` flags set every 6 seconds (default value for `Transport.MetadataTTL`).

Quick soultion is to set `MetadataTTL` to crazy high value but for the future is a good option to have these parameters exposed.

P.S.
Thanks for the great package for kafka :)

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -113,8 +113,27 @@ type Transport struct {
 	// If nil, context.Background() is used instead.
 	Context context.Context
 
+	// Options for cluster metadata discovery.
+	//
+	// If nil, DefaultDiscoverOptions are used.
+	DiscoverOptions *DiscoverOptions
+
 	mutex sync.RWMutex
 	pools map[networkAddress]*connPool
+}
+
+// DiscoverOptions specify Metadata request parameters used by connPool to discover cluster metadata.
+type DiscoverOptions struct {
+	IncludeClusterAuthorizedOperations bool
+	IncludeTopicAuthorizedOperations   bool
+	AllowAutoTopicCreation             bool
+}
+
+// DefaultDiscoverOptions are the default options for Metadata request parameters.
+var DefaultDiscoverOptions = &DiscoverOptions{
+	IncludeClusterAuthorizedOperations: true,
+	IncludeTopicAuthorizedOperations:   true,
+	AllowAutoTopicCreation:             false,
 }
 
 // DefaultTransport is the default transport used by kafka clients in this
@@ -248,6 +267,8 @@ func (t *Transport) grabPool(addr net.Addr) *connPool {
 		wake:   make(chan event),
 		conns:  make(map[int32]*connGroup),
 		cancel: cancel,
+
+		discoverOptions: t.discoverOptions(),
 	}
 
 	p.ctrl = p.newConnGroup(addr)
@@ -265,6 +286,13 @@ func (t *Transport) context() context.Context {
 		return t.Context
 	}
 	return context.Background()
+}
+
+func (t *Transport) discoverOptions() *DiscoverOptions {
+	if t.DiscoverOptions == nil {
+		return DefaultDiscoverOptions
+	}
+	return t.DiscoverOptions
 }
 
 type event chan struct{}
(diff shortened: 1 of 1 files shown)


transport.go
```diff
@@ -113,10 +113,29 @@ type Transport struct {
 	// If nil, context.Background() is used instead.
 	Context context.Context
 
+	// Options for cluster metadata discovery.
+	//
+	// If nil, DefaultDiscoverOptions are used.
+	DiscoverOptions *DiscoverOptions
```
achille-roussel, line 119: Could we have the boolean values configured on the `Transport` type directly?

achille-roussel (commented):
Hello @e-chip, thanks for submitting a PR!

I left a question, let me know what you think.

transport.go
e-chip: With exposed booleans we will loose backwards compatibility as default values will be set to `false`. Or we have to make ugly inversion of logic (smth like `DoNotIncludeClusterAuthorizedOperations`)

achille-roussel: You make a good point, let me look back into this in more details and I'll get back to you.

feketiko: Hi @achille-roussel! 

We are running into the same issue on PROD.
Is there a way to prioritise this, please?

Thanks!

achille-roussel: We merged a change that maybe will have a positive impact on the original issue here https://github.com/segmentio/kafka-go/pull/821

Would you be able to let me know if this does address the concern or we need to move forward with this PR?

e-chip: That's definitely solves the issue, like chopping off the legs in order not to hit your head while running 😆 

I would prefer my solution because it leaves space for future development, does not break backward compatibility and conforms with open/close principle.

dnwe: @e-chip sorry I didn't see this old PR when I raised the problem recently. See the original issue https://github.com/segmentio/kafka-go/issues/818 for what I thought the next step should be _after_ disabling the current code in order to potentially expose the ACL fetchers on the various Describe operations etc.

There's no "backward compatibility" to preserve here, because none of the existing client code was using the responses anyway, so disabling is the most immediate remedy. It also doesn't really make sense to be caching the authorized operations responses in this background metadata updater because you'd typically always want the latest information from the cluster for those.

achille-roussel: > I would prefer my solution because it leaves space for future development, does not break backward compatibility and conforms with open/close principle.

What I am inclined to is making sure we introduce APIs to serve known use cases; knowing how it will be employed is highly valuable to figure out what the right interface would look like. At this time it appears we haven't had a need to configure those parameters, so the best decision we can make is only driven by our experience and assumptions, therefore we are more likely to make the wrong choices.

achille-roussel:
I truly appreciate your contribution @e-chip, and the work you've done has helped us land on a lower scope change with #821, which we believe addresses the original issue.

I'll close this pull request for now, we will revisit the proposal when a use case for retrieving Cluster and Topic authorized operations arises.

Issue #683, opened 2021-06-16
Unknown Topic Or Partition error, even though the topic should be created automatically


tim-hilt:
**Describe the bug**

When publishing a message, I get `Unknown Topic Or Partition: the request is for a topic or partition that does not exist on this broker`, event though it [should create the message by default, as far as I understand the `README.md`](https://github.com/segmentio/kafka-go#to-create-topics).

It worked before, when I was using the exact same setup on NixOS (until I switched to Arch Linux yesterday).

**Kafka Version**

Latest [docker-image from confluentinc](https://hub.docker.com/r/confluentinc/cp-kafka) (should be 2.7.1 if it's the latest official release from their GitHub, as of today)

**To Reproduce**

Start Kafka using the following `docker-compose.yaml`:

``` yaml
version: "3.3"
services:
  zookeeper:
    image: confluentinc/cp-zookeeper:latest
    environment:
      ZOOKEEPER_CLIENT_PORT: 2181
      ZOOKEEPER_TICK_TIME: 2000
    ports:
      - 22181:2181
  kafka:
    image: confluentinc/cp-kafka:latest
    depends_on:
      - zookeeper
    ports:
      - 29092:29092
    environment:
      KAFKA_BROKER_ID: 1
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:9092,PLAINTEXT_HOST://localhost:29092
      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
      KAFKA_INTER_BROKER_LISTENER_NAME: PLAINTEXT
      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
```

execute the following commands:

``` shell
docker-compose -f "FILE UNDER WHICH YOU'VE STORED THE ABOVE" up -d
mkdir tmp
cd tmp
go mod init example.com/tmp
go get -u github.com/segmentio/kafka-go
vim main.go
```

Fill `main.go` with the following lines:

``` go
package main

import (
	"context"

	"github.com/segmentio/kafka-go"
)

func main() {
	w := &kafka.Writer{
		Addr:  kafka.TCP("localhost:29092"),
		Topic: "hello",
	}

	err := w.WriteMessages(context.Background(), kafka.Message{
		Key:   []byte("hello"),
[...]

PR #775, opened 2021-10-31
Allow Writer to automatically create topics.

rhansen2:
Potential Fix for #683 

Writer now sets AutoCreateTopics to true when calling for metadata.

Transport will now make a metadata request to the server if the metadata cache does not contain the requested topic and AutoCreateTopics is true. We then update the metadata cache after sending the metadata request to the server.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -360,7 +360,9 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
 		var requestNeeded bool
 		if m.AllowAutoTopicCreation {
 			for _, topic := range cachedMeta.Topics {
-				requestNeeded = topic.ErrorCode == int16(UnknownTopicOrPartition)
+				if topic.ErrorCode == int16(UnknownTopicOrPartition) {
+					requestNeeded = true
+				}
 			}
 		}
 
--- a/transport.go
+++ b/transport.go
@@ -358,9 +358,9 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
 		// It's true when we want to auto-create topics and we don't have the topic in our
 		// cache.
 		var requestNeeded bool
-		for _, topic := range cachedMeta.Topics {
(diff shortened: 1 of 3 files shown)


transport.go
```diff
@@ -344,14 +344,29 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
...
 		}
-		return filterMetadataResponse(m, state.metadata), nil
+
+		cachedMeta := filterMetadataResponse(m, state.metadata)
+		// requestNeeded indicates if we need to send this metadata request to the server.
+		// It's true when we want to auto-create topics and we don't have the topic in our
+		// cache.
+		var requestNeeded bool
+		for _, topic := range cachedMeta.Topics {
+			if topic.ErrorCode == int16(UnknownTopicOrPartition) && m.AllowAutoTopicCreation {
```
achille-roussel, line 362: Should we check for AllowAutoTopicCreation before the loop?

achille-roussel (approved):
Looking good :) I left a question on a minor detail but otherwise it's good to merge 👍

transport.go
```diff
@@ -344,14 +344,31 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
...
+
+		cachedMeta := filterMetadataResponse(m, state.metadata)
+		// requestNeeded indicates if we need to send this metadata request to the server.
+		// It's true when we want to auto-create topics and we don't have the topic in our
+		// cache.
+		var requestNeeded bool
+		if m.AllowAutoTopicCreation {
+			for _, topic := range cachedMeta.Topics {
+				if topic.ErrorCode == int16(UnknownTopicOrPartition) {
+					requestNeeded = true
```
achille-roussel, line 364: Probably a very minor optimization, tho we could break here after seeing `requestNeeded`.

rhansen2:
I think i've gotten the test worked out, should be ready for a final review.

achille-roussel (approved):
🚢

PR #776, opened 2021-11-01
Check for broker existence when forming partition replica and ISR lists

bhavanki:
Data for each partition returned by Conn.ReadPartitions includes the
list of replicas and the ISR for the partition. The broker data is
copied from the list of currently available brokers, retrieved from
Kafka metadata. It can happen that a broker is not present in metadata
(due to being down, for example), but still listed as a replica for a
partition. (For example, broker 2 may be down but the ID 2 can still be
listed as a replica for a partition.)

The logic that copies broker data from the list of available brokers
into partition data now omits any that are not present in the metadata
list. Without this change, partition data receives the copy of a nil
object in its replica or ISR list (ID 0, host nil, default port 9092),
which is useless.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -995,9 +995,11 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
 			}
 
 			makeBrokers := func(ids ...int32) []Broker {
-				b := make([]Broker, len(ids))
-				for i, id := range ids {
-					b[i] = brokers[id]
+				var b []Broker
+				for _, id := range ids {
+					if br, ok := brokers[id]; ok {
+						b = append(b, br)
+					}
 				}
 				return b
 			}
--- a/conn.go
+++ b/conn.go
@@ -995,7 +995,7 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
 			}
 
 			makeBrokers := func(ids ...int32) []Broker {
-				var b []Broker
+				b := make([]Broker, 0, len(ids))
 				for _, id := range ids {
 					if br, ok := brokers[id]; ok {
 						b = append(b, br)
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -995,9 +995,11 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
 			}
 
 			makeBrokers := func(ids ...int32) []Broker {
-				b := make([]Broker, len(ids))
-				for i, id := range ids {
-					b[i] = brokers[id]
+				var b []Broker
```
stevevls, line 998: This is pretty minor, but you can avoid potential allocations when growing the slice by allocating the space up-front:

```suggestion
				b := make([]Broker, 0, len(ids))
```

stevevls (commented):
LGTM!  I wouldn't be surprised if it's prohibitively tough, but is there any way we could add unit tests around this behavior?

conn.go
```diff
@@ -995,9 +995,11 @@ func (c *Conn) ReadPartitions(topics ...string) (partitions []Partition, err err
 			}
 
 			makeBrokers := func(ids ...int32) []Broker {
-				b := make([]Broker, len(ids))
```
collinvandyck, line 998: I think the original code here was trying to avoid excess allocations. You can do this with your code too, by changing this:

```
var b []Broker
```

to this:

```
b := make([]Broker, 0, len(ids))
```

That form sets the capacity of the slice to `len(ids)` but the length to 0, so your appends still work as expected, but you're guaranteed to avoid additional allocations, at the cost of possibly having a slice which has more capacity than its length.

bhavanki:
> LGTM! I wouldn't be surprised if it's prohibitively tough, but is there any way we could add unit tests around this behavior?

Done! I split the `makeBrokers` func out and wrote a couple of focused unit tests for it.

collinvandyck (approved):

PR #782, opened 2021-11-08
Ensure tests run under Go 1.16

bhavanki:
Uses of time.UnixMilli() introduced recently, which only work with Go
1.17 or higher, are replaced with equivalent code that works under Go
1.16. If/when this module moves past Go 1.16, these changes may be
reverted.


achille-roussel (approved):
Thanks for the fix!

PR #783, opened 2021-11-08
Fix go vet issues: example tests, returning a mutex copy

bhavanki:
Example tests are renamed to fit required naming conventions. This
caused them to be relocated in generated docs.

Also, a minor problem with writer.newBatchQueue is fixed, in order to
avoid an unnecessary copy of a newly created object holding a mutex.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -867,8 +867,8 @@ func (w *Writer) chooseTopic(msg Message) (string, error) {
 type batchQueue struct {
 	queue []*writeBatch
 
-	mutex sync.Mutex
-	cond  sync.Cond
+	mutex *sync.Mutex
+	cond  *sync.Cond
 
 	closed bool
 }
@@ -912,14 +912,16 @@ func (b *batchQueue) Close() {
 	b.closed = true
 }
 
-func newBatchQueue(initialSize int) *batchQueue {
+func newBatchQueue(initialSize int) batchQueue {
 	bq := batchQueue{
 		queue: make([]*writeBatch, 0, initialSize),
+		mutex: &sync.Mutex{},
+		cond: &sync.Cond{},
 	}
 
-	bq.cond.L = &bq.mutex
+	bq.cond.L = bq.mutex
 
-	return &bq
+	return bq
 }
 
 // partitionWriter is a writer for a topic-partion pair. It maintains messaging order
@@ -941,7 +943,7 @@ type partitionWriter struct {
 func newPartitionWriter(w *Writer, key topicPartition) *partitionWriter {
 	writer := &partitionWriter{
 		meta:  key,
-		queue: *newBatchQueue(10),
+		queue: newBatchQueue(10),
 		w:     w,
 	}
 	go func() {
--- a/writer.go
+++ b/writer.go
@@ -912,14 +912,14 @@ func (b *batchQueue) Close() {
 	b.closed = true
 }
 
-func newBatchQueue(initialSize int) batchQueue {
+func newBatchQueue(initialSize int) *batchQueue {
 	bq := batchQueue{
 		queue: make([]*writeBatch, 0, initialSize),
 	}
 
 	bq.cond.L = &bq.mutex
 
-	return bq
+	return &bq
 }
 
 // partitionWriter is a writer for a topic-partion pair. It maintains messaging order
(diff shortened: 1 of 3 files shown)


bhavanki:
Looking at the batchQueue change again, I'm wondering if it's actually a significant change. I only made it mechanistically to make `go vet` happy.

writer.go
```diff
@@ -912,14 +912,14 @@ func (b *batchQueue) Close() {
...
-func newBatchQueue(initialSize int) batchQueue {
+func newBatchQueue(initialSize int) *batchQueue {
 	bq := batchQueue{
 		queue: make([]*writeBatch, 0, initialSize),
 	}
 
 	bq.cond.L = &bq.mutex
 
-	return bq
+	return &bq
```
achille-roussel, line 922: The existing code is concerning indeed, I wonder if it works just because the function gets inlined... TBD.

I can see two options:
- change the type of `queue` to `*batchQueue` so we don't make a copy
- have an `init` method on `batchQueue` instead of using a constructor

achille-roussel:
Hey @bhavanki, I just wanted to follow up, do you need any help to land this PR?

writer.go
rhansen2: FWIW per the docs. Mutexes mustn't  be copied after their first use.

writer.go
```diff
@@ -941,7 +941,7 @@ type partitionWriter struct {
 func newPartitionWriter(w *Writer, key topicPartition) *partitionWriter {
 	writer := &partitionWriter{
 		meta:  key,
-		queue: newBatchQueue(10),
+		queue: *newBatchQueue(10),
```
rhansen2, line 944: Isn't this also a mutex copy?

bhavanki:
> Hey @bhavanki, I just wanted to follow up, do you need any help to land this PR?

I've been preoccupied with some other more pressing issues, but I'm getting back to this one now! As long as folks continue to review and give feedback, it's all good.

writer.go
bhavanki: `go vet` doesn't complain about it, so perhaps not. If a pointer deference made a copy (in Go generally), then every dereference would produce a fresh copy and you'd never be able to work with the original. At least, that's my thinking, but I'm still new-ish to Go, so I'm happy to continue getting your insight!

writer.go
bhavanki: Correct, that's what `go vet` was complaining about:

```
./writer.go:922:9: return copies lock value: github.com/segmentio/kafka-go.batchQueue contains sync.Mutex
```

To be clear, the problem is that returning a `batchQueue` value copies the mutex within. (Perhaps that mutex field should be a pointer ...) The `queue` field doesn't seem to be a problem.

bhavanki: OK! I took a different tack. Rather than having this function return a pointer, I instead changed the types of the mutex and cond fields in batchQueue to pointers. That seems like a much more focused fix for the problem. @achille-roussel and @rhansen2 , what do you think?

writer.go
rhansen2: I don't believe the pointer dereference makes a copy but I think this initialization syntax turns into an assignment which would copy the dereferenced struct including the mutex.

bhavanki: Oh, that's tricky! Thanks for the extra information!

rhansen2: Found  https://cs.opensource.google/go/x/tools/+/refs/tags/v0.1.7:go/analysis/passes/copylock/testdata/src/a/copylock_func.go;l=119 
and 
https://github.com/golang/go/issues/16227

as well that talk about this exact case as well as copying zero valued locks being fine.

bhavanki: @rhansen2 I reworked this change in a new commit back on Monday, so that this field is no longer initialized with a pointer dereference. But, I'm not sure now what kind of change you are in favor of, based on your last comment. Would you prefer that the code be reverted to copying mutexes as it originally did - since it's actually safe - even though vet complains about it?

rhansen2: I'll defer to @achille-roussel  on this.

achille-roussel (approved):
The fix is looking good 👍

writer.go
achille-roussel: I agree that the ideal code shape would be to no have a pointer-to-mutex, as this is an uncommon pattern (sometimes presented as an anti-pattern), so despite being correct it may be cause of concerns for future readers. 

An alternative approach could be to not have a constructor-style function, and instead use an initialization method:
```go
func (bq *batchQueue) init() {
  // do all the initialization here, no mutex copies happen
  ...
}
```

All that being said, a pragmatic move could be to accept the current change as it does fix the `go vet` warning and remains correct.

Up to you @bhavanki

bhavanki: @achille-roussel I'll leave a comment with the concerns around introducing a pointer-to-mutex, and then merge what's here now. Someone can jump in later on and further revise the code to improve it more.

PR #784, opened 2021-11-11
Changed safeTimeout from 10*time.Second to 60*time.Second to resolve …

lujiacn:
…the tcp i/o timeout issue for large date reading


achille-roussel:
Hello @lujiacn, thanks for opening a PR!

Would you be able to add a test that demonstrate the original issue you were running into?

lujiacn:
@achille-roussel, thanks for reply. I would like to close this PR, change to 60 secs only temporary solution for my case, if connection waiting more than 60 secs, still facing the possible tcp timeout issue.

PR #786, opened 2021-11-13
fix panic cause by waitgroup reuse

lqczzz:
## issue desctiption

panic: sync: WaitGroup is reused before previous Wait has returned

## issue cause analysis

when genegration object created, it will only be done called by 

`gen.close()` in one goroutine

and `gen.close()` will only be called when capturing gen.done or cg.done signal 

see: https://github.com/segmentio/kafka-go/blob/main/consumergroup.go#L804-L821

`select {
	case <-cg.done:
		gen.close()
		return memberID, ErrGroupClosed // ErrGroupClosed will trigger leave logic.
	case cg.next <- &gen:
	}

	// wait for generation to complete.  if the CG is closed before the
	// generation is finished, exit and leave the group.
	select {
	case <-cg.done:
		gen.close()
		return memberID, ErrGroupClosed // ErrGroupClosed will trigger leave logic.
	case <-gen.done:
		// time for next generation!  make sure all the current go routines exit
		// before continuing onward.
		gen.close()
		return memberID, nil
	}`

Since gen.close() will run `g.wg.Wait()`, but there is another goroutine would call `gen.Start()`, which will exec `g.wg.Add(1)`
https://github.com/segmentio/kafka-go/blob/main/reader.go#L330-L333

which leads to panic 

image

## solution:
1. add mutex for wg.Add() and wg.Wait() since `generation.Start` may run concurrently with `generation.close`
2. `Start()` function should watch gen.Done before exec `gen.wg.Add(1)`

### DIFF
--- a/consumergroup.go
+++ b/consumergroup.go
@@ -325,8 +325,8 @@ type Generation struct {
 	once sync.Once
 	done chan struct{}
 
-	wgMutex sync.Mutex
-	wg      sync.WaitGroup
+	wgMu sync.Mutex
+	wg   sync.WaitGroup
 
 	retentionMillis int64
 	log             func(func(Logger))
@@ -340,9 +340,9 @@ func (g *Generation) close() {
 		close(g.done)
 	})
 
-	g.wgMutex.Lock()
+	g.wgMu.Lock()
 	g.wg.Wait()
-	g.wgMutex.Unlock()
+	g.wgMu.Unlock()
 }
 
 // Start launches the provided function in a go routine and adds accounting such
@@ -359,9 +359,14 @@ func (g *Generation) close() {
 // progress for this consumer and potentially cause consumer group membership
 // churn.
 func (g *Generation) Start(fn func(ctx context.Context)) {
-	g.wgMutex.Lock()
-	defer g.wgMutex.Unlock()
+	g.wgMu.Lock()
+	defer g.wgMu.Unlock()
 
+	select {
+	case <-g.done:
+		return
+	default:
+	}
 	g.wg.Add(1)
 	go func() {
 		fn(genCtx{g})
--- a/consumergroup.go
+++ b/consumergroup.go
@@ -324,7 +324,9 @@ type Generation struct {
 
 	once sync.Once
 	done chan struct{}
-	wg   sync.WaitGroup
+
+	wgMutex sync.Mutex
+	wg      sync.WaitGroup
 
 	retentionMillis int64
 	log             func(func(Logger))
@@ -337,7 +339,10 @@ func (g *Generation) close() {
 	g.once.Do(func() {
 		close(g.done)
 	})
+
+	g.wgMutex.Lock()
 	g.wg.Wait()
+	g.wgMutex.Unlock()
 }
 
 // Start launches the provided function in a go routine and adds accounting such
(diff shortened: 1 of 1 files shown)


stevevls:
Thank you for opening this issue and all the debugging to figure out the root cause.  That's an interesting edge case where the `Reader` is trying to kick off go routines with `Start` and racing with an error!

I think we should make some changes, though.  The expectation is that the function passed to `Generation.Start` will always be run, and these changes would make it unclear to the caller whether that will happen.  There are use cases where the function passed to `Start` may do bookkeeping or provide be used for important synchronization.  Things get tricky because the `Reader` is just one of many clients to `ConsumerGroup`.   It's a type that we export, so there are other people (including some projects at Segment) that also use `ConsumerGroup` directly.  And as is always the case with exported types, we need to be careful about changing contracts. 😅 

So in terms of requirements, I think that we should say:
1. We don't want `Start` and `close` to race with one another.
2. We want all functions passed to `Start` to run

To clarify on the latter case when `close` has already been called, we would expect the function passed to `Start` runs but should exit pretty much immediately b/c it should be listening on the generation's context.  _But_ it will get the opportunity to run.

lqczzz:
> Thank you for opening this issue and all the debugging to figure out the root cause. That's an interesting edge case where the `Reader` is trying to kick off go routines with `Start` and racing with an error!
> 
> I think we should make some changes, though. The expectation is that the function passed to `Generation.Start` will always be run, and these changes would make it unclear to the caller whether that will happen. There are use cases where the function passed to `Start` may do bookkeeping or provide be used for important synchronization. Things get tricky because the `Reader` is just one of many clients to `ConsumerGroup`. It's a type that we export, so there are other people (including some projects at Segment) that also use `ConsumerGroup` directly. And as is always the case with exported types, we need to be careful about changing contracts. 😅
> 
> So in terms of requirements, I think that we should say:
> 
> 1. We don't want `Start` and `close` to race with one another.
> 2. We want all functions passed to `Start` to run
> 
> To clarify on the latter case when `close` has already been called, we would expect the function passed to `Start` runs but should exit pretty much immediately b/c it should be listening on the generation's context. _But_ it will get the opportunity to run.

thanks for your reply！

i really agree that we should be careful about changing any export type/function/variable..etc ! 

but i think is hard to meet these two requirements

suppose we give opportunity to fn like 

```
func (g *Generation) Start(fn func(ctx context.Context)) {
	g.wgMu.Lock()
	defer g.wgMu.Unlock()

	select {
	case <-g.done:
                // give opportunity to run fn even if g.close() called
                fn(genCtx{g})
		return
	default:
	}
        g.wg.Add(1)
        // ...
    }

or like :

func (g *Generation) Start(fn func(ctx context.Context)) {
	g.wgMu.Lock()
	defer g.wgMu.Unlock()
       
        // waitgroup can be reuse even if g.wg.Wait() called
        // but it's meanlesss for reuse waitgroup because no waiters any more
        g.wg.Add(1)
        // ...
    }

```



these changes breaks the  synchronization-promise of `close` function 
futhur more
consumerGroup will enter next generation,see
https://github.com/segmentio/kafka-go/blob/main/consumergroup.go#L674-L684

------

waiting for your genius idea, haha 😄

stevevls:
Heh @lqczzz .  I just opened up #791 with my suggestion of how we could do this.  I chose to open a PR because I think it's clearer to talk concretely about the code rather than try to type up suggestions here.  It accomplishes the goals, but it's possible that there's a simpler way to do it, so I'm open to feedback. 😄  Let me know what you think!

stevevls:
This should be resolved by #791

PR #788, opened 2021-11-15
Nettest fix

rhansen2:
Seems like In newer versions of Kafka the high water mark being equal to the offset does not mean there are no messages in the batch.


achille-roussel (approved):
Nice work 🙌

PR #791, opened 2021-11-18
Fix a rare panic during consumer group rebalance

stevevls:
Currently, there is a `sync.WaitGroup` that is used for process accounting
with go routines started by `Generation.Start`.  It is used when a
generation is ending to ensure that all of the Start-ed functions exit
before the next generation can start.  There is an edge case that can
cause panic as written up in #786 due to unsafe use of the WaitGroup.
Basically, it is possible that `close()` is already in process and is
waiting on `WaitGroup.Wait()` when `Start()` comes in and calls
`WaitGroup.Add(1)`.

The contract of `Start()` is that the provided function is to be run.
Code in the wild may depend on this behavior, so it's not an option to
return from `Start()` without running the provided function.

Accordingly, this PR updates the code to coordinate between close() and
`Start()` such that the panic case is no longer possible while preserving
the existing contract.  It uses channels and a mutex in order to create
two cases: the normal case where the generation is alive and the edge
case where the generation has already ended.

### DIFF
--- a/consumergroup.go
+++ b/consumergroup.go
@@ -374,6 +374,7 @@ func (g *Generation) close() {
 // churn.
 func (g *Generation) Start(fn func(ctx context.Context)) {
 	g.lock.Lock()
+	defer g.lock.Unlock()
 
 	// this is an edge case: if the generation has already closed, then it's
 	// possible that the close func has already waited on outstanding go
@@ -384,14 +385,12 @@ func (g *Generation) Start(fn func(ctx context.Context)) {
 	// such a case, fn should immediately exit because ctx.Err() will return
 	// ErrGenerationEnded.
 	if g.closed {
-		g.lock.Unlock()
 		go fn(genCtx{g})
 		return
 	}
 
 	// register that there is one more go routine that's part of this gen.
 	g.routines++
-	g.lock.Unlock()
 
 	go func() {
 		fn(genCtx{g})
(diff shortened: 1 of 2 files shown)


consumergroup.go
```diff
@@ -354,15 +373,43 @@ func (g *Generation) close() {
...
+	// ErrGenerationEnded.
+	if g.closed {
+		g.lock.Unlock()
+		go fn(genCtx{g})
+		return
+	}
+
+	// register that there is one more go routine that's part of this gen.
+	g.routines++
+	g.lock.Unlock()
```
achille-roussel, line 394: I know the current code is correct, but how about deferring the unlock since it happens (almost) right before the return in both cases. I wouldn't expect `Start` to be called so frequently that we want to minimize the critical section.
```suggestion
	g.lock.Lock()
	defer g.lock.Unlock()

	// this is an edge case: if the generation has already closed, then it's
	// possible that the close func has already waited on outstanding go
	// routines and exited.
	//
	// nonetheless, it's important to honor that the fn is invoked in case the
	// calling function is waiting e.g. on a channel send or a WaitGroup.  in
	// such a case, fn should immediately exit because ctx.Err() will return
	// ErrGenerationEnded.
	if g.closed {
		go fn(genCtx{g})
		return
	}

	// register that there is one more go routine that's part of this gen.
	g.routines++
```

achille-roussel (approved):
Looking good 👍

consumergroup.go
stevevls: Sounds good.  I'm still getting used to the fact that defer is cheap. 😄

PR #796, opened 2021-11-20
writer: allow user customized partition.

3AceShowHand:
Sometimes, the user may want to customize the logic of how to dispatch the message to the partition. For example, our app's logic would decide the partition by fields, https://github.com/pingcap/ticdc/blob/master/cdc/sink/dispatcher/table.go#L33.

This is very useful for such scenario.

### DIFF
--- a/message.go
+++ b/message.go
@@ -12,8 +12,10 @@ type Message struct {
 	// not already specified on the writer itself.
 	Topic string
 
-	// Partition is read-only and MUST NOT be set when writing messages
+	// When CustomPartition is true, Partition will be used.
+	CustomPartition bool
 	Partition     int
+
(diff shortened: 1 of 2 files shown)


dominicbarnes:
I think the best way to accomplish this right now is to using the existing `Balancer` configuration/interface. You receive the `Message` as well as a list of possible partitions, so you should have all the context you need to accomplish this.

Let me know if that doesn't fit your use-case, or if you'd like more guidance on how to accomplish this.

dominicbarnes:
I'm going to close this PR, as I'm pretty certain that a custom `Balancer` should suffice here, feel free to follow up if we can provide more guidance on this.

PR #804, opened 2021-12-03
Fix metadata requests leak

### LINKED ISSUE #803: Writer.close() does not stop the requests for metadata, started by Writer.writeMessages()
First of all, thank you for writing this great library.

**Describe the bug**
When calling `writer.Close()` after `writer.WriteMessages()`, the `Writer` keeps causing the periodic making of metadata requests. This happens through `Writer`'s underlying `Transport` where the metadata requests originate from the `discover()` entry point. This is problematic in some of the integration tests that I am writing, where I expect no such leaking of resources.

**Kafka Version**
6.0.1

**To Reproduce**
```golang
func main() {
	c := make(chan bool)

	writer := kafka.Writer{
		Addr: kafka.TCP("localhost:29092"),
		Topic: "topic",
	}

	writer.WriteMessages(context.Background(), kafka.Message{
		Key:   []byte("key"),
		Value: []byte("value"),
	})

	writer.Close();
	<-c
}
```

**Expected behavior**
When calling `writer.Close()` after `writer.WriteMessages()`, I expect `Writer` to close all connections that it caused. And to stop any requests from happening that it previously caused.

arjun-1:
As far as I can tell, `p.refc` keeps track of the amount of times `ref()` was invoked. `ref()` is invoked 2 times before `conPool` `p` is created, hence the starting value is 2. After this, every invocation of `ref()` increments `p.refc`. 

Every time `unref()` is invoked, `p.refc` is decremented. My guess is that we never want to invoke `unref()` more times than `ref` was invoked. Hence `unref()` should take effect exactly when `p.refc != 0`

Fixes #803

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -317,7 +317,7 @@ func (p *connPool) ref() {
 }
 
 func (p *connPool) unref() {
-	if atomic.AddUintptr(&p.refc, ^uintptr(0)) == 0 {
+	if atomic.AddUintptr(&p.refc, ^uintptr(0)) != 0 {
 		p.mutex.Lock()
 		defer p.mutex.Unlock()
 


transport.go
```diff
@@ -317,7 +317,7 @@ func (p *connPool) ref() {
 }
 
 func (p *connPool) unref() {
-	if atomic.AddUintptr(&p.refc, ^uintptr(0)) == 0 {
+	if atomic.AddUintptr(&p.refc, ^uintptr(0)) != 0 {
```
achille-roussel, line 320: The reference counter here intends to track the number of inflight operations so we delay closing the pool until the last inflight operations has completed.

At first sight it seems like this change would result in closing the connection pool on the first call to `unref` isn't it?

achille-roussel:
We have similar needs to wait for open connections to be closed in kafka-go tests, this is how we solved for it:
* https://github.com/segmentio/kafka-go/blob/a375a40e5b09ef4a25db803f9616377b3b85f470/testing/conn.go
* https://github.com/segmentio/kafka-go/blob/4da3b721ca38db775a5024089cdc4ba14f84e698/client_test.go#L85-L102

Let me know if this is useful.

achille-roussel:
This may also be related to https://github.com/segmentio/kafka-go/pull/805, let me know if the change helped address your issues.

arjun-1:
> We have similar needs to wait for open connections to be closed in kafka-go tests, this is how we solved for it:
> 
> * https://github.com/segmentio/kafka-go/blob/a375a40e5b09ef4a25db803f9616377b3b85f470/testing/conn.go
> * https://github.com/segmentio/kafka-go/blob/4da3b721ca38db775a5024089cdc4ba14f84e698/client_test.go#L85-L102
> 
> Let me know if this is useful.

Oh that's a handy snippet! I will try it out tonight (as well as changes from PR #805) and will let you know more. In any case it seems this change has unintended side effects (I wasn't aware of the intention of `refc`), so this PR can be closed 👍.

PR #805, opened 2021-12-05
fix writer async close

### LINKED ISSUE #778: Close function runs before Completion function when writing async
**Describe the bug**
When making use of the Completion function on an async write, there is a race condition between the Completion function and the Close function.
This results in us not receiving the offset information from the async write response channel reliably on small batch counts.

**Kafka Version**
2.8.0

**To Reproduce**
Creation of the kafka writer:

```
func newKafkaWriter(config *util.Config) *kafka.Writer {
	batchTimeout := time.Duration(5) * time.Second
	writeTimeout := time.Duration(120) * time.Second
	readTimeout := time.Duration(180) * time.Second
	return kafka.NewWriter(kafka.WriterConfig{
		Brokers:  []string{config.Kafka.Broker},
		Balancer: &kafka.RoundRobin{},
		Dialer:   newKafkaDialer(config),
		BatchSize:    50,
		BatchBytes:   20971520,
		BatchTimeout: batchTimeout,
		RequiredAcks: 0,
		Async:        true,
		MaxAttempts:  2,
		ErrorLogger:  log.Default(), 
		WriteTimeout: writeTimeout,
		ReadTimeout:  readTimeout,
	})
}
```

Creation of the completion function:

```
func (kw *KafkaWriter) kafkaCompletion() func(messages []kafka.Message, err error) {
      return func(messages []kafka.Message, err error) {
		if err != nil {
			kw.errCh <- err
		}
		for _, message := range messages {
			kw.responseCh <- KafkaTopicOffsetPartition{
				Topic:     message.Topic,
				Partition: message.Partition,
				Offset:    message.Offset,
			}
		}
	}
}
```

The WriteMessage function:

```
func (kw *KafkaWriter) WriteMessage(topic string, batchId string, value string, writeResponse bool) error {
	if kw.kwriter == nil {
		kw.kwriter = newKafkaWriter(kw.Config)
	}
	kafkaMsg := kafka.Message{
		Topic: topic,
		Value: []byte(value),
		Headers: []kafka.Header{{
			Key:   "batchId",
			Value: []byte(batchId),
		}},
	}
	kw.kwriter.Completion = kw.kafkaCompletion()
	kerr := kw.kwriter.WriteMessages(context.Background(), kafkaMsg)
	if kerr != nil {
		fmt.Println(kerr)
		fmt.Println("Establishing new writer kafka connection")
		kw.kwriter = newKafkaWriter(kw.Config)
		kw.kwriter.Completion = kw.kafkaCompletion()
		err := kw.kwriter.WriteMessages(context.Background(), kafkaMsg)
		if err != nil {
			fmt.Println("Failed to write Kafka message to topic [" + kafkaMsg.Topic + "] - " + err.Error())
			return err
		}
	}
	return nil
}
```

The for loop to loop through the responseChannel and get the earliest offset for each partition:

```
for response := range responseCh {
			if _, ok := offsetMap[response.Partition]; !ok || response.Offset < offsetMap[response.Partition] {
                                offsetMap[response.Partition] = response.Offset
			}
		}
```

The creation of the response channel:

```
responseCh := make(chan service.KafkaTopicOffsetPartition, 50)
```

The KafkaTopicOffsetPartition struct:

```type KafkaTopicOffsetPartition struct {
	Topic     string
	Partition int
	Offset    int64
}
```

**Expected behavior**
I would expect the close function to be blocked by the Completion function so that we are able to reliably get the offset information from the response channel;.

**Additional context**
This only happens on small batches ( < 20 records) and it is intermittent. The failure rate is around 50%.

achille-roussel:
This PR simplifies and fixes internal synchronization in the `kafka.Writer` type.

There were a few issues:
* marking the writer as closed was done via an atomic operations in `Close`, racing with the check in `WriteMessages`
* increments of the `partitionWriter` wait group was done asynchronously after starting goroutines, racing with calls to `Wait` on these wait groups

I removed the sync group in `partitionWriter` and instead managed the count of inflight operations globally on a single wait group in `Writer`; we only care about the writer's `Close` call to block until all inflight operations are done, there is no need to have a hierarchical synchronization scheme.

Fixes #778

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -185,12 +185,10 @@ type Writer struct {
 	// If nil, DefaultTransport is used.
 	Transport RoundTripper
 
-	// Atomic flag indicating whether the writer has been closed.
-	closed uint32
-	group  sync.WaitGroup
-
 	// Manages the current set of partition-topic writers.
+	group   sync.WaitGroup
 	mutex   sync.Mutex
+	closed  bool
 	writers map[topicPartition]*partitionWriter
 
 	// writer stats are all made of atomic values, no need for synchronization.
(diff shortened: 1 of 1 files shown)


rhansen2 (approved):

egaldamez:
Hello, this is an interesting fix. I'm currently facing an issues in which I'm trying to Close a writer so that buffered messages would be flushed and sent to my cluster. However, this doesn't always happen. My scenario is as follows:

1. I have a go application (v1.7) that accepts http requests and these requests are transformed into kafka messages (buffered by the application and flushed every 15 seconds or 1000 messages)
2. I use vegeta to send traffic to this application @ 1000 msg/s
3. I stop vegeta. At this point, my application has some messages buffered.
4. I send an OS interrupt signal to my application (which is configured to listen to these signals)
5. I expect that the messages buffered are flushed, but this does not always happen especially when I am sending a high volume of messages.

So I'm wondering if this could be because of the race conditions addressed in this PR. I'm looking forward to fetching the latest version and testing this for myself :) thank you.

Issue #653, opened 2021-05-04
Corruption when receiving message sets with zero messages


Nilix007:
**Describe the bug**
We noticed panics very similar to #457 when reading messages with a consumer group.

<details>
<summary>Stack trace</summary>
<pre>
panic: runtime error: makeslice: len out of range
goroutine 38 [running]:
github.com/segmentio/kafka-go.(*messageSetReaderV2).readMessage(0xc0004fc018, 0x79e519, 0xc0001d3a30, 0xc0001d3a20, 0xc0001d39f4, 0x2, 0x2, 0x10000000000, 0x0, 0x2, ...)
        /go/pkg/mod/github.com/segmentio/kafka-go@v0.4.15/message.go:556 +0x7ba
github.com/segmentio/kafka-go.(*messageSetReader).readMessage(0xc0004fc000, 0x79e519, 0xc0001d3a30, 0xc0001d3a20, 0x2, 0x1759d3d9915, 0x0, 0xc000318be0, 0xc000186780, 0xc0001d3a00, ...)
        /go/pkg/mod/github.com/segmentio/kafka-go@v0.4.15/message.go:140 +0x78
github.com/segmentio/kafka-go.(*Batch).readMessage(0xc0004f8100, 0xc0001d3a30, 0xc0001d3a20, 0xc0001d3a00, 0xc0001d39f0, 0x0, 0x1, 0x1, 0x1, 0x0)
        /go/pkg/mod/github.com/segmentio/kafka-go@v0.4.15/batch.go:228 +0x79
github.com/segmentio/kafka-go.(*Batch).ReadMessage(0xc0004f8100, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/pkg/mod/github.com/segmentio/kafka-go@v0.4.15/batch.go:183 +0x11a
github.com/segmentio/kafka-go.(*reader).read(0xc0001d3ed8, 0x789eb8, 0xc0001403c0, 0x79e519, 0xc00068c8c0, 0x79e50a, 0x0, 0x0)
        /go/pkg/mod/github.com/segmentio/kafka-go@v0.4.15/reader.go:1453 +0x425
github.com/segmentio/kafka-go.(*reader).run(0xc0001d3ed8, 0x789eb8, 0xc0001403c0, 0x79e50a)
        /go/pkg/mod/github.com/segmentio/kafka-go@v0.4.15/reader.go:1272 +0x2b2
[...]

PR #807, opened 2021-12-07
Fix batch offset advance

iddqdeika:
This is a hotfix for  #653, as well as #621

as sarama does - if we got 0 new messages at the end of correctly read batch - then we need to increment batch offset manually.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1475,7 +1475,7 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
 			// if there is no new messages at the end of correct batch -
 			// then we need to increment next batch offset manually.
 			if err == io.EOF && msgCount == 0 {
-				//batch.offset++
+				batch.offset++
 			}
 			batch.Close()
 			break
--- a/reader.go
+++ b/reader.go
@@ -1472,6 +1472,8 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
 		}
 
 		if msg, err = batch.ReadMessage(); err != nil {
+			// if there is no new messages at the and of correct batch -
+			// then we need to advance next batch offset manually.
 			if err == io.EOF && msgCount == 0 {
 				batch.offset++
 			}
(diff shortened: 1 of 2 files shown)


iddqdeika:
@nlsun sure, you mean unit test? I'll try to add such test case.

case is that in our FetchRequest we set offset = X, but all messages in returned RecordBatch has offset < X.

in this case at the end of RecordBatch parsing we have old offset in batch struct, leadeng to cycling this stem infinitely.

Went away to prepare unit test =)

iddqdeika:
also skipped messages in ResultBatches documented in 
5.3.1 Record Batch paragraph on https://kafka.apache.org/documentation/#messageformat

iddqdeika:
@nlsun 
this maneuver costed us 8 hours of tarot divination, but job is done.
without fix in reader test fails (not always, because of compaction trigger time lag).

nlsun:
amazing work! i'm gonna see if I can trim down the test case's runtime/consistency and then dive deeper to understand the issue myself in order to compare solutions (this was also proposed https://github.com/segmentio/kafka-go/pull/709)

iddqdeika:
secret is in kafkas on disk messages storing.
kafka splits messages into log segments (files) and in single RecordSet can be only messages from single segment(file)

if deleted (compacted) message is at the end of segment - then fetch request returns recordset from this segment. and, of cource, deleted message would be skipped.

for example we have messages with offsets
0, 1, 2, 3, 4, 5, 6.
if kafka desires to split them into 2 segments (0,1,2,3), (4,5,6) and message 3 and 6 has the same key - then after compaction with offset 0 kafka returns batch with 0,1,2 messages. after this batch kafka-go tries to fetch message with 2+1 offset = 3, but it is last message in first segment, thats why kafka just returns batch with 0,1,2 offsest once more, and we need to increment offset once more to fetch truly new offsets.

hope this would help you =)

iddqdeika:
I saw #709, but I found that implementation too complicated...

nlsun:
This PR has been incorporated into https://github.com/segmentio/kafka-go/pull/813

PR #813, opened 2021-12-18
Fix batch offset when a batch has 0 messages

nlsun:
Based off of https://github.com/segmentio/kafka-go/pull/807 and also related to https://github.com/segmentio/kafka-go/pull/709

```
Move the fix inside `Batch.readMessage`

The reasoning is that `Batch.readMessage` is normally where we update
the batch offset so the goal is to do this additional update to the
batch offset in the same location.

Also adjusts the test case retry so that the test case may finish
earlier than the previous sleep time while also giving the test more
time to finish to reduce flakiness.

The test case also now ensures that the final state is fully compacted
before declaring success.
```

The solution has been updated to jump to the last offset as indicated by the response header. In addition a test is added to ensure that the new way of moving offsets still works when the message is truncated.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -245,11 +245,11 @@ func (batch *Batch) readMessage(
 			// consumed or a batch whose connection is in an error state.
 			batch.err = dontExpectEOF(err)
 		case batch.msgs.remaining() == 0:
-			// Log compaction can create batches with 0 messages.
+			// Log compaction can create batches with 0 unread messages.
 			//
 			// If the "next offset" reaches the "originally requested offset"
-			// and we have 0 messages remaining, then there were 0 messages in
-			// the batch.
+			// and we have 0 messages remaining, then there were 0 unread
+			// messages in the batch.
 			//
 			// We normally set the batch offset to the "next" batch offset upon
 			// reading a message but since there were no messages to read we
--- a/batch.go
+++ b/batch.go
@@ -190,6 +190,8 @@ func (batch *Batch) ReadMessage() (Message, error) {
 			return
 		},
 	)
+	// A batch may start before the requested offset so skip messages
+	// until the requested offset is reached.
 	for batch.conn != nil && offset < batch.conn.offset {
 		if err != nil {
 			break
@@ -243,6 +245,19 @@ func (batch *Batch) readMessage(
 			// consumed or a batch whose connection is in an error state.
 			batch.err = dontExpectEOF(err)
 		case batch.msgs.remaining() == 0:
+			// Log compaction can create batches with 0 messages.
+			//
+			// If the "next offset" reaches the "originally requested offset"
+			// and we have 0 messages remaining, then there were 0 messages in
+			// the batch.
+			//
+			// We normally set the batch offset to the "next" batch offset upon
+			// reading a message but since there were no messages to read we
+			// update it now instead.
+			if batch.offset == batch.conn.offset {
+				batch.offset++
+			}
+
 			// Because we use the adjusted deadline we could end up returning
 			// before the actual deadline occurred. This is necessary otherwise
 			// timing out the connection for real could end up leaving it in an
--- a/batch.go
+++ b/batch.go
@@ -266,17 +266,20 @@ func (batch *Batch) readMessage(
 			err = checkTimeoutErr(batch.deadline)
 			batch.err = err
 
-			// Checks:
-			// - `batch.err` for a "success" from the previous check
+			// Checks the following:
+			// - `batch.err` for a "success" from the previous timeout check
 			// - `batch.msgs.lengthRemain` to ensure that this EOF is not due
 			//   to MaxBytes truncation
-			// - `batch.lastOffset` to ensure that the message format allows
-			//   compaction
+			// - `batch.lastOffset` to ensure that the message format contains
+			//   `lastOffset`
 			if batch.err == io.EOF && batch.msgs.lengthRemain == 0 && batch.lastOffset != -1 {
-				// Log compaction can create batches with 0 unread records.
+				// Log compaction can create batches that end with compacted
+				// records so the normal strategy that increments the "next"
+				// offset as records are read doesn't work as the compacted
+				// records are "missing" and never get "read".
 				//
 				// In order to reliably reach the next non-compacted offset we
-				// jump to the saved lastOffset.
+				// jump past the saved lastOffset.
 				batch.offset = batch.lastOffset + 1
 			}
 		}
--- a/batch.go
+++ b/batch.go
@@ -28,6 +28,15 @@ type Batch struct {
 	offset        int64
 	highWaterMark int64
 	err           error
+	// The last offset in the batch.
+	//
+	// We use lastOffset to skip offsets that have been compacted away.
+	//
+	// We store lastOffset because we get lastOffset when we read a new message
+	// but only try to handle compaction when we receive an EOF. However, when
+	// we get an EOF we do not get the lastOffset. So there is a mismatch
+	// between when we receive it and need to use it.
+	lastOffset int64
 }
 
 // Throttle gives the throttling duration applied by the kafka server on the
@@ -227,10 +236,12 @@ func (batch *Batch) readMessage(
 		return
 	}
 
-	offset, timestamp, headers, err = batch.msgs.readMessage(batch.offset, key, val)
+	var lastOffset int64
+	offset, lastOffset, timestamp, headers, err = batch.msgs.readMessage(batch.offset, key, val)
 	switch err {
 	case nil:
 		batch.offset = offset + 1
+		batch.lastOffset = lastOffset
 	case errShortRead:
 		// As an "optimization" kafka truncates the returned response after
 		// producing MaxBytes, which could then cause the code to return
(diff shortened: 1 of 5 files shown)


nlsun:
@iddqdeika I made a few tweaks to your original PR https://github.com/segmentio/kafka-go/pull/807, thoughts on the changes?

achille-roussel (approved):
Spectacular 🙌 

Thanks for your thorough investigation into this issue @nlsun and the fix looks great as well 👍

reader_test.go
```diff
@@ -1669,3 +1671,159 @@ func TestErrorCannotConnectGroupSubscription(t *testing.T) {
 			"must fail when it cannot connect")
 	}
 }
+
+// Tests that the reader can read zero message batches from log compacted
+// topics.
+//
+// This test forces varying sized chunks of duplicated messages along with
+// configuring the topic with a minimal `segment.bytes` in order to
+// guarantee that at least 1 batch can be compacted down to 0 messages.
```
iddqdeika, line 1680: It seems to me that there cannot be batch without messages. Our case here is that can be batch without **_new_** (unread by current conn) messages at the end of the batch. If kafka's segment was compacted down to 0 messages, it would be skipped during Fetch and kafka would try to return batch from next segment.

Test case here leads to segment been compacted to at least 1 message, but not at the end of segment. I tried to create fully compacted segments and they were skipped by kafka during fetch..

iddqdeika:
@nlsun thanks! Finally it looks great =) I'm proud to be the part of this fix
 
We're waiting very much for fix to be merged, because we need compaction in upcoming features (in our proj) =D

leedavis81:
@nlsun just wanted to confirm that this does resolve the initial issue I reported in #709 however I do have an instance where a batch has multiple compacted records within it, and the consumer fails to advanced. 

You can see from the headers here I have a record count of 1, but a `lastOffsetDelta` of 3...

```headers  {firstOffset:33268272 length:7160 crc:-909961870 magic:2 v1:{attributes:0 timestamp:0} v2:{leaderEpoch:59 attributes:0 lastOffsetDelta:3 firstTimestamp:1636184302331 lastTimestamp:1636184302331 producerID:-1 producerEpoch:-1 baseSequence:0 count:1}}```

how difficult would it be to accommodate for this scenario in this change?

File: batch.go [hunk 2]
```diff
@@ -243,6 +245,19 @@ func (batch *Batch) readMessage(
...
+			//
+			// If the "next offset" reaches the "originally requested offset"
+			// and we have 0 messages remaining, then there were 0 messages in
+			// the batch.
+			//
+			// We normally set the batch offset to the "next" batch offset upon
+			// reading a message but since there were no messages to read we
+			// update it now instead.
+			if batch.offset == batch.conn.offset {
+				batch.offset++
```
leedavis81, line 258: Where a batch contains two compacted records at its tail advancing the offset by one will cause a subsequent read to retrieve the same batch (as we've not fully passed it). When that batch is processed a second time, it will read a valid record (no error will be returned) and our batch.offset will be moved backwards as we fall into the `case nil:` and the following is applied: `batch.offset = offset + 1`.

leedavis81 (changes requested):

batch.go
leedavis81: One other option could be to only advance the `batch.offset` under the `err = nil` case where <= to the returned offset (meaning all `batch.offset++` increments in the `errShortRead` case will e preserved)

```
		if batch.offset <= offset {
			batch.offset = offset + 1
		}
```

nlsun: @leedavis81 If I understand correctly, we only enter the `errShortRead` state when we reach the end of a batch so advancing the offset by 1 should advance us to the next batch.

I think (not certain) that compacted records in the middle of a batch will only show up in the offset delta and won't cause a `errShortRead`.

I wonder if this new scenario is a different one? Are there any more details you could dig up about where the reader is getting stuck and what the kafka log segment looks like? I tried playing around with the unit test in this PR by increasing the segment size and also varying the duplicated messages but haven't been able to get stuck anywhere else yet.

leedavis81: So the scenario you want is a batch with 3 records, where two are compacted at the end (obviously need to be differing keys). What happens here is the consumer reads record 1 OK, then moves onto record 2 where an errShortRead is triggered and you fall into the case where no records remain and the batch.offset is advanced (as per the fix in this PR).

As no messages reside in the batch, attempts to read batch+3 fail and a request is made to the broker for that offset. The broker then returns the same batch (as that's where the compacted entry resides) and the consumer begins to read that (still with batch.offset set to offset+3). The first read succeeds and returns a new offset with no error, falling into the err = nil case the batch.offset is then moved back to offset+1 and the cycle repeats.

reader_test.go
nlsun: thanks for the clarification, made some updates to the comments

batch.go
leedavis81: @nlsun this is a fix I used today to resolve the production issues we were experiencing.. https://github.com/leedavis81/kafka-go/pull/1/files

iddqdeika: last successfully read offset would be returned incremented by one after batch parsing.
not batch offset.

we only have a problem when there is compacted messages at the end of batch, leading us to have no new messages.
if we have no new messages - we have no new offset.

BUT if we increment offset manually - that case would be avoided.

iddqdeika: > atch.offset is advanced (as per the fix in this PR).

no. if we already read at least 1 message - then batch.offsed is not equal to batch.conn.offset.

we only advanse offset if there is no new messages in batch. and that incremented offset will be saved in batch.conn during batch.close() call.

thats why:
1) send Fetch req with offset 1
2) got batch with offset 1 and no more
3) set batch.offset to 1+1 and during `batch.close()` set `conn.offset = batch.offset`
4) send Fetch req with offset 2
5) got no new messages, leading to `batch.offset == batch.conn.offset` and `batch.offset` would be incremented by 1 to  3
6) during batch.close() call conn.offset will become 3 as well
7) send Fetch req with offset 3
8) got no new messages, move batch.offset to 4
9) PROFIT! =)

nlsun: It sounds to me like there's yet another condition that is causing a `errShortRead` that is currently unhandled that is not the "read 0 new records in a batch due to compacted records" being handled in this fix.

Are we getting an `errShortRead` for a reason besides reaching the end of a batch? My understanding is that compacted records don't actually show up as a message and are just reflected in the next non-compacted message having a higher offset delta

leedavis81: Yeah that's my understanding too, where you have either an empty batch (record is compacted) or a _single_ trailing compacted record in a batch, then this fix will be fine.

There is however a scenario where you could have multiple compacted records at the end of the batch, in which this wouldn't solve as we would never increment past the last (or second compacted) entry (based on the description I detailed above).

nlsun: Ok, so I've got a heavily modified branch with a lot of print statements that has an improved test that tests varying lengths of consecutive compacted records. From reading the print statements I'm pretty confident that we can successfully cross multiple compacted records.

branch: https://github.com/segmentio/kafka-go/compare/fix-batch-offset-advance-nlsunsuggestion-multicompact-test?expand=1 and you can run the test with
```
go test -run TestReaderReadCompactedMessage -count 1 ./...
```

nlsun: @leedavis81 what do you think about tracking this additional bug in a separate github issue? That way we can merge this specific fix for those that need it.

My understanding is that the core issue is if something causes us to stop reading a batch then the `batch.offset` will be reset to whatever offset kafka sends back when starting a new connection. We are now handling the case where the end of a batch causes us to stop reading, but it seems like you are running into an additional issue that is causing the reader to exit as well.

It could be that we simply want to prevent the `batch.offset` from going backwards in which case the workaround in your branch would be best, which would be good to discuss in a separate issue

leedavis81: Your understanding is spot on, and moving that to a seperate issue is perfectly fine. I do intend on testing your branch shortly, just haven't had the time to do it yet.

nlsun: So I was talking to someone about this and they got me thinking about `lastOffsetDelta` again, and I remembered that in the debug branch it was always `1` which I tried to change before but gave up before. This time I tried changing the `BatchSize` on the writer from `2` to `3` and bam it got stuck on what hopefully is what you're running into, will let you know when I fix this.

iddqdeika:
seems this pull will never get merget this way =(

this pushes us to use sarama (which simply counts new messages from batch), but sarama's API makes me sad...

nlsun:
@iddqdeika if you are ok with using a branch you can make one using the solution in https://github.com/segmentio/kafka-go/compare/fix-batch-offset-advance-nlsunsuggestion-multicompact-test?expand=1 (you'll want to remove the print statements), I'll try and finish the regression test for that branch this week but I can't promise it

You can alternatively make a branch with the other suggested solution as a temporary solution https://github.com/leedavis81/kafka-go/pull/1/files

batch.go
nlsun: @leedavis81 alright, hoping the million-th time's the charm eh? Could you try out this newly updated PR on your test case?

leedavis81: i'm running this test now, should have some feedback shortly

leedavis81: Hi @nlsun , I can confirm this patch resolved both instances where we had single and multiple compacted records at the tail of a batch. Nice work! 

This fix would close out https://github.com/segmentio/kafka-go/pull/709 (and a handful of other PRs by the looks of it)

achille-roussel (approved):
Really nice work, and the implementation is neat.

Ship it!

nlsun:
@leedavis81 thanks for your patience and all the help testing
@iddqdeika thanks for your patience and providing the unit test
@achille-roussel thanks for the reviews and suggestions

PR #816, opened 2021-12-30
performance,typo: prealloc slice for performance and fix typoe.

xiekeyi98:


achille-roussel (approved):
Looking good, thanks for the improvements!

PR #820, opened 2022-01-07
Skip refresh metadata for errored topic, fix #806

### LINKED ISSUE #806: WriteMessages function will block the goroutine while topic not exist and  client no permission to create topic
**Describe the bug**
When the topic does not exist and the client does not have permission to create the topic, the WriteMessages function will block indefinitely.

**Kafka Version**
v1.1.0

**To Reproduce**
Steps to reproduce the behavior.  Bonus points for a code sample.
1) Topic does not exist
2) Client does not have permission to create topic
3) Topics are automatically created when the client writes a message
4) The goroutine will be blocked permanently without returning an error
5) This problem exists in v0.4.23 and later versions

**Expected behavior**
WriteMessages function will not block the goroutine and get errors return while client no permission to create topic

1000Delta:
Skip refresh metadata for errored topic, fixed the problem metadata refresh indefinitely, 
such as write message to a topic that not exist, but kafka disabled topic auto creation.

Fixes #806

Refer to the implementation of #673

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -416,7 +416,18 @@ func (p *connPool) roundTrip(ctx context.Context, req Request) (Response, error)
 		// we didn't have that topic in our cache so we should update
 		// the cache.
 		if m.AllowAutoTopicCreation {
-			p.refreshMetadata(ctx, m.TopicNames)
+			topicsToRefresh := make([]string, 0, len(resp.Topics))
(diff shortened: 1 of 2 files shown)


achille-roussel (approved):
Thanks for investigating this issue and submitting a fix!

The changes look good and well tested 👍

Issue #818, opened 2022-01-04
periodic metadata request always sent with Include{Cluster,Topic}AuthorizedOperations set to true


dnwe:
**Describe the bug**
In the periodic metadata refresh that [kafka-go](/segmentio/kafka-go) caches in transport.go, it appears to ask Kafka to also determine and iterate ACLs in its response by hardcoding `IncludeClusterAuthorizedOperations` and `IncludeTopicAuthorizedOperations` both to `true` https://github.com/segmentio/kafka-go/blob/e1f5945be0b2e606b6826e9e5b0ec0a9ce68bd93/transport.go#L588-L591 

However, kafka-go doesn't appear to actually make use of either the `ClusterAuthorizedOperations` nor the per-topic `TopicAuthorizedOperations` anywhere in its code, so this is causing unecessary load on the brokers as this is a non-trivial thing for them to determine each time.

**Kafka Version**
2.8.1

**To Reproduce**
Any use of kafka-go exhibits this behaviour.

**Expected behavior**
As a first change, kafka-go should hardcode these both to `false` by default as currently there is no need for them.
As an extension, the `roundTrip` code should perform a similar `requestNeeded` check as is done for the AllowAutoTopicCreation whereby if a metadata.Request arrives in roundTrip that has `ClusterAuthorizedOperations` or `TopicAuthorizedOperations` explicitly set to true (e.g., from a user's DescribeXYZ admin function) then it will send the request to Kafka rather than serving the Metadata response from the cache.

**Additional context**
N/A

PR #821, opened 2022-01-10
fix: omit {Cluster,Topic}AuthorizedOperations

dnwe:
kafka-go doesn't appear to actually make use of either the
ClusterAuthorizedOperations nor the per-topic TopicAuthorizedOperations
responses anywhere in its code, so this is causing unecessary load on
the brokers as this is a non-trivial thing for them to determine each
time.

Contributes-to: #818


achille-roussel (approved):
The change looks good, and probably a simpler way to address https://github.com/segmentio/kafka-go/pull/767

Thanks for submitting!

Issue #814, opened 2021-12-22
Regression after upgrading from v0.4.23: unexpected EOF


Reasno:
**Describe the bug**
If the topic has no messages, the kafka reader starts to report "the kafka reader got an unknown error reading partition x of my-topic at offset y: unexpected EOF"

**Kafka Version**
2.3.0

**To Reproduce**
Run kafka reader in consumer group mode when the partition has no more messages.

**Expected behavior**
No unexpected EOF reported, as pre 0.4.24

**Additional context**
The problem only occurs in v0.4.24 and above (tested with v0.4.23, v0.4.24, v0.4.25). I believe the regression is caused by PR #788 where the highwater mark check is removed just to pass a test. The check seems to be added intentionally in prior PRs.

edit: fix version number

Gerifield:
We had a similar error that disappeared as soon as we reverted the version from 0.4.25 to 0.4.23.

lixianmin:
yes, found the same error, and reverted to 0.4.23 solved this problem

rhansen2:
Hello, thanks for submitting this issue!

Is anyone able to provide a reproduction of this issue? I haven't had any success in replicating it.

marselester:
> Is anyone able to provide a reproduction of this issue? I haven't had any success in replicating it.

@rhansen2 here you go https://gist.github.com/marselester/a7cc16279e990a8d42c41fd9f557b341.

rhansen2:
Thanks for the reproduction @marselester For now I'm going to revert #788

rhansen2:
The revert of #788 is available in https://github.com/segmentio/kafka-go/releases/tag/v0.4.26. Based on the reproduction provided by @marselester I believe that this regression should now be fixed.

PR #823, opened 2022-01-12
don't attempt to read messages when offset is greater than hwm

rhansen2:
Based on
> The high watermark offset is the offset of the last message that was
successfully copied to all of the log’s replicas.

the high watermark is the last message that was copied to all of the
logs replicas and consumers should not read past that message.

Previously we stopped reading messages when offset == highWaterMark
which led to messages being skipped when the offset was equal to the
highWaterMark.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -857,7 +857,11 @@ func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
 
 	var msgs *messageSetReader
 	if err == nil {
-		msgs, err = newMessageSetReader(&c.rbuf, remain)
+		if highWaterMark < offset {
+			msgs = &messageSetReader{empty: true}
+		} else {
+			msgs, err = newMessageSetReader(&c.rbuf, remain)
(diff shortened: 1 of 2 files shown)


rhansen2:
Potentially fixes #814

achille-roussel (approved):
+1 on getting someone to validate that it resolves the `io.ErrUnexpectedEOF` errors, the change looks meaningful tho.

marselester:
Thank you for working on it. I've tried this fix, the issue still persists.

conn.go
```diff
@@ -857,7 +857,11 @@ func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
 
 	var msgs *messageSetReader
 	if err == nil {
-		msgs, err = newMessageSetReader(&c.rbuf, remain)
+		if offset == highWaterMark {
```
rhansen2, line 860: HWM is the offset of the last durable message +1

PR #824, opened 2022-01-13
Revert "Nettest fix"

rhansen2:
Revert removal of high water mark check. Although this fixes nettest it's not the correct behavior.
Reverts segmentio/kafka-go#788


rhansen2:
Revert to fix #814

PR #826, opened 2022-01-17
add TryReadMessage and TryFetchMessage

smallnest:


achille-roussel (commented):
Have you considered alternative approaches to introducing new APIs? (e.g. configuring the reader in non-blocking mode)

smallnest:
> Have you considered alternative approaches to introducing new APIs? (e.g. configuring the reader in non-blocking mode)

Yes. Add a config option is another solution but it may change ReadMessage's meaning and affect exited productions.

Issue #819, opened 2022-01-07
`partitionWriter` has unlimited batch queue size.


sko00o:
**Describe the bug**
I made a [benchmark](https://github.com/sko00o/benchmark-kafka-go-clients) for 3 kafka go clients.
As a result, I noticed that kafka-go is the fastest producer. But it cost too much memory comparing with other two clients.

**Kafka Version**
kafka:2.12-2.4.0

**To Reproduce**
I set `Writer.Async = true` for better produce performance. Here is my writer [config](https://github.com/sko00o/benchmark-kafka-go-clients/blob/main/client_segmentio.go#L67):

```go
	producer := kafka.Writer{
		Addr:         kafka.TCP(brokers),
		Topic:        topic,
		Balancer:     &kafka.Hash{},
		BatchTimeout: BatchTimeout,
		BatchSize:    BatchSize,
		BatchBytes:   BatchBytes,
		RequiredAcks: kafka.RequireNone,
		Async:        true,
	}
```

Then run following test for producer

```sh
./kafka-benchmark -numMessages 10000000 -msgSize 512 -client kafka-go -mode producer
```

**Expected behavior**
Limit memory usage in async writer.

**Additional context**
I notice that `writer.WriteMessages` will create `partitionWriter` for each topicPartition. 
And `partitionWriter` will hold a `writeBatch` queue for batch process.
Following related code:
* https://github.com/segmentio/kafka-go/blob/37ab7b1b613c78b5edac7f76bb174b6ce2130752/writer.go#L648
* https://github.com/segmentio/kafka-go/blob/37ab7b1b613c78b5edac7f76bb174b6ce2130752/writer.go#L696
* https://github.com/segmentio/kafka-go/blob/37ab7b1b613c78b5edac7f76bb174b6ce2130752/writer.go#L969

`partitionWriter.queue` is a slice with unlimited length. Batch messages in queue will increase
while `queue.Put()` much more than `queue.Get()`.

In my opinion, I would like to have a config option to limit this batch queue length.
So I try to use a buffered channel implement a fixed size queue for this.

3AceShowHand:
why not used a buffered channel ?

PR #827, opened 2022-01-19
feature: limit batch queue size in partitionWriter.

sko00o:
kafka-go writer will create queues for batch messages, but there is no limit on length, batch queue will increase as fast as `queue.Put()` called.

Contributes-to: #819

Signed-off-by: Wenyu Li <lwylwy@outlook.com>

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -100,6 +100,11 @@ type Writer struct {
 	// The default is to try at most 10 times.
 	MaxAttempts int
 
+	// Limit on how many batches will be buffered in a partitionWriter.
+	//
+	// Defaults to 0, means unlimited queue size.
+	BatchQueueSize int
+
 	// Limit on how many messages will be buffered before being sent to a
 	// partition.
 	//
@@ -797,6 +802,13 @@ func (w *Writer) batchTimeout() time.Duration {
 	return 1 * time.Second
 }
 
+func (w *Writer) batchQueueSize() int {
+	if w.BatchQueueSize > 0 {
+		return w.BatchQueueSize
+	}
+	return 0
+}
+
 func (w *Writer) readTimeout() time.Duration {
 	if w.ReadTimeout > 0 {
 		return w.ReadTimeout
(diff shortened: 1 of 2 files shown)


achille-roussel:
Hello @sko00o 

I reviewed your benchmark code and I am a little concerned about modifying the API footprint of kafka-go to optimize for this use case:

I am curious whether the benchmark to put together is representative of the type of workloads that you will be seeing in production? It seems that you are writing all messages as fast as you can, one at a time from a single goroutine. In my experience, typical stream processors that rely on kafka tend to have more tamed load patterns, or use throttling mechanisms that limit the throughput of producers.

I want to make sure that new APIs we introduce are effective at supporting production use case, there is a long term maintenance cost we need to account for, which is why I am asking.

---
An alternative we discussed before with the team would be to have a `Flush` method on the `Writer` type, so clients have more control over when they choose to flush batches; do you think it would be useful in your case as well to limit memory utilization?

sko00o:
Thanks for reply @achille-roussel 

> I am curious whether the benchmark to put together is representative of the type of workloads that you will be seeing in production? 

I'm working on some streaming log process projects that use Kafka as sink target. And I would call `writer.WriteMessages` in many goroutines.
I'm not sure that using throttling mechanisms before calling `WriteMessages` will solve my issue. There is no method to ensure that batch messages have been sent out.

> An alternative we discussed before with the team would be to have a `Flush` method on the `Writer` type, so clients have more control over when they choose to flush batches; do you think it would be useful in your case as well to limit memory utilization?

It would be useful to have `Flush` method, but the problem will be more complex. We already control the `BatchTimeout` , `BatchBytes` and `BatchSize`, maybe we don't have to care about when to call `Flush`. We have to set interval or max queue size for auto `Flush` call. I think blocking on `queue.Put` is easier.

achille-roussel:
> I think blocking on queue.Put is easier.

Would it make sense to configure the writer in blocking mode then? (with Async=false)

Receiving writes from multiple goroutines and flushing when batches fill up is the common use case that the `kafka.Writer` type is usually used to solve for.

sko00o:
> Would it make sense to configure the writer in blocking mode then? (with Async=false)

I create a benchmark for sync writer on multiple goroutines. https://github.com/sko00o/benchmark-kafka-go-clients/commit/54ea71dae737c51930f8075fcd2dde6ada006164 (I can limit the memory usage by limit the goroutines on write process. But it seems too slow.)

And I also create a benchmark for block batch queue based on my PR. https://github.com/sko00o/benchmark-kafka-go-clients/commit/b4f252baa337cd8238f1e6101e6d14c6924f34e0

achille-roussel:
Do you have any details to share about your use cases?

I see you are making use of requiredAcks=none, which imply that maybe guaranteed deliveries are not a requirement here, maybe there's an option where we allow the `kafka.Writer` to drop messages to limit memory usage rather than blocking when configured in "fire and forget" mode?

sko00o:
Sorry for late response.  `requiredAcks=none` is set for benchmark only here.  we need guarantee for delivery.

Use case here: 
* We need to pull jobs from message queue, the job describe the logging source and process actions.
* After process the logs line by line, we send the final logs to Kafka.

Maybe blocking in "fire and forget" mode is a poor design. I could run slowly on limited memory rather than drop messages before send. So the only solution here is set Async=false and limit the goroutines on writer process, right?

achille-roussel:
Yes, if you need guarantees of message delivery you should use `Async: false` so you application receives feedback on success and failure of writes.

Issue #729, opened 2021-09-01
Request for Admin API to create ACLs with SASL


Mayuresh-Gharat:
We are trying to use this client in order to create topics in Kafka, programmatically. Currently, there does not seem to be a way to create the ACLs programmatically. It would be great if we can use this client to do both create ACLs and topics programmatically. Kafka open-source has an admin API for creating ACLs: https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/admin/KafkaAdminClient.html#createAcls-java.util.Collection-org.apache.kafka.clients.admin.CreateAclsOptions-

achille-roussel:
Hello @Mayuresh-Gharat, thanks for starting the conversion!

kafka-go does not currently expose the CreateAcls API, but this could be easily added to the `kafka.Client` type.

We always welcome contributions, would you be able to submit a pull request to add the APIs you need?

Mayuresh-Gharat:
@achille-roussel  I am pretty new to GoLang and might need some guidance on this. I would be happy to take a stab at this. Would you be able to point me to a starting point in the code that I can follow as an example, to achieve this?

achille-roussel:
Sure thing!

I would start by looking at how other APIs are implemented, for example [`kafka.(*Client).CreateTopics`](https://github.com/segmentio/kafka-go/blob/master/createtopics.go), you should be able the follow these models to implement CreateAcls.

The request and response messages sent to and received from Kafka are declared in the `protocol` sub-packet, for example: https://github.com/segmentio/kafka-go/blob/master/protocol/createtopics/createtopics.go

You can find the reference documentation for the Kafka protocol at https://kafka.apache.org/protocol.html#protocol_api_keys

achille-roussel:
Hello @Mayuresh-Gharat, are you still interested in making this contribution?

Mayuresh-Gharat:
Hi, @achille-roussel sorry I completely forgot about this. I would be happy to hand it over to anyone interested in this. I don't have cycles to work on this right now (even though I would like to).

achille-roussel:
Sounds good, we will close the error for now and put this on the backlog, feel free to reopen if you find the need to contribute these new APIs in the future 👍

PR #839, opened 2022-02-01
Add CreateAcls Admin API support

zirkome:
This adds the `CreateACLs` method to the `Client` struct to supports the `CreateAcls` Admin API.

See https://github.com/segmentio/kafka-go/issues/729
See https://kafka.apache.org/protocol.html#The_Messages_CreateAcls

Signed-off-by: Guillaume Fillon <guillaume.fillon@auth0.com>


resource.go
```diff
@@ -0,0 +1,37 @@
...
+// https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/resource/ResourceType.java
+type ResourceType int8
+
+const (
+	ResourceTypeUnknown ResourceType = 0
+	ResourceTypeAny     ResourceType = 1
+	ResourceTypeTopic   ResourceType = 2
+	ResourceTypeGroup   ResourceType = 3
+	// See https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/config/ConfigResource.java#L36
+	ResourceTypeBroker          ResourceType = 4
```
zirkome, line 12: IMO this should have been under a `ConfigResourceType` type but instead decided to keep it for backward compatibility.

docker-compose.yml
```diff
@@ -21,6 +21,8 @@ services:
       KAFKA_LISTENERS: 'PLAINTEXT://:9092,SASL_PLAINTEXT://:9093'
       KAFKA_ADVERTISED_LISTENERS: 'PLAINTEXT://localhost:9092,SASL_PLAINTEXT://localhost:9093'
       KAFKA_SASL_ENABLED_MECHANISMS: 'PLAIN,SCRAM-SHA-256,SCRAM-SHA-512'
+      KAFKA_AUTHORIZER_CLASS_NAME: 'kafka.security.auth.SimpleAclAuthorizer'
```
zirkome, line 24: Newer Kafka version (>2.4.0) are now using `kafka.security.auth.AclAuthorizer` something to keep in mind if Kafka is upgraded.

zirkome:
@rhansen2 Hi I see that you were assigned to the PR. Is there any I can do to help move this forwards? Thank you!

rhansen2:
Hi @zirkome, 
Thanks for your submission!

Currently this PR is not passing the CI steps due to failing tests. I think the code looks sane so once the tests are fixed-up I think we're good to move forward.

Thanks!

zirkome:
@rhansen2 The tests are not passing because it requires CI changes (i.e. enabling ACL authorizer) which I've made in the PR. But for obvious security reason it doesn't run the tests on this very PR's CI config. Should I make another PR with these changes or this is something you could take care of?

.circleci/config.yml
```diff
@@ -120,6 +120,8 @@ jobs:
     working_directory: *working_directory
     environment:
       KAFKA_VERSION: "2.0.1"
+      KAFKA_AUTHORIZER_CLASS_NAME: 'kafka.security.auth.SimpleAclAuthorizer'
```
rhansen2, line 123: Should these configs be set in the kafka container?

zirkome: I see what you mean but all `environment` are alias from the same yaml anchor and can't be overridden AFAIK. Also depending on the Kafka version `KAFKA_AUTHORIZER_CLASS_NAME` needs to be different so this also can't be added to that one global environment in `kafka-011`.
Please advise.

rhansen2: The `Override Values` section of this article https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/  shows how to extend anchored values. You should be able to to use that syntax to set the authorizer specifically for each version without needed to redefine the entire environment. Lemmie know if that helps thanks!

zirkome: @rhansen2 Fixed it using that syntax, I have no way to test if it's correct though.

rhansen2: You can test by running the circleci jobs locally or by examining the output of the jobs in circleci (I believe you should be able to view the runs for your PR from the details links). It looks like the class name is incorrect for the AclAuthorizer and should be `kafka.security.authorizer.AclAuthorizer`

zirkome: @rhansen2 Indeed good catch on the class name! Will update and see why things fails.

zirkome: @rhansen2 Fixed the test as well as it was blocking the other tests. All green now!

rhansen2 (approved):
This all looks good to me. Thanks for the contribution!

zirkome:
@rhansen2 Thanks for the help and merging the PR. Would you know when will this PR be released? Does kafka-go have a fixed release cadence?

rhansen2:
We don't really have a fixed release cadence. I've cut https://github.com/segmentio/kafka-go/releases/tag/v0.4.29 which contains these changes.

PR #843, opened 2022-02-07
attempt to detect improper TLS config

rhansen2:
when an io.ErrUnexpectedEOF is returned when reading
a correlation ID, attempt to detect if the messages
from the broker was a TLS Alert Message and decorate
the returned error

### DIFF
--- a/protocol/response.go
+++ b/protocol/response.go
@@ -147,9 +147,5 @@ func looksLikeUnexpectedTLS(size int32) bool {
 		return false
 	}
 	version := int(sizeBytes[1])<<8 | int(sizeBytes[2])
-
-	if version <= tls.VersionTLS13 && version >= tls.VersionTLS10 {
-		return true
-	}
-	return false
+	return version <= tls.VersionTLS13 && version >= tls.VersionTLS10
 }
--- a/protocol/response.go
+++ b/protocol/response.go
@@ -1,6 +1,9 @@
 package protocol
 
 import (
+	"crypto/tls"
+	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 )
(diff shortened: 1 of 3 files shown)


protocol/response.go
```diff
@@ -35,6 +37,27 @@ func ReadResponse(r io.Reader, apiKey ApiKey, apiVersion int16) (correlationID i
...
+			// tries to determine if that's what has happened.
+			// We first deconstruct the initial 4 bytes of the message
+			// from the size which was read earlier.
+			// Next, we examine those bytes to see if they looks like a TLS
+			// error message. If they do we wrap the io.ErrUnexpectedEOF
+			// with some context.
+			var sizeBytes [4]byte
+			binary.BigEndian.PutUint32(sizeBytes[:], uint32(size))
+			if looksLikeUnexpectedTLS(sizeBytes) {
+				err = fmt.Errorf("%w: broker appears to be expecting TLS", io.ErrUnexpectedEOF)
```
rhansen2, line 54: Not sure what the best way to phrase this message is.

protocol/response.go
```diff
@@ -109,3 +131,25 @@ func WriteResponse(w io.Writer, apiVersion int16, correlationID int32, msg Messa
...
+
+	if sizeBytes[0] != tlsAlertByte {
+		return false
+	}
+	version := int(sizeBytes[1])<<8 | int(sizeBytes[2])
+
+	if version <= tls.VersionTLS13 && version >= tls.VersionTLS10 {
+		return true
+	}
+	return false
```
achille-roussel, line 154: Could this be simplified?
```suggestion
	return version <= tls.VersionTLS13 && version >= tls.VersionTLS10
```

achille-roussel (approved):
Looking good to me!

PR #844, opened 2022-02-10
Every example with your own module

robsonpeixoto:
If you are using gopls with `build.experimentalWorkspaceModule` enabled the gopls will not start because the module name are being reused for different modules.

This PR will put all examples in your own module.


achille-roussel (approved):
Thanks for the fix!

PR #846, opened 2022-02-12
Improving some internal error-handling

dominicbarnes:
The goal of this PR is to set us up to improve visibility into errors being returned by kafka-go by wrapping those errors before bubbling them up throughout the call stack. The hope is that by making each error into a clear chain, it should be much easier to troubleshoot by narrowing in on the most relevant branch of code that is failing.

To start towards this more lofty goal, the first step (and the main focus of this PR) is to replace "strict" error-checking (eg: `err == io.EOF`) with more modern/flexible checks (eg: `errors.Is(err, io.EOF)`). This will allows us to freely wrap errors internally while not changing any error-handling logic at higher levels. Without this, blindly wrapping errors could result in subtle bugs as the older methods of checking errors would no longer behave the same way.

While I did include _some_ error-wrapping changes, the changes are pretty focused on the error-checking instead, which is already a decent size PR. We can use subsequent PRs to refactor other components (eg: reader, writer) and catch any other error-checking places I may have missed.

### DIFF
--- a/addoffsetstotxn.go
+++ b/addoffsetstotxn.go
@@ -53,7 +53,7 @@ func (c *Client) AddOffsetsToTxn(
 		GroupID:         req.GroupID,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("failed to add offsets to transaction: %w", err)
+		return nil, fmt.Errorf("kafka.(*Client).AddOffsetsToTxn: %w", err)
 	}
 
 	r := m.(*addoffsetstotxn.Response)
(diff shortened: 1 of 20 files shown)


dominicbarnes:
I had made some changes to `conn.go`, but it looks like the `x/net/nettest` package [explicitly expects an unwrapped `net.Error` ](https://cs.opensource.google/go/x/net/+/27dd8689:nettest/conntest.go;l=399) which causes the test to fail when a `fmt.Errorf` is used.

It's possible this is just an expectation of a `net.Conn` interface/contract we can't/shouldn't change, so we may have to tread more carefully there.

achille-roussel (approved):

PR #848, opened 2022-02-15
update lz4 + cleanup dependencies

achille-roussel:
This PR updates does a bit of dependency cleanup:

- update LZ4 dependency from v1 to v4
- remove dependency on github.com/eapache/go-xerial-snappy by copying the code in compress/snappy/go-xerial-snappy; this dependency is used for testing purposes only to validate that our implementation is compatible with the reference
- remove github.com/golang/snappy in favor of github.com/klauspost/compress/snappy, which is a more modern implementation of the snappy compression protocol, and that we already have a dependency on
- update github.com/klauspost/compress from v1.9.8 to v1.14.2


rhansen2 (approved):
LGTM

achille-roussel:
Thanks for the review!

PR #855, opened 2022-02-23
Add kafka-go/records package

achille-roussel:
## kafka-go/records

I'm opening this PR to submit a suggestion to add a new package for higher-level components built on top of the kafka-go client APIs. In a similar way to what we did with the `topics` package, I would like to introduce a new `records` package in kafka-go.

There are two main APIs that I have implemented in this PR:
- `records.Storage` which is an interface that abstracts the storage of opaque values indexed by broker address, topic, partition, and base offset. This interface comes with two implementations, on in memory backed by a Go map, and another using the local file system to record and index the values.
- `records.Cache` which implements a caching logic for responses to fetch requests.

I made changes to the way record batches are decoded to delay decompression of the records in order to allow copying the compressed data directly though the cache to the storage layer without requiring re-encoding.

The README has more extensive documentation on the purposes of those new building blocks.

Let me know if you have any questions or feedback!

## Breaking Changes

As part of working on this change, I would like to introduce a breaking change to the `kafka.RecordReader` interface, let me give some context first.

Currently, the record reader interface has a single method `ReadRecord`. When reading records, the keys and values are abstracted by `kafka.Bytes` interfaces which is an extension to `io.ReadCloser`. This model requires applications to close keys and values (after checking that they weren't nil) for every record that they read. In practice, this resource management is unnecessary due to the way we implement the readers underneath, and results in very verbose programs.

In this PR, I have changed the `kafka.RecordReader` to implement `io.Closer`, and changes `kafka.Bytes` to not require closing anymore. Applications now only need to close the record reader received in fetch responses rather than closing each key and value of each record. This model also more closely resembles more closely the design of `http.Response` in the standard library, where programs need to close the body after consuming it.

Since this is a breaking change, we would likely release this change under a new v0.5 release line as it has been the model we used in the past. While being pre-v1 gives us flexibility to make those types of changes, we need to be cautious about those as they can be costly for users to adapt their applications. I will write a section in the README to document how to migrate from v0.4 to v0.5.


protocol/buffer.go
```diff
@@ -534,31 +580,21 @@ func (ref *pageRef) ReadByte() (byte, error) {
...
+	offset := int64(ref.offset)
+	length := int64(ref.length)
+	if off >= length {
 		return 0, io.EOF
 	}
-
-	if off+int64(len(b)) > limit {
-		b = b[:limit-off]
+	if off+int64(len(b)) > length {
+		b = b[:length-off]
```
rhansen2, line 595: Is this needed because a pageRef could end partially through a page and we would potentially read past the end of the pageRef without this adjustment?

records/time.go
```diff
@@ -0,0 +1,14 @@
...
+
+func Time(t int64) time.Time {
+	return time.Unix(t/1000, (t%1000)*int64(time.Millisecond)).UTC()
+}
+
+func Timestamp(t time.Time) int64 {
+	if t.IsZero() {
+		return 0
+	}
+	return t.UnixNano() / int64(time.Millisecond)
```
chriso, line 13:
```suggestion
	return t.UnixMilli()
```

records/time.go
```diff
@@ -0,0 +1,14 @@
+package records
+
+import "time"
+
+func Time(t int64) time.Time {
+	return time.Unix(t/1000, (t%1000)*int64(time.Millisecond)).UTC()
```
chriso, line 6:
```suggestion
	return time.UnixMilli(t).UTC()
```

protocol/record.go
```diff
@@ -292,11 +296,11 @@ func (rs *RecordSet) WriteTo(w io.Writer) (int64, error) {
 	return n, nil
 }
 
-func makeTime(t int64) time.Time {
-	return time.Unix(t/1000, (t%1000)*int64(time.Millisecond))
+func MakeTime(t int64) time.Time {
+	return time.Unix(t/1000, (t%1000)*int64(time.Millisecond)).UTC()
```
chriso, line 300:
```suggestion
	return time.UnixMillis(t).UTC()
```

records/README.md
```diff
@@ -0,0 +1,81 @@
...
+In such cases, random access patterns are common and the cost of retrieving
+offsets form arbitrary positions within Kafka partitions can be high: fetching
+records over the network introduces latency and read amplification (due to Kafka
+always returning batches of compressed records), and the requested partition
+ranges are likley to have been paged out which increases I/O pressure on the
+Kafka brokers.
+
+For high throughput systems, those constraints result in impractical performance
+degradations. The use of a local cache is an effective solution to reduce the
+Kafka borker access frequency.
```
chriso, line 22:
```suggestion
Kafka broker access frequency.
```

records/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Caching Records: [records.Cache](https://pkg.go.dev/github.com/segmentio/kafka-go/records#Cache)
+
+Applications using Kafka typically consume partitions sequentially from the
+first to the last offsets, in which case the data is always new. In some cases,
+clients might be designed to leverage Kafka as more than a pub/sub system, using
+the ability to reference records by offset as a general purpose storage solution.
+
+In such cases, random access patterns are common and the cost of retrieving
+offsets form arbitrary positions within Kafka partitions can be high: fetching
```
chriso, line 14:
```suggestion
offsets from arbitrary positions within Kafka partitions can be high: fetching
```

records/README.md
```diff
@@ -0,0 +1,81 @@
...
+records over the network introduces latency and read amplification (due to Kafka
+always returning batches of compressed records), and the requested partition
+ranges are likley to have been paged out which increases I/O pressure on the
```
chriso, line 17:
```suggestion
ranges are likely to have been paged out which increases I/O pressure on the
```

records/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+cache := &records.Cache{
+    SizeLimit: 64 * 1024 * 1024 * 1024, // 64GiB
+    Storage:   storage,
+}
+...
+```
+
+Record batches are stored as they were returned by Kafka in response to the
+fetch requests. Besides being compute, space an bandwidth efficient (leveraging
```
chriso, line 75:
```suggestion
fetch requests. Besides being compute, space and bandwidth efficient (leveraging
```

records/cache.go
```diff
@@ -0,0 +1,521 @@
...
+	"github.com/emirpasic/gods/trees/redblacktree"
+	"github.com/segmentio/kafka-go"
+	"github.com/segmentio/kafka-go/protocol"
+	"github.com/segmentio/kafka-go/protocol/fetch"
+)
+
+// Key is a comparable type representing the identifiers that record batches
+// can be indexed by in a storage instance.
+type Key struct {
+	Addr            string
```
chriso, line 20: Would a leader change have any effect here?

records/cache.go
```diff
@@ -0,0 +1,521 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func (c *Cache) storeTopicPartitionRecords(ctx context.Context, addr, topic string, partition *fetch.ResponsePartition) (inserts, evictions int, sizeDelta int64, err error) {
+	var inserted bool
+	// TODO: should we cache control batches?
```
chriso, line 353: Where can I read more about control batches?

chriso (approved):

protocol/buffer.go
achille-roussel: Potentially. This is mostly to satisfy the `io.ReaderAt` interface I believe. I revisited the implementation to simplify it a bit when I was debugging.

records/time.go
achille-roussel: `UnixMilli` was added in 1.17, kafka-go still maintains compatibility with stable Go versions, so 1.15 at this time (and 1.16 when 1.18 is released), so unfortunately I don't think we can use this API yet.

records/cache.go
achille-roussel: It shouldn't, the intent is for this field to be the cluster address.

records/cache.go
achille-roussel: https://kafka.apache.org/documentation/#controlbatch

records/time.go
chriso: Ah k good to know.

Issue #864, opened 2022-03-18
Explicitly perform tls handshake in Transport to report better errors.


rhansen2:
Currently many errors end up bubbling only an UnexpectedEOF. This makes debugging issues very difficult. Let's update the transport to explicitly perform the TLS handshake so we can decorate the errors with details.

**Supporting documentation**
The existing Conn type already does this.

PR #866, opened 2022-03-18
explicitly perform TLS handshake on connGroup connections

rhansen2:
Addresses #864 and brings more parity to the Dialer and the Transport.


rhansen2:
We've decided to hold off on this for the time being.

Issue #865, opened 2022-03-18
Implement a wrapper for TLS Conns to add details to errors.


rhansen2:
There are situations where a tls.Conn only reports an EOF which eventually bubbles up to the caller as an UnexpectedEOF. This error can occur for many reasons and makes debugging difficult. Let's decorate our TLS connections to allow us to add more detailed information to error messages.

PR #867, opened 2022-03-18
TLS error enhancements

rhansen2:
Addresses #865


achille-roussel:
We're gonna try to put this under a debug tag so it doesn't modify the behavior of production workloads.

transport.go
```diff
@@ -1126,6 +1126,33 @@ func (g *connGroup) releaseConn(c *conn) bool {
...
+	errch := make(chan error, 1)
+
+	go func() {
+		defer close(errch)
+		errch <- conn.Handshake()
+	}()
+
+	select {
+	case <-ctx.Done():
+		err = ctx.Err()
```
achille-roussel, line 1145: Should we always wait for the goroutine to terminate before returning?
```suggestion
		err = ctx.Err()
		<-errch
```

PR #871, opened 2022-03-22
Fix logger example

route:
Since `msg` is formatted we should call `Printf` because `Println` doesn't support variadic arguments


rhansen2 (approved):
Thanks for the contribution!

PR #873, opened 2022-03-27
Add the field to avoid creating missing topic.

### LINKED ISSUE #872: Adjust `Writer` in order to explicitly set `AllowAutoTopicCreation` for the `WriteMessages` method
**Status Quo**

`Writer`'s method [`WriteMessages`](https://github.com/segmentio/kafka-go/blob/6b6b5bd764c9d2db34b6406631c80802424e48aa/writer.go#L591) calls the method [`partitions`](https://github.com/segmentio/kafka-go/blob/6b6b5bd764c9d2db34b6406631c80802424e48aa/writer.go#L726) which uses the transport to [fetch](https://github.com/segmentio/kafka-go/blob/6b6b5bd764c9d2db34b6406631c80802424e48aa/writer.go#L734-L737) topic's partition. The transport sends a [request](https://github.com/segmentio/kafka-go/blob/6b6b5bd764c9d2db34b6406631c80802424e48aa/protocol/metadata/metadata.go#L9-L14) with the `AllowAutoTopicCreation` field hardcoded to `true`. It leads to undesired side-effect when the client is trying to create missing topic prior to publishing messages.

**Describe the solution you'd like**

- `AllowAutoTopicCreation` is a field of the `Writer` struct to enable for explicit configuration of client behaviour.

- Transport's [request payload](https://github.com/segmentio/kafka-go/blob/6b6b5bd764c9d2db34b6406631c80802424e48aa/writer.go#L736) relies on the user defined configuration.

**Supporting documentation**

- [`allow.auto.create.topics`](https://kafka.apache.org/documentation/#consumerconfigs_allow.auto.create.topics) of the Kafka _Consumer_ API.

**_Note!_** There is no `allow.auto.create.topics` configuration available for the Kafka Producer API. It looks to be a particular decision of the library creators/maintainers.

kislerdm:
Add field AllowAutoTopicCreation to Writer struct.
It prevents the side-effect of creating missing topic.

Closes #872

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ if err := w.Close(); err != nil {
 
 ```go
 // Make a writer that publishes messages to topic-A.
-// The topic is being if missing.
+// The topic is being created if missing.
 w := &Writer{
     Addr:                   TCP("localhost:9092"),
     Topic:                  "topic-A",
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ if err := w.Close(); err != nil {
 
 ```go
 // Make a writer that publishes messages to topic-A.
-// The topic is being created if missing.
+// The topic will be created if it is missing.
 w := &Writer{
     Addr:                   TCP("localhost:9092"),
     Topic:                  "topic-A",
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ for i := 0; i < retries; i++ {
     ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
     defer cancel()
     
-    // an attempt to creating topic prior to publishing the message is being made.
+    // attempt to create topic prior to publishing the message
     err = w.WriteMessages(ctx, messages...)
     if errors.Is(err, LeaderNotAvailable) || errors.Is(err, context.DeadlineExceeded) {
         time.Sleep(time.Millisecond * 250)
--- a/README.md
+++ b/README.md
@@ -337,6 +337,7 @@ to use in most cases as it provides additional features:
 - Synchronous or asynchronous writes of messages to Kafka.
 - Asynchronous cancellation using contexts.
 - Flushing of pending messages on close to support graceful shutdowns.
+- Creation of a missing topic before publishing a message. *Note!* it was the default behaviour up to the version `v0.4.30`.
 
 ```go
 // make a writer that produces to topic-A, using the least-bytes distribution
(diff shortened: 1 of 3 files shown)


writer.go
```diff
@@ -185,6 +185,9 @@ type Writer struct {
 	// If nil, DefaultTransport is used.
 	Transport RoundTripper
 
+	// AllowAutoTopicCreation notifies writer to create topic is missing.
```
achille-roussel, line 188: While it's not a breaking change per-say since the code will continue to compile, the new behavior we introduce is the inverse of what the program used to do.

Do you think we could add a bit more documentation to reflect that fact?

achille-roussel (commented):
It looks like tests that relied on automatic topic creation are failing (which seems expected), for example:
```
--- FAIL: TestWriter (0.00s)
    --- FAIL: TestWriter/writing_a_message_to_a_non-existant_topic_creates_the_topic (0.02s)
        writer_test.go:735: unexpected error [3] Unknown Topic Or Partition: the request is for a topic or partition that does not exist on this broker
        writer_test.go:736: [3] Unknown Topic Or Partition: the request is for a topic or partition that does not exist on this broker
```
Do you think you could update the tests to account for the behavior change?

writer.go
kislerdm: @achille-roussel hey! indeed, you are right wrt docu update. As for failing test, I'd say that the new field shall be inversion of `AllowAutoTopicCreation` for backwards compatibility for default behaviour. It'd also "fix" the test as a side effect :) WDYT?

README.md
```diff
@@ -369,6 +370,55 @@ if err := w.Close(); err != nil {
...
+    },
+}
+
+var err error
+const retries = 3
+for i := 0; i < retries; i++ {
+    ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+    defer cancel()
+    
+    // an attempt to creating topic prior to publishing the message is being made.
```
erikdw, line 405:
```suggestion
    // attempt to create topic prior to publishing the message
```

README.md
```diff
@@ -369,6 +370,55 @@ if err := w.Close(); err != nil {
 }
 ```
 
+### Missing topic creation before publication
+
+```go
+// Make a writer that publishes messages to topic-A.
+// The topic is being created if missing.
```
erikdw, line 377:
```suggestion
// The topic will be created if it is missing.
```

writer_test.go
```diff
@@ -160,6 +160,10 @@ func TestWriter(t *testing.T) {
 			scenario: "writing a message to a non-existant topic creates the topic",
 			function: testWriterAutoCreateTopic,
 		},
+		{
+			scenario: "terminates on an attempts to write a message to a non-existant topic",
```
erikdw, line 164:
```suggestion
			scenario: "terminates on an attempt to write a message to a nonexistent topic",
```

erikdw (changes requested):
Just some minor language nits in comments & strings.

kislerdm:
> Just some minor language nits in comments & strings.

Hey @erikdw, thanks for your feedback. Suggested changes have been applied.

erikdw:
@kislerdm : I'll leave final approval to @achille-roussel since he left some more functional comments.

achille-roussel (approved):
Looking good!

writer.go
```diff
@@ -185,6 +185,9 @@ type Writer struct {
...
+	AllowAutoTopicCreation bool
```
yogeshg, line 189: Any reason to not introduce this in the writer config too?
https://github.com/segmentio/kafka-go/issues/1278

If I am using WriterConfig to create the writer, is it safe to just set this config after calling `NewWriter`?

yogeshg: @erikdw @achille-roussel Here's a PR in case 1278 is a desirable feature request: https://github.com/segmentio/kafka-go/pull/1279 if

Issue #874, opened 2022-03-29
Panic when reading message from start of compacted topic


GFriedrich:
**Describe the bug**
When trying to read from a compacted topic using the `FirstOffset` configuration without a consumer group, the library panics with the following stacktrace.
```
panic: markRead: negative count

goroutine 68 [running]: [github.com/segmentio/kafka-go.(*messageSetReader).markRead(0xc0004587a0)](http://github.com/segmentio/kafka-go.(*messageSetReader).markRead(0xc0004587a0))
    external/com_github_segmentio_kafka_go/message_reader.go:345 +0x11a
[github.com/segmentio/kafka-go.(*messageSetReader).readMessageV2(0xc0004587a0](http://github.com/segmentio/kafka-go.(*messageSetReader).readMessageV2(0xc0004587a0), 0x2c709, 0xc000507ac8, 0xc000507ab8, 0x2c708, 0x2c708, 0xc000059800, 0xc0005078f8, 0x416b3b, 0xc0005078f8, ...)
    external/com_github_segmentio_kafka_go/message_reader.go:329 +0x49d
[github.com/segmentio/kafka-go.(*messageSetReader).readMessage(0xc0004587a0](http://github.com/segmentio/kafka-go.(*messageSetReader).readMessage(0xc0004587a0), 0x2c709, 0xc000507ac8, 0xc000507ab8, 0x2c708, 0xc000507a5c, 0x17fd1b29700, 0x0, 0xc0004922a0, 0x0, ...)
    external/com_github_segmentio_kafka_go/message_reader.go:136 +0xc5
[github.com/segmentio/kafka-go.(*Batch).readMessage(0xc000195880](http://github.com/segmentio/kafka-go.(*Batch).readMessage(0xc000195880), 0xc000507ac8, 0xc000507ab8, 0x0, 0x17fd1b29700, 0x100000001, 0xc0004922a0, 0xc0004922a0, 0xc000507ab8, 0x2)
    external/com_github_segmentio_kafka_go/batch.go:240 +0x79
[github.com/segmentio/kafka-go.(*Batch).ReadMessage(0xc000195880](http://github.com/segmentio/kafka-go.(*Batch).ReadMessage(0xc000195880), 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
    external/com_github_segmentio_kafka_go/batch.go:192 +0x11a
[...]

PR #551, opened 2020-11-02
Ensure headerCount is positive before making headers slice

godwhoa:
Other libraries seems to be doing:
https://github.com/Shopify/sarama/blob/8c3fe6939f5667b9390af8865f834995182b8b58/record.go#L104
https://github.com/apache/kafka/blob/927edfece3db8aab7d01850955f9a65e5c110da5/clients/src/main/java/org/apache/kafka/common/record/DefaultRecord.java#L356
https://github.com/dpkp/kafka-python/blob/a27ab881726ed1a2d952867a1fa266573165d6aa/kafka/record/default_records.py#L241

Might fix #457, not sure if it was due to an actual corrupted record or a bug elsewhere in the decoder.

PR #882, opened 2022-04-11
fix: don't assume the header count will be positive

achille-roussel:
It seems the change in https://github.com/segmentio/kafka-go/pull/551 either did not make it to the main branch, or was reverted at some point.

It was reported in https://github.com/segmentio/kafka-go/issues/874.

I don't know under which conditions Kafka would return a negative header count but apparently kafka does this sometimes 🤷


stevevls (approved):

achille-roussel:
Thanks for the review!

PR #885, opened 2022-04-12
Fix Transport SASL configuration

PzaThief:
Transport struct has "SASL" instead of "SASLMechanism"


achille-roussel (approved):
Thanks for the fix!

PR #886, opened 2022-04-12
Add ProbeCoordinator

jweibel22:
We're experiencing timing issues in our acceptance test setup because it takes a while for the kafka consumer group coordinator to become available and we'd like to be able to probe it instead of adding sleeps to our code.

This PR is just a suggestion, you probably don't want to define this method on the Conn, but it would be really nice if it was somehow possible to wait for the group coordinator to become available.

WDYT?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -313,6 +313,11 @@ func (c *Conn) DeleteTopics(topics ...string) error {
 	return err
 }
 
+func (c *Conn) ProbeCoordinator() error {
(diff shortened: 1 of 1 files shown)


rhansen2:
Hi @jweibel22 

Could you explain a bit more about the new usage pattern this method enables that isn't currently possible? Potentially with a code sample.

Thanks!

rhansen2:
Closing due to inactivity. Feel free to reopen if you'd like to pick this back up.

Issue #599, opened 2021-02-03
Goroutines leak in (*connPool).discover & (*conn).run


Antonboom:
### Describe the bug
**go.uber.org/goleak** founds unexpected goroutines in case of `kafka.(*Client).CreateTopics` error.

</details>

### Kafka Version
Confluent Platform 5.5.1 with **Kafka 2.5.1**.

### To Reproduce
**docker-compose.yaml**
```yaml
version: '3.7'

services:
  zookeeper:
    image: artifactory.sp.corp/docker/confluentinc/cp-zookeeper:5.5.1
    environment:
      ZOOKEEPER_CLIENT_PORT: 2181
    expose:
      - "2181"

  kafka:
    image: artifactory.sp.corp/docker/confluentinc/cp-kafka:5.5.1
    depends_on:
      - zookeeper
    ports:
      - "127.0.0.1:9092:9092"
    environment:
      KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:9092
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
      KAFKA_AUTO_CREATE_TOPICS_ENABLE: "true"
```

**kafka_leak_test.go**
```go
package kafka

import (
	"context"
	"testing"
	"time"

	"github.com/segmentio/kafka-go"
	"go.uber.org/goleak"
)

func TestKafkaLeak(t *testing.T) {
	goleak.VerifyNone(t)
	defer goleak.VerifyNone(t)

	cli := &kafka.Client{
		Addr: kafka.TCP("127.0.0.1:9092"),
	}

	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer func() {
		cancel()
		time.Sleep(time.Second) // Wait for shutdown inside kafka-go.
	}()

	_, err := cli.CreateTopics(ctx, &kafka.CreateTopicsRequest{
		Addr:   cli.Addr,
		Topics: []kafka.TopicConfig{{Topic: "topic"}},
	})
	if err != nil {
		t.Fatalf("unexpected error: %f", err)
	}
}
```

**Test output:**
```
=== RUN   TestKafkaLeak
[...]

PR #887, opened 2022-04-14
Close idle connections

lovromazgon:
I stumbled upon the fact that goroutines are left running after `Writer` is closed. I also found issue #599, specifically this comment https://github.com/segmentio/kafka-go/issues/599#issuecomment-967287388 where it was explained as "expected behavior". I am not convinced that it is actually expected and prepared a PR that "fixes" the behavior, I am curious to hear what you think.

Fixes https://github.com/segmentio/kafka-go/issues/599.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -205,9 +205,6 @@ type Writer struct {
 	// are safe to use concurrently from multiple goroutines, there is no need
 	// for extra synchronization to access this field.
 	roundRobin RoundRobin
-
-	// non-nil when a transport was created by NewWriter, remove in 1.0.
-	transport *Transport
 }
 
 // WriterConfig is a configuration type used to create new instances of Writer.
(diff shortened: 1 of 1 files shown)


rhansen2:
HI @lovromazgon,

Could you elaborate more on why you're not convinced things are worked as expected?

lovromazgon:
I changed my mind and will close the PR. I'll still explain my understanding so others can point out any flaws in my reasoning.

So, because `Transport` is passed to the client, its lifecycle begins outside of the client and is therefore expected to be managed by the caller. If no transport is configured, the client falls back to the default transport and makes sure idle connections are closed after the client itself is closed.

While that reasoning is sound, I would argue that in most cases `Transport` is treated as a configuration object rather than an object with its own lifecycle. For example, if I want to configure TLS I'm forced to create a `Transport` and manage it (i.e. close idle connections), while all I want to essentially achieve is configure the client. Most of the time configuring the client is even done by instantiating and passing a `Transport` directly to the struct / constructor, without ever storing it in a variable. The unexpected thing is then that the client doesn't clean up after itself. Again, I now understand that it can't do that since the transport might be shared, but that doesn't mean it wasn't unexpected when I saw this.

It seems like there's no way around it, creating the transport means you have to deal with it from start to end.

PR #890, opened 2022-04-20
fix: handle error in response

cyx:
While using this in production we ran into a panic where `DescribeConfigs` resulted in a `*net.OpError` (due to the kafka cluster rebalancing).

The panic was because we cast everything to `*Response` even though there are cases where it's an error. Handling this case with an error type seemed more appropriate.

### DIFF
--- a/protocol/describeconfigs/describeconfigs.go
+++ b/protocol/describeconfigs/describeconfigs.go
@@ -1,6 +1,7 @@
 package describeconfigs
 
 import (
+	"fmt"
 	"strconv"
 
 	"github.com/segmentio/kafka-go/protocol"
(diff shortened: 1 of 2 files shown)


protocol/describeconfigs/describeconfigs.go
```diff
@@ -87,11 +88,19 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
...
+			response.Resources = append(
+				response.Resources,
+				v.Resources...,
+			)
+
+		case error:
+			return nil, v
+
+		default:
+			return nil, fmt.Errorf("Unknown result type: %T", result)
```
rhansen2, line 102: Looks like based on https://pkg.go.dev/github.com/segmentio/kafka-go@v0.4.31/protocol#Merger this case should trigger a panic.

rhansen2 (commented):
Hi @cyx,

Thanks for submitting this PR!

protocol/describeconfigs/describeconfigs.go
rhansen2: Hi @cyx, just wanted to bump this thread. Lemmie know if you don't have time to update this PR and I could take it over. Thanks!

cyx: Hi @rhansen2 yes if you have bandwidth would appreciate a fix for this indeed. We seem to be hitting a lot of these in production.

cyx: As for your assertion, I'm not sure about that -- what we're getting is a `*net.OpError` which goes to line 98 error case. If you meant that the default should trigger a panic, sounds good to me.

rhansen2:
Due to some changes in main I recreated this branch and we merged it via #925

PR #892, opened 2022-04-21
Fix Reader handling for OffsetOutOfRange (#891)

### LINKED ISSUE #891: reader doesn't recover (quickly) from OffsetOutOfRange "Offset Out Of Range"
**Describe the bug**

I had a case where the kafka topic was recreated. I am not using a consumer group for this topic, and so it reports, e.g. 

```
error initializing the kafka reader for partition 0 of SOMETOPIC: [1] Offset Out Of Range: the requested offset is outside the range of offsets maintained by the server for the given topic/partition
```

**Kafka Version**

> * What version(s) of Kafka are you testing against?

3.0.0

> * What version of kafka-go are you using?

0.4.30

**To Reproduce**

1. `docker-compose up -d`
2. `go run ./consumer.go`
3. wait for the messages to get processed, leave it running
4. `docker-compose rm -vfs kafka; docker-compose up -d`
5. the topic was recreated, the reader is in a perpetual error loop
6. produce a single message, this message will not be received as its offset will be 0 again.

***docker-compose.yaml***
```yaml
---
version: '3.4'

services:
  zookeeper:
    image: confluentinc/cp-zookeeper:latest
    network_mode: host
    environment:
      ZOOKEEPER_CLIENT_PORT: 2181
      ZOOKEEPER_ADMIN_SERVER_PORT: 10880

  kafka:
    hostname: localhost
    depends_on:
      - zookeeper
    image: confluentinc/cp-kafka:latest
    network_mode: host
    environment:
      KAFKA_BROKER_ID: 1
      KAFKA_ZOOKEEPER_CONNECT: localhost:2181
      KAFKA_ADVERTISED_LISTENERS: PRIMARY://localhost:9092
      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PRIMARY:PLAINTEXT
      KAFKA_INTER_BROKER_LISTENER_NAME: PRIMARY
      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1

```

***consumer.go***
```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"os/signal"
	"syscall"
	"time"

)

const broker = "localhost:9092"
const topic = "my-topic"

func produce(ctx context.Context) error {
	writer := kafka.NewWriter(kafka.WriterConfig{
		Brokers:   []string{broker},
		Topic:     topic,
		BatchSize: 1,
		Logger:    kafka.LoggerFunc(log.New(log.Writer(), "[kafka writer] ", 0).Printf),
	})

	const count = 100
	msgs := make([]kafka.Message, count)
	for i := 0; i < count; i++ {
		msgs[i] = kafka.Message{
			Partition: 0,
			Key:       []byte{byte(i)},
			Value:     []byte(fmt.Sprintf("message %v", i)),
		}
	}
	return writer.WriteMessages(ctx, msgs...)
}

func runner() int {
	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
	defer cancel()

	if err := produce(ctx); err != nil {
		log.Printf("produce: %v", err)
		return 1
	}

		Brokers:   []string{broker},
		Topic:     topic,
		Partition: 0,
		Logger:    kafka.LoggerFunc(log.New(log.Writer(), "[kafka reader] ", 0).Printf),
	})

	defer func() {
		if err := reader.Close(); err != nil {
			panic(err)
		}
	}()
	for {
		select {
		case <-ctx.Done():
			return 0
		default:
		}
		msg, err := reader.ReadMessage(ctx)
		if err != nil {
			log.Printf("ReadMessage: %v", err)
			time.Sleep(1 * time.Second)
			continue
		}
		log.Printf("offset=%v key=%v value=%q", msg.Offset, msg.Key, string(msg.Value))
	}
}

func main() {
	os.Exit(runner())
}

```

**Expected Behavior**

> A clear and concise description of what you expected to happen.

Either the `ReadMessage` call should return with an error, or the reader should gracefully restart at a valid offset.

**Observed Behavior**

> A clear and concise description of the behavior you observed.

Perpetual output of `error initializing the kafka reader for partition 0` message, reader never resumes reading until the offset is once again where it was (this could be a matter of days).

```
[kafka reader] initializing kafka reader for partition 0 of my-topic starting at offset 100
[kafka reader] the kafka reader for partition 0 of my-topic is seeking to offset 100
[kafka reader] error initializing the kafka reader for partition 0 of my-topic: [1] Offset Out Of Range: the requested offset is outside the range of offsets maintained by the server for the given topic/partition
[kafka reader] initializing kafka reader for partition 0 of my-topic starting at offset 100
[kafka reader] the kafka reader for partition 0 of my-topic is seeking to offset 100
[kafka reader] error initializing the kafka reader for partition 0 of my-topic: [1] Offset Out Of Range: the requested offset is outside the range of offsets maintained by the server for the given topic/partition
[kafka reader] initializing kafka reader for partition 0 of my-topic starting at offset 100
[kafka reader] the kafka reader for partition 0 of my-topic is seeking to offset 100
[kafka reader] error initializing the kafka reader for partition 0 of my-topic: [1] Offset Out Of Range: the requested offset is outside the range of offsets maintained by the server for the given topic/partition
[kafka reader] initializing kafka reader for partition 0 of my-topic starting at offset 100
[kafka reader] the kafka reader for partition 0 of my-topic is seeking to offset 100
[kafka reader] error initializing the kafka reader for partition 0 of my-topic: [1] Offset Out Of Range: the requested offset is outside the range of offsets maintained by the server for the given topic/partition
...
```

**Additional Context**
***probelmatic code from reader.go***
```go
func (r *reader) run(ctx context.Context, offset int64) {
...
		conn, start, err := r.initialize(ctx, offset)
		switch err {
		case nil:
		case OffsetOutOfRange:
			// This would happen if the requested offset is passed the last
			// offset on the partition leader. In that case we're just going
			// to retry later hoping that enough data has been produced.
				log.Printf("error initializing the kafka reader for partition %d of %s: %s", r.partition, r.topic, OffsetOutOfRange)
			})
			continue
```
```

wwade:
Provide a new configuration `AutoOffsetReset` that can be used to
reset the non-group-id reader's offset when the topic has been
recreated (or the current offset is no longer valid for any reason).
The `AutoOffsetReset` can have a value of 0 (no action), or
FirstOffset / LastOffset.

Added a new test case in `TestReader` to verify that, on setting the
offset initially past the last offset in the topic, the reader will
restart from the beginning once it detects this OffsetOutOfRange
error, instead of perpetually reporting OffsetOutOfRange and waiting
for the producer to catch up.

Fixes #891.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -495,6 +495,16 @@ type ReaderConfig struct {
 	// Only used when GroupID is set
 	StartOffset int64
 
+	// AutoOffsetReset determines from if, and whence, the reader should restart consuming if
+	// it hits an OffsetOutOfRange error. If non-zero, it must be set to one of FirstOffset or
+	// LastOffset. If AutoOffsetReset is not enabled, the consumer will only be able to resume
+	// reading once the partition's offset has once again reached the current offset value.
+	//
+	// Default: 0 (reset is not enabled)
+	//
+	// Only used when GroupID is *not* set
+	AutoOffsetReset int64
+
 	// BackoffDelayMin optionally sets the smallest amount of time the reader will wait before
 	// polling for new messages
 	//
@@ -1208,6 +1218,7 @@ func (r *Reader) start(offsetsByPartition map[topicPartition]int64) {
 				stats:           r.stats,
 				isolationLevel:  r.config.IsolationLevel,
 				maxAttempts:     r.config.MaxAttempts,
+				autoOffsetReset: r.config.AutoOffsetReset,
 			}).run(ctx, offset)
 		}(ctx, key, offset, &r.join)
 	}
@@ -1233,6 +1244,7 @@ type reader struct {
 	stats           *readerStats
 	isolationLevel  IsolationLevel
 	maxAttempts     int
+	autoOffsetReset int64
 }
 
 type readerMessage struct {
@@ -1268,11 +1280,20 @@ func (r *reader) run(ctx context.Context, offset int64) {
 		case nil:
 		case OffsetOutOfRange:
 			// This would happen if the requested offset is passed the last
-			// offset on the partition leader. In that case we're just going
+			// offset on the partition leader. If autoReset is not enabled, we're just going
 			// to retry later hoping that enough data has been produced.
 			r.withErrorLogger(func(log Logger) {
 				log.Printf("error initializing the kafka reader for partition %d of %s: %s", r.partition, r.topic, OffsetOutOfRange)
 			})
+			switch r.autoOffsetReset {
+			case FirstOffset, LastOffset:
+				offset = r.autoOffsetReset
+			case 0:
+			default:
+				r.withErrorLogger(func(log Logger) {
+					log.Printf("invalid ReaderConfig.AutoOffsetReset value %v for reader of %s", r.autoOffsetReset, r.topic)
+				})
+			}
 			continue
 		default:
 			// Perform a configured number of attempts before
--- a/reader_test.go
+++ b/reader_test.go
@@ -49,6 +49,11 @@ func TestReader(t *testing.T) {
 		},
 
 		{
+			scenario: "setting missing offset for reader",
+			function: testReaderSetMissingOffset,
+		},
+
+		{
 			scenario: "calling Lag returns the lag of the last message read from kafka",
 			function: testReaderLag,
 		},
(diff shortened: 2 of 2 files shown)


wwade:
@dominicbarnes mind having a look at this one?

wwade:
From https://github.com/segmentio/kafka-go/issues/891#issuecomment-1119778790

> What's the use-case for having a topic re-created on the fly? This seems like it might be mostly useful during development, but that is just speculation on my part.
> 
> With that in mind, I'm more inclined to return an error (allowing the program to crash and restart) than to include logic that will automatically recover, as it seems to be an edge-case.
> 
> I'll bring this up on the PR as well, we can have further discussion there.

There's certainly no specific use case, but more that it (could be) an operational reality at some point. I like the idea of returning an error, for sure, that makes a lot of sense to me. That was, in fact, what I was expecting in the first place :smile:

dominicbarnes:
This has been fixed by https://github.com/segmentio/kafka-go/pull/917 which instead opts to return an error explicitly rather than try to automatically reset offsets.

PR #893, opened 2022-04-21
Add "human readable" description for offset in reader logs

wwade:
When the reader logs an offset value, if the value is a special value,
i.e. FirstOffset / LastOffset, it will log something like

    setting the offset ... from -2 to 20

With this change, it'll say "(first offset)" or "(last offset)",
respectively, for example:

    setting the offset ... from (first offset) to 20

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -996,7 +996,7 @@ func (r *Reader) Offset() int64 {
 	offset := r.offset
 	r.mutex.Unlock()
 	r.withLogger(func(log Logger) {
-		log.Printf("looking up offset of kafka reader for partition %d of %s: %d", r.config.Partition, r.config.Topic, offset)
+		log.Printf("looking up offset of kafka reader for partition %d of %s: %s", r.config.Partition, r.config.Topic, toHumanOffset(offset))
 	})
 	return offset
 }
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -1578,3 +1578,14 @@ func extractTopics(members []GroupMember) []string {
...
 }
+
+func describeOffset(offset int64) string {
+	switch offset {
+	case FirstOffset:
+		return "(first offset)"
+	case LastOffset:
+		return "(last offset)"
+	default:
+		return fmt.Sprintf("%d", offset)
```
achille-roussel, line 1589: How about replacing this with `strconv.FormatInt` for stronger type safety?
```suggestion
		return strconv.FormatInt(offset, 10)
```

reader.go
```diff
@@ -1578,3 +1578,14 @@ func extractTopics(members []GroupMember) []string {
 
 	return topics
 }
+
+func describeOffset(offset int64) string {
+	switch offset {
+	case FirstOffset:
+		return "(first offset)"
+	case LastOffset:
+		return "(last offset)"
```
achille-roussel, line 1587: Maybe the resulting log messages would read more naturally if we removed the surrounding parenthesis?
```suggestion
		return "first offset"
	case LastOffset:
		return "last offset"
```

reader.go
```diff
@@ -1578,3 +1578,14 @@ func extractTopics(members []GroupMember) []string {
 
 	return topics
 }
+
+func describeOffset(offset int64) string {
```
achille-roussel, line 1582: An alternative model could be to use a type implementing `fmt.Formatter`?

```go
type humanOffset int64

func (offset humanOffset) Format(w fmt.State, _ rune) {
  ...
}
```
One advantage is it removes the need to allocate an intermediary string to encode the offset into the decimal representation, it can be written directly to the output with `fmt.Fprintf(w, "%d", int64(offset))`.

I don't think it would have a measurable impact tho, up to you to choose with model you think would be the most expressive.

wwade: I like it, thanks!

reader.go
wwade: Done.

reader.go
wwade: Done.

wwade (commented):
Thanks for the review!

achille-roussel (approved):
Thanks for the contribution!

Issue #877, opened 2022-04-01
Add static checker to ensure that we use errors.Is


achille-roussel:
**Describe the solution you'd like**
Following up from https://github.com/segmentio/kafka-go/pull/846, we should have automation to ensure that we don't introduce regressions on error checking.

PR #914, opened 2022-05-16
enable errorlint linter and fix issues

dominicbarnes:
This PR enables the [errorlint](https://github.com/polyfloyd/go-errorlint) linter in golangci-lint. This specifically ensures that errors are properly checked, avoiding things like direct equality (`==`) checks and type coercion/switch, to ensure as much compatibility with improvements to errors in go 1.13+.

This is a continuation of #877, but will need to be followed up by one more PR to enable another linter to ensure we wrap errors coming from external packages.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,7 @@
 linters:
   enable:
     - bodyclose
+    - errorlint
     - goconst
     - godot
     - gofmt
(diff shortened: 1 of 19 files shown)


rhansen2 (approved):
LGTM

PR #917, opened 2022-05-20
Reader: allow config to return OffsetOutOfRange errors

### LINKED ISSUE #891: reader doesn't recover (quickly) from OffsetOutOfRange "Offset Out Of Range"

dominicbarnes:
This PR aims to change the handling of `OffsetOutOfRange` errors in the `Reader`, but includes some quirks to retain backwards-compatibility. This will fix #891 and makes #892 unnecessary.

In short, when a topic is re-created (ie: deleted and re-created with same name), any readers that are already running will begin encountering `OffsetOutOfRange` errors, but will retry them indefinitely, blocking the program from making progress. This error can also occur when a particular broker advances past the leader, so retrying made sense as this was usually temporary.

The original issue and PR related to this opted to automatically "reset" the offset when this error was encountered, but this felt a little too magical and I've opted instead to return an error to allow the client to handle it in whatever way makes sense for the use-case. Since topic re-creation feels like an edge-case (such as only happening during active development), it feels best to be more explicit rather than adopting this logic into the reader itself.

A configuration flag is being added, largely to allow existing clients to remain unaffected by these changes, with the expectation that this behavior should change in a future release of kafka-go.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,11 +3,13 @@ module github.com/segmentio/kafka-go
 go 1.15
 
 require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/klauspost/compress v1.14.2
 	github.com/pierrec/lz4/v4 v4.1.14
-	github.com/stretchr/testify v1.6.1
+	github.com/stretchr/testify v1.7.1
 	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c
 	github.com/xdg/stringprep v1.0.0 // indirect
 	golang.org/x/crypto v0.0.0-20190506204251-e1dfcc566284 // indirect
 	golang.org/x/net v0.0.0-20190404232315-eb5bcb51f2a3
+	gopkg.in/yaml.v3 v3.0.0-20220512140231-539c8e751b99 // indirect
 )
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,6 @@
-github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
+github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/klauspost/compress v1.14.2 h1:S0OHlFk/Gbon/yauFJ4FfJJF5V0fc5HbBTJazi28pRw=
 github.com/klauspost/compress v1.14.2/go.mod h1:/3/Vjq9QcHkK5uEr5lBEmyoZ1iFhe47etQ6QUkpK6sk=
 github.com/pierrec/lz4/v4 v4.1.14 h1:+fL8AQEZtz/ijeNnpduH0bROTu0O3NZAlPjQxGn8LwE=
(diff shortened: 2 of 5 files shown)


achille-roussel (approved):
This is looking great 🙌  thanks for the contribution!

wwade:
@dominicbarnes thanks for taking this on, much appreciated. I rolled it in to my project and ran my test again and it worked nicely.

PR #918, opened 2022-05-23
Fix auto topic creation test

caioaao:
Test was deferring the topic deletion so it wasn't actually testing the auto topic creation


caioaao:
Left the tests running to show the failure - it seems auto topic creation is broken and the test wasn't able to catch it

caioaao:
For context, it seems it was an issue with kafka's docs: https://www.mail-archive.com/users@kafka.apache.org/msg09182.html
Confluent's go client also had the same issue: https://github.com/confluentinc/confluent-kafka-go/issues/615

PR #925, opened 2022-06-03
fix: handle error in response

rhansen2:
While using this in production we ran into a panic where DescribeConfigs resulted in a *net.OpError (due to the kafka cluster rebalancing).

The panic was because we cast everything to *Response even though there are cases where it's an error. Handling this case with an error type seemed more appropriate.

This is a recreation of https://github.com/segmentio/kafka-go/pull/890 which maintains the original authors commits.

### DIFF
--- a/protocol/describeconfigs/describeconfigs.go
+++ b/protocol/describeconfigs/describeconfigs.go
@@ -1,6 +1,7 @@
 package describeconfigs
 
 import (
+	"fmt"
 	"strconv"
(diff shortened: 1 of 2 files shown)


achille-roussel (approved):
🙌 

Thanks for the fix!

dmarkhas:
Guys the same issue happens in all other request... (we are seeing it in ListGroups).

PR #931, opened 2022-06-14
fix(protocol): fix type assertion error

kscooo:
In the event of a sudden network outage (such as a sudden wifi disconnection or vpn request to reconnect), the `ListGroups` method panics because the result is incorrectly asserted as Response, but is actually an `error` type (e.g. `context deadline exceeded` or `*net.OpError` etc)

### DIFF
--- a/protocol/listgroups/listgroups.go
+++ b/protocol/listgroups/listgroups.go
@@ -65,6 +65,7 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 		}
 		brokerResp := m.(*Response)
 		respGroups := []ResponseGroup{}
+
 		for _, brokerResp := range brokerResp.Groups {
 			respGroups = append(
 				respGroups,
--- a/protocol/listgroups/listgroups.go
+++ b/protocol/listgroups/listgroups.go
@@ -63,7 +63,7 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 		if err != nil {
 			return nil, err
 		}
-		brokerResp := result.(*Response)
+		brokerResp := m.(*Response)
 
 		respGroups := []ResponseGroup{}
 		for _, brokerResp := range brokerResp.Groups {
--- a/protocol/listgroups/listgroups.go
+++ b/protocol/listgroups/listgroups.go
@@ -1,6 +1,8 @@
 package listgroups
 
 import (
+	"fmt"
+
 	"github.com/segmentio/kafka-go/protocol"
 )
 
@@ -57,8 +59,15 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 	response := &Response{}
 
 	for r, result := range results {
-		brokerResp := result.(*Response)
-		respGroups := []ResponseGroup{}
+		brokerResp, ok := result.(*Response)
+		if !ok {
+			if err, ok := result.(error); ok {
+				return nil, fmt.Errorf("listgroups.(*Response).Merge: %w", err)
+			}
+			return nil, fmt.Errorf("listgroups.(*Response).Merge expect got %T", result)
+		}
+
+		respGroups := make([]ResponseGroup, 0, len(brokerResp.Groups))
 
 		for _, brokerResp := range brokerResp.Groups {
 			respGroups = append(
--- a/protocol/listgroups/listgroups.go
+++ b/protocol/listgroups/listgroups.go
@@ -65,7 +65,6 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 		}
 		brokerResp := m.(*Response)
 		respGroups := []ResponseGroup{}
-		
 		for _, brokerResp := range brokerResp.Groups {
 			respGroups = append(
 				respGroups,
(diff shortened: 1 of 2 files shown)


protocol/listgroups/listgroups.go
```diff
@@ -57,8 +59,15 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 	response := &Response{}
 
 	for r, result := range results {
-		brokerResp := result.(*Response)
-		respGroups := []ResponseGroup{}
+		brokerResp, ok := result.(*Response)
```
nlsun, line 62: @kscooo good catch! What do you think about modeling your bug fix after this one? https://github.com/segmentio/kafka-go/blob/da91759e946436612f6e31af58971162bf1d058d/protocol/listoffsets/listoffsets.go#L151

kscooo: forgive me for not knowing the rest of the source code for this project, it looks like this fix is much better

nlsun: No problem! The `protocol` package (`"github.com/segmentio/kafka-go/protocol"`) already has an implementation of the `Result` function though, do you think you could use that instead of redefining it in this file?

kscooo: Removed, sorry for not noticing that the `*Response` implements `Message` interface

protocol/listgroups/listgroups.go
```diff
@@ -57,10 +59,18 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
...
-		respGroups := []ResponseGroup{}
+		brokerResp, err := protocol.Result(result)
+		if err != nil {
+			return nil, err
+		}
+
+		resp, ok := brokerResp.(*Response)
+		if !ok {
+			return nil, fmt.Errorf("BUG: result must be a response but not %T", brokerResp)
+		}
```
nlsun, line 70:
```suggestion
		m, err := protocol.Result(result)
		if err != nil {
			return nil, err
		}
		brokerResp := result.(*Response)
```

protocol/listgroups/listgroups.go
```diff
@@ -57,10 +59,18 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
...
 
-		for _, brokerResp := range brokerResp.Groups {
+		respGroups := make([]ResponseGroup, 0, len(resp.Groups))
+		for _, brokerResp := range resp.Groups {
```
nlsun, line 73:
```suggestion
		respGroups := []ResponseGroup{}
		for _, brokerResp := range brokerResp.Groups {
```

nlsun (commented):
some tweaks

protocol/listgroups/listgroups.go
```diff
@@ -57,9 +59,13 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 	response := &Response{}
 
 	for r, result := range results {
+		m, err := protocol.Result(result)
+		if err != nil {
+			return nil, err
+		}
 		brokerResp := result.(*Response)
```
nlsun, line 66:
```suggestion
		brokerResp := m.(*Response)
```

protocol/listgroups/listgroups.go
```diff
@@ -57,9 +59,13 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 	response := &Response{}
 
 	for r, result := range results {
-		brokerResp := result.(*Response)
-		respGroups := []ResponseGroup{}
+		m, err := protocol.Result(result)
+		if err != nil {
+			return nil, err
+		}
+		brokerResp := m.(*Response)
```
nlsun, line 67:
```suggestion
```

protocol/listgroups/listgroups.go
```diff
@@ -57,9 +59,12 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
 	response := &Response{}
 
 	for r, result := range results {
-		brokerResp := result.(*Response)
+		m, err := protocol.Result(result)
+		if err != nil {
+			return nil, err
+		}
+		brokerResp := m.(*Response)
 		respGroups := []ResponseGroup{}
```
nlsun, line 67:
```suggestion
		respGroups := []ResponseGroup{}
		
```

protocol/listgroups/listgroups.go
```diff
@@ -57,9 +59,13 @@ func (r *Response) Merge(requests []protocol.Message, results []interface{}) (
...
-
+
```
nlsun, line 68:
```suggestion

```

protocol/listgroups/listgroups.go
nlsun, line 67:
```suggestion
		respGroups := []ResponseGroup{}

```

protocol/listgroups/listgroups.go
```diff
@@ -1,6 +1,8 @@
 package listgroups
 
 import (
+	"fmt"
+
```
nlsun, line 5:
```suggestion
```

nlsun:
Thanks for the fix!

PR #936, opened 2022-06-24
refactor: enable wrapcheck linter and address all issues

dominicbarnes:
This PR aims to improve our ability to troubleshoot by wrapping more errors with context, relying on the `wrapcheck` linter to detect all places where errors likely should be wrapped.

Opening this as a draft to begin with while we review as a team and see if there are any caveats to worry about when wrapping errors.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,8 +7,34 @@ linters:
     - gofmt
     - goimports
     - prealloc
+    - wrapcheck
 
   disable:
     # Temporarily disabling so it can be addressed in a dedicated PR.
     - errcheck
     - goerr113
+
+linters-settings:
+  govet:
+    disable:
+      - deepequalerrors
+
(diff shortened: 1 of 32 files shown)


dominicbarnes:
Turning on this linter is fraught with problems, especially since we implement so many low-level interfaces like `net.Conn` and `io.Reader`, which can sometimes be very strict with returned errors as they don't leverage `errors.Is` very much (if at all).

I could add a bunch more exclusions to the `wrapcheck` linter config, but that seems to be a game of whack-a-mole. I think the deeper an error exists, the more likely it is to be in the chain of one of these interface methods, so we may need to stick to the higher-level interfaces, but we may have already lost a lot of context by that point.

In any case, this PR seems way too risky to merge as-is, so I'm going to close it and see if I can revisit this another time.

PR #941, opened 2022-07-01
feat(reader): handle io.ErrNoProgress more gracefully

### LINKED ISSUE #932: the kafka reader got an unknown error reading partition X of Y at offset Z: multiple Read calls return no data or error
**Describe the bug**
We have 2 different topics on our kafka cluster.  Both have multiple partitions, and both see millions of messages per day, with one of them almost hitting a half-billion and the other in the 1-10 million range.
I have two separate golang processes running, one reading from one topic, and the other reading from the other topic.
Every day I see hundreds of messages in my log that look like:
`the kafka reader got an unknown error reading partition 46 of mytopicname at offset 929168893: multiple Read calls return no data or error`

It has been this way since I started using this library, maybe 1 year ago, and I have tried to keep the library up to date.
I can confirm there is data at that partition and offset:
```
$ ./bin/kafka-console-consumer.sh --bootstrap-server kafka2-00.xxx.net:9092 --topic mytopicname --partition 46 --offset 929168893 --max-messages 1
{"route".......
```

**Kafka Version**
Kafka version 2.4.0
kafka-go version 0.4.32

**To Reproduce**
```go
kafkaReader := kafka.NewReader(kafka.ReaderConfig{
		Brokers: strings.Split(env.KafkaBrokers, ","),
		GroupID: env.KafkaGroup,
		Topic:   env.KafkaTopic,
		ErrorLogger: kafka.LoggerFunc(func(s string, i ...interface{}) {
			log.WithField("error", fmt.Sprintf(s, i...)).Info("Kafka library debug")
		}),
	})

msg, err := consumer.kafkaReader.FetchMessage(ctx)
if err != nil {
	panic(err)
}
myCh <- msg
```

**Expected Behavior**
Not to receive this error if this partition + offset has data, which it does.
Instead, receive that data/mesage.

**Observed Behavior**
`the kafka reader got an unknown error reading partition 46 of mytopicname at offset 929168893: multiple Read calls return no data or error`

dominicbarnes:
This change adjusts the `Reader` to intentionally close the connection on `io.ErrNoProgress`, with the expectation that a retry would likely work. Since we're explicitly handling this error/failure, we're suppressing the logs/metrics that were brought up in #932

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1321,6 +1321,14 @@ func (r *reader) run(ctx context.Context, offset int64) {
 				errcount = 0
 				continue
 
+			case errors.Is(err, io.ErrNoProgress):
+				// This error is returned by the Conn when it believes the connection
+				// has been corrupted, so we need to explicitly close it. Since we are
+				// explicitly handling it and a retry will pick up, we can suppress the
+				// error metrics and logs for this case.
+				conn.Close()
+				break readLoop
+
 			case errors.Is(err, UnknownTopicOrPartition):
 				r.withErrorLogger(func(log Logger) {
 					log.Printf("failed to read from current broker for partition %d of %s at offset %d, topic or parition not found on this broker, %v", r.partition, r.topic, toHumanOffset(offset), r.brokers)


dominicbarnes:
Before merging this, I'll update the code to close the connection and break the read loop. This would retain the current functionality, and just bypass the error log + metrics. In addition, I'm still looking for a reproducible test case, if that is something we can pull together.

rhansen2 (approved):

achille-roussel (approved):

PR #942, opened 2022-07-01
protocol: Use standardized result type checking

nlsun:
Fixing the rest of the issues pointed out by https://github.com/segmentio/kafka-go/pull/931


achille-roussel (approved):
Thanks for fixing!

PR #943, opened 2022-07-02
add consumer group apis to client

rhansen2:
JoinGroup,SyncGroup,LeaveGroup

### DIFF
--- a/leavegroup.go
+++ b/leavegroup.go
@@ -1,6 +1,115 @@
 package kafka
 
-import "bufio"
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/segmentio/kafka-go/protocol/leavegroup"
+)
+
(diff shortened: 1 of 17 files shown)


leavegroup.go
```diff
@@ -1,6 +1,115 @@
...
+		// If we're using a version of the api without the
+		// members array in the response, just add a member
+		// so the api is consistent across versions.
+		res.Members = []LeaveGroupResponseMember{
+			{
+				ID:              req.Members[0].ID,
+				GroupInstanceID: req.Members[0].GroupInstanceID,
+			},
+		}
+		return res, nil
```
achille-roussel, line 99: One way to make this a bit less of a special case would be to assign `r.Members` instead of `res.Members`, and let the code carry on with the regular path. That way we always take the code path below even when dealing with older API versions.

Just a suggestion, feel free to leave it as-is if you think it's more readable.

achille-roussel (approved):
This is looking great 🙌 Thanks for adding support for these APIs!

PR #947, opened 2022-07-08
Refactor Consumer Group to use Client

rhansen2:


achille-roussel (commented):
We discussed the PR during our team meeting today and agreed on making the following changes:

- base the change on the 0.4 branch since it contains minor modifications to the exported APIs (removed an unused field + change fields from `int32` to `int` types)
- explore passing a parent context to the methods of the `coordinator` interface
- add the `AllowAutoTopicCreation` field to the top level `MetadataRequest` type so we don't need to bypass the abstraction layers

reader.go
```diff
@@ -708,12 +710,56 @@ func NewReader(config ReaderConfig) *Reader {
...
+			SASL:     kafkaDialer.SASLMechanism,
+			TLS:      kafkaDialer.TLS,
+			ClientID: kafkaDialer.ClientID,
+		}
+
+		if transport.ClientID == "" {
+			transport.ClientID = DefaultClientID
+		}
+
+		r.transport = transport
```
achille-roussel, line 756: We do something similar in `kafka.Writer` I believe, would it be worth abstracting this logic internally?

consumergroup.go
```diff
@@ -73,15 +70,23 @@ type ConsumerGroupConfig struct {
...
 	// Topics is the list of topics that will be consumed by this group.  It
 	// will usually have a single value, but it is permitted to have multiple
 	// for more complex use cases.
 	Topics []string
 
+	// A transport used to send messages to kafka clusters.
+	//
+	// If nil, a Transport will be created used the provided ClientID or the DefaultClientID
+	// if not provided.
+	Transport RoundTripper
```
achille-roussel, line 82: Would it be possible to use `kafka.DefaultTransport` instead of creating a new transport when the program doesn't provide one?

rhansen2: Need to set the client id for the consumer group; Let's see what happens if multiple consumers use the same ID. 
Otherwise document users are expected to create a unique transport for the Reader.

moolitayer:
Hi @rhansen2 @achille-roussel,

Thank you for working on this awsome library!

Recently we've seen an issue on upgrading from v0.4.35 to v0.4.36. The issue went away after downgrad:
```
Failed to join X: [79] Member ID Required: the group member needs to have a valid member id before actually entering a consumer group
Failed to join group X: kafka.(*Client).JoinGroup: unexpected EOF
Failed to sync group X: kafka.(*Client).SyncGroup: unexpected EOF
Failed to sync group X: [27] Rebalance In Progress: the coordinator has begun rebalancing the group, the client should rejoin the group
kafka.(*Client).SyncGroup: unexpected EOF
```

Do you think it might be ralted to this change? Should I create an issue for it with our env details + code?

rhansen2:
Could be related. Please open an issue and we'll take a look, thanks!

jhartzell:
> Hi @rhansen2 @achille-roussel,
> 
> Thank you for working on this awsome library!
> 
> Recently we've seen an issue on upgrading from v0.4.35 to v0.4.36. The issue went away after downgrad:
> 
> ```
> Failed to join X: [79] Member ID Required: the group member needs to have a valid member id before actually entering a consumer group
> Failed to join group X: kafka.(*Client).JoinGroup: unexpected EOF
> Failed to sync group X: kafka.(*Client).SyncGroup: unexpected EOF
> Failed to sync group X: [27] Rebalance In Progress: the coordinator has begun rebalancing the group, the client should rejoin the group
> kafka.(*Client).SyncGroup: unexpected EOF
> ```
> 
> Do you think it might be ralted to this change? Should I create an issue for it with our env details + code?

Also seeing the same issue under a Strimzi cluster

rhansen2:
@moolitayer or @jhartzell Are either of you able to create an issue with a reproduction of the issues you're seeing?

PR #948, opened 2022-07-13
Update CI to test kafka-go against kafka 2.8.1

code-haven:


code-haven:
Hmm, I see that `go test -race` hangs when I run against 2.8.1 but works just fine with 2.7.1.  Will try and figure out what's happening here.

Issue #927, opened 2022-06-05
p.Writer.WriteMessages: context deadline exceeded on MSK IAM


ayoul3:
**Describe the bug**
I activated IAM authentication on a Kafka cluster 2.7.0 (AWS MSK). Reading topics worked like a charm.  
However when writing to a partition I suffered timeout errors at the discovery query when listing all brokers and partitions. It turns out that brokers on port 9098 does not like this [empty](https://github.com/segmentio/kafka-go/blob/main/transport.go#L599) Request metadata v8. Specifically the empty Topics field translated to 0xffffffff at the network level. 

I reached out to AWS MSK to look into it,  but maybe there is something to be done at the client to avoid running into this issue.
The regular Kafka Java client does not have this issue, so there is a chance they'll just snooze off my bug report. We'll see.

I did not find an open issue on this, so putting this one here for visibility. Curious to know if someone else could reproduce.

**Kafka Version**
2.7.0
> * What version(s) of Kafka are you testing against?
2.7.0
> * What version of kafka-go are you using?
v0.4.28, v.0.4.29, v0.4.31, v0.4.32

**To Reproduce**
* Initialise an MSK cluster version 2.7.0 and activate IAM authentication.
* Create a topic and define variables KAFKA_SERVERS, AWS_REGION, KAFKA_TOPIC
* Run the following script
```go
package main

import (
	"context"
	"crypto/tls"
	"os"
	"strings"
	"time"

	"github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/session"
	sigv4 "github.com/aws/aws-sdk-go/aws/signer/v4"
	"github.com/rs/zerolog/log"
	"github.com/segmentio/kafka-go"
	"github.com/segmentio/kafka-go/sasl/aws_msk_iam"
)

func main() {
	awsRegion := aws.String(os.Getenv("AWS_REGION"))
	sess, err := session.NewSession(&aws.Config{
		Region: awsRegion,
	})
	if err != nil {
		log.Logger.Fatal().Err(err)
	}

	kafkaDialer := &kafka.Dialer{
		Timeout:   10 * time.Second,
		DualStack: true,
		SASLMechanism: &aws_msk_iam.Mechanism{
			Signer: sigv4.NewSigner(sess.Config.Credentials),
			Region: *awsRegion,
		},
		TLS: &tls.Config{},
	}

	brokers := strings.Split(os.Getenv("KAFKA_SERVERS"), ",")
[...]

PR #950, opened 2022-07-14
remove nullable from metadata request topics

rhansen2:
Kafka protocol doesn't define nullable arrays.

https://kafka.apache.org/protocol.html#protocol_types

Potential fix for #927


rhansen2:
Not it.

Issue #957, opened 2022-07-27
WriteErrors don't return the actual error


mhmtszr:
**Describe the bug**

> We can't see the actual errors of our batch write request. It creates new error and this error does not contain any information about the error `return fmt.Sprintf("kafka write errors (%d/%d)", err.Count(), len(err))`. We need to see the detail of the error.

**Kafka Version**

> * github.com/segmentio/kafka-go v0.4.29

**To Reproduce**

> Get any error while processing the batch write request. 

**Expected Behavior**

> We need to see the detail of the error.

**Observed Behavior**

> We can't see the actual error we only see the count of the errors.

rhansen2:
Hi @mhmtszr,

See https://pkg.go.dev/github.com/segmentio/kafka-go?utm_source=godoc#WriteErrors for an example of inspecting individual errors within a WriteErrors.

Hopefully that helps!

Thanks!

mhmtszr:
@rhansen2 hi,

Thanks this solved our problem but I think there should be a cleaner way, we shouldn't have to do this conversion what do you think?

rhansen2:
What did you have in mind?

mhmtszr:
Instead of this
```
func (err WriteErrors) Error() string {
	return fmt.Sprintf("kafka write errors (%d/%d)", err.Count(), len(err))
}
```

We may error message like this
```
func (err WriteErrors) Error() string {
	var sb strings.Builder
	for i, writeError := range err {
		sb.WriteString(fmt.Sprintf("%d: %s ", i, writeError.Error()))
	}
	return fmt.Sprintf("kafka write errors (%d/%d), errors: ", err.Count(), len(err), sb.String())
}
```

In this way we can log the producer errors without doing any convertion.

rhansen2:
I think that sounds good @mhmtszr, would you be able to submit a pull request?

mhmtszr:
@rhansen2 https://github.com/segmentio/kafka-go/pull/964 I created a pull request and will wait for the reviews thanks.

PR #964, opened 2022-08-03
Update WriteErrors error message

mhmtszr:
We can't see the actual errors of our batch write request. It creates new error and this error does not contain any information about the error return fmt.Sprintf("kafka write errors (%d/%d)", err.Count(), len(err)). We need to see the detail of the error.

Issue thread: https://github.com/segmentio/kafka-go/issues/957

### DIFF
--- a/error.go
+++ b/error.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"syscall"
 )
 
@@ -701,5 +702,9 @@ func (err WriteErrors) Count() int {
 }
 
 func (err WriteErrors) Error() string {
-	return fmt.Sprintf("kafka write errors (%d/%d)", err.Count(), len(err))
+	var sb strings.Builder
+	for i, writeError := range err {
+		sb.WriteString(fmt.Sprintf("Message %d: %s \n", i, writeError.Error()))
+	}
+	return fmt.Sprintf("kafka write errors (%d/%d), errors: %s", err.Count(), len(err), sb.String())
 }
--- a/error.go
+++ b/error.go
@@ -685,7 +685,6 @@ func makeError(code int16, message string) error {
 //		// handle other errors
 //		...
 //	}
-//
 type WriteErrors []error
 
 // Count counts the number of non-nil errors in err.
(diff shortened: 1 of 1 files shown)


error.go
```diff
@@ -701,5 +702,9 @@ func (err WriteErrors) Count() int {
 }
 
 func (err WriteErrors) Error() string {
-	return fmt.Sprintf("kafka write errors (%d/%d)", err.Count(), len(err))
+	var sb strings.Builder
+	for i, writeError := range err {
+		sb.WriteString(fmt.Sprintf("Message %d: %s \n", i, writeError.Error()))
```
achille-roussel, line 707:
```suggestion
		fmt.Fprintf(&sb, "Message %d: %s \n", i, writeError.Error())
```

achille-roussel (approved):
Thanks for the contribution!

mhmtszr:
Hi @achille-roussel, are we going to merge this pr?

error.go
```diff
@@ -701,5 +702,9 @@ func (err WriteErrors) Count() int {
 }
 
 func (err WriteErrors) Error() string {
-	return fmt.Sprintf("kafka write errors (%d/%d)", err.Count(), len(err))
+	var sb strings.Builder
+	for i, writeError := range err {
```
rhansen2, line 706: Can write error ever be nil here?

mhmtszr: i don't think so, if err is empty then it will never execute it what do you think? @rhansen2

rhansen2: It is not possible for one element of the array to be nil?

rhansen2: Based on what's documented [here](https://pkg.go.dev/github.com/segmentio/kafka-go#WriteErrors) I believe we have to handle the case when an item in the array is nil.

mhmtszr: @rhansen2 thanks you are right, I added validation of that.

error.go
```diff
@@ -701,5 +701,12 @@ func (err WriteErrors) Count() int {
...
+		if writeError == nil {
+			continue
+		}
+		fmt.Fprintf(&sb, "Message %d: %s \n", i, writeError.Error())
```
rhansen2, line 709: Sorry, one list nit, it looks like the individual errors are at batch level not individual message level, `Message` here might be a bit misleading.

mhmtszr: @rhansen2 I removed the string builder part, instead of that I used a slice to be cleaner what do you think about it?

rhansen2: Thanks @mhmtszr This all looks good to me!

achille-roussel (approved):

PR #981, opened 2022-09-06
add connection counter

techmanishk:


dominicbarnes:
Thanks for your contribution @techmanishk! Would you mind adding a description to this PR so it's more clear what functionality you are proposing we add? There also seems to be a minor lint issue that you should be able to correct as well.

dominicbarnes:
It's been a while since we've heard anything about this PR, so I'm going to close it as it is not clear what is being added/proposed, and there seem to be a number of lint/test issues that need to be fixed in addition to needing a rebase. Feel free to re-open with additional context after it has been updated/rebased.

Issue #852, opened 2022-02-17
Create a sane Writer + Batch System


arianitu:
**Describe the solution you'd like**

The Writer abstraction is a bad design when it comes to writing messages and batching.

*Current System*

Users call WriteMessages which takes N messages. The user must collect N messages themselves and then call WriteMessages which is basically creating a batching system on top of a batching system. What's worse is if a user calls WriteMessages in a for loop (the most common way of writing messages from streams, files, etc.) it throttles you to about 1 / second by default.

There's lots of issues where users switch BatchTimeout to 10 * time.Millisecond which is just a hack to get the desired / intuitive behaviour.  

*New System*

Users called WriteMessage which batches internally and sends them at either BatchSize or BatchTimeout. This API is intuitive and anyone can use it. Users who send messages in for loop (most common way of writing messages from streams, files, etc.) will get the expected behaviour.

The user no longer has to create a batching system on top of the existing batching system and just has to call WriteMessage which all their messages and it will handle the rest.

nlsun:
Hi @arianitu , have you had a chance to try the Writer in "asynchronous mode"? https://github.com/segmentio/kafka-go/blob/b952e635d4f1b688854c1097cd7ce6eeaf50b273/writer.go#L54-L74

hadrienk:
I think the `WriteMessages()` implementation is problematic in several ways: 

It can either be asynchronous or not, depending on the config. I think we should have dedicated async methods. Something like:
```go 
func (w *BufferedWriter) WriteMessagesAsync(ctx context.Context, messages ...kafka.Message) <-chan error {
  // ...
}
```
The current WriteMessages method can buffer, timeout or be async. This is too complicated. In some cases we want to write a messages immediately. We should have a method for that: 
```
// WriteMessages write the messages. This method blocks until all the messages are sent.
func (w *BufferedWriter) WriteMessages(ctx context.Context, messages ...kafka.Message)  error {
  // ...
}
```

In cases where performances matter we could add a buffered API: 

```go
// WriteBuffered saves the messages in a buffer, flushing every BufferSize.
// If there's room in the buffer, no message are written. The returned error is the first
// error returned by the Flush method.
func (w *BufferedWriter) WriteBuffered(ctx context.Context, messages ...kafka.Message) error {
	// [...]
}

// Flush sends and then empties the content of the buffer. 
func (w *BufferedWriter) Flush(ctx context.Context) error {
  // [...]
}
```

[...]

Issue #707, opened 2021-07-15
Why is sarama so fast compared to segmentio's kafka-go? 


alexec:
Original issue: https://github.com/Shopify/sarama/issues/1647 and #326

As with everything, I want it to be fast and reliable, but reliably trumps speed every time.

Do you have a feature comparison table?

achille-roussel:
Hello @alexec, thanks for opening this issue.

"fast" is a qualitative metric, could you describe the performance requirements of your application and how you have determined that kafka-go was the bottleneck?

If this is of any use, we employ kafka-go in production and frequently achieve throughput in the 10K msg/s consumed or produced on a single process, sometimes a lot more depending on how much work the program has to do. This is surely "fast enough" for our use cases, but I understand that different systems may need to achieve much higher throughput.

alexec:
Good question. I was getting about 400 TPS with Sarama. With kafka-go (and 200 less LOC) I'm getting 400 TPS reading using kafka-go too (and dumping to a log), but I want to write messages back to another topic. When I add `WriteMessages` into the loop, then TPS drops to 150 TPS. My assuming is that Sarama is committing in batches. I'm going to try different batch sizes and report back.

alexec:
This is not a bug, just need to set configuration correctly.

3AceShowHand:
> This is not a bug, just need to set configuration correctly.

Hi, can you describe the configuration more in detail, which one is more critical to solve the issue ?

Issue #326, opened 2019-07-25
Why write message so slow!


shukyoo:
**Describe the bug**
I compare sarama and kafka-go to write msg.
kafka-go wait about 2 seconds to finish
but sarama finish immediately!
why this so slow!

**this is stat:**
```
	Dials: 1 Writes: 1 Messages: 1 Bytes: 17 Rebalances: 1 Errors: 0 DialTime: {
		Avg: 7.452734 ms Min: 7.452734 ms Max: 7.452734 ms
	}
	WriteTime: {
		Avg: 1.017793708 s Min: 1.017793708 s Max: 1.017793708 s
	}
	WaitTime: {
		Avg: 5.342901 ms Min: 5.342901 ms Max: 5.342901 ms
	}
```

**this is code:**
```go
	w := kafka.NewWriter(kafka.WriterConfig{
		Brokers: []string{"127.0.0.1:1024"},
		Topic:   "test1",
		Balancer: &kafka.LeastBytes{},
	})

	w.WriteMessages(context.Background(),
		kafka.Message{
			Key:   []byte("Key-A"),
			Value: []byte("Hello World!"),
		},
	)

	fmt.Printf("%+v\n", w.Stats())

	w.Close()
```

achille-roussel:
I believe there may be multiple factors that come into play in the example you submitted:

- The default settings are optimized for production workloads where multiple messages are batched together because this is how kafka is commonly used to achieve high throughput and efficient resource usage.

- On the first write, the writer will also have to open connections to the kafka cluster, which will then remain open and be reused for future writes. The connection pooling logic in kafka-go may be using a different strategy which front-loads more work instead of doing things lazily.

My guess is whether you publish 1 or 1000 messages in a single batch, the publish time will be roughly the same.

Would you be able to give more details about the kafka cluster you're using (number of brokers, number of partitions, etc...), as well as explaining your use case a bit more?

If you're looking at reducing the batch timeout in the [WriterConfig](https://godoc.org/github.com/segmentio/kafka-go#WriterConfig), this will reduce the amount time that the writer waits for more data before publishing.

If you wish to never block on writes, you can also configure the writer to work in asynchronous mode.

Let me know if this is helpful!

shukyoo:
I made a batch write test, write msg 100 times
this is code
```go
	w := kafka.NewWriter(kafka.WriterConfig{
		Brokers: []string{"127.0.0.1:1024"},
		Topic:   "test1",
[...]

PR #986, opened 2022-09-12
Write messages nobatch

yusufozturk:
This PR does not change anything for the existing functions but simply appending a new function to cover requirements like:

#852 
#326 
#707 

New fuction is called "WriteMessageGroup" because user needs to group messages. This function skips the batching part, and sends messages to producer directly. So it should be faster than current function for the pre-batched messages.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -27,29 +27,29 @@ import (
 // by the function and test if it an instance of kafka.WriteErrors in order to
 // identify which messages have succeeded or failed, for example:
 //
-//	// Construct a synchronous writer (the default mode).
-//	w := &kafka.Writer{
-//		Addr:         Addr: kafka.TCP("localhost:9092", "localhost:9093", "localhost:9094"),
-//		Topic:        "topic-A",
-//		RequiredAcks: kafka.RequireAll,
-//	}
-//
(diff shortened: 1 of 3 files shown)


yusufozturk:
May I kindly ask for your help and review on this PR? @achille-roussel / @nlsun

We want to start using the new function via the original package. Thanks for this great library!

nlsun:
Thanks @yusufozturk for the PR, created a new issue to further discuss this topic https://github.com/segmentio/kafka-go/issues/994

writer.go
```diff
@@ -680,6 +680,57 @@ func (w *Writer) WriteMessages(ctx context.Context, msgs ...Message) error {
...
+		Partition:    partition,
+		Topic:        w.Topic,
+		RequiredAcks: w.RequiredAcks,
+		Compression:  w.Compression,
+		Records: &writerRecords{
+			msgs: msgs,
+		},
+	})
+
+	if pr.Error != nil {
```
ybrustin, line 727: "pr" can be nil here (either because of RequiredAcks=RequireNone or err != nil).
Also, I would check "err" before accessing "pr"

PR #988, opened 2022-09-14
Write the base offset of the record set

ehwoahthere:
The base offset was always 0 before but probably should be the offset of the first record.


achille-roussel:
Hello @ehwoahthere, thanks for the contribution.

Would you be able to add a test that demonstrates the need for this change?

ehwoahthere:
@achille-roussel updated with a test case!

Issue #726, opened 2021-08-25
the kafka reader got an unknown error reading partition 


nsomarouthu:
### we're often getting a reader error ~500k/day. 
```
the kafka reader got an unknown error reading partition 9 of SOME_TOPIC at offset 3: read tcp IP_ADDRESS:46406->IP_ADDRESS:9093: i/o timeout
```
https://github.com/segmentio/kafka-go/blob/a4890bd956b5658ca57e964dd28381c0ee4fd617/reader.go#L1366-L1375

### Second issue

The errors happened when kafka reader is committing the message after it has been processed successfully, the message was re-consumed by another replica 

"msg": "debezium.Consumer: failed to commit message: write tcp IP_ADDRESS:49610->IP_ADDRESS:9093: use of closed network connection"
Receiving Successfully handled message and a while afterward getting failed to commit message

achille-roussel:
Hello @nsomarouthu, thanks for reporting the issue.

Do you know whether these errors are causing adverse effect on your application? Or are they simply transient errors that are handled and recovered by kafka-go automatically?

I/O timeouts seem like frequent problems in network systems, so I would not be too concerned about it, but let me know if this is causing any other issues!

nsomarouthu:
Hello  @achille-roussel,  If there is any setting in kafka.ReaderConfig to reduce the number of errors that we are experiencing currently with our configuration. Reader exit the loops when reporting this issue and reconnect. This doesn't impact the reading functionality of the message, but this seems expensive.

chihweichang:
@achille-roussel I am in the same team/company as @nsomarouthu, working together on the 2 issues.
For the 1st issue,  we have many `the kafka reader got an unknown error reading partition` as well as numerous `no messages received from kafka within the allocated time for partition` logged.  I believe this is because our applications set `ReaderConfig.MaxWait` https://github.com/segmentio/kafka-go/blob/master/reader.go#L395 to 100ms. There are some messages that we want the application to fetch and handle as soon as possible.  Do you see any concern with 

For the 2nd issue, we have a few `stopped commit for group xxx` messages about 5 seconds after finished handling and trying to commit the Kafka message. I believe this is related to `ConsumerGroupConfig.Timeout` (default to 5s) https://github.com/segmentio/kafka-go/blob/master/consumergroup.go#L161
Is there a way to set a different (longer) value when creating a `NewReader` with a `ReaderConfig`
https://github.com/segmentio/kafka-go/blob/master/reader.go#L704

[...]

PR #989, opened 2022-09-14
#726 make kafka reader message ReadBatchTimeout configurable 

moogacs:
PR makes message `safetyTimeout` as `ReadBatchTimeout `  configurable because `SetReadDeadline` for msg could differentiate based on the network and  make it configurable instead of opinionated configuration

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1525,14 +1525,8 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
 	var size int64
 	var bytes int64
 
-	deadline := time.Now().Add(r.maxSafetyTimeout)
-	conn.SetReadDeadline(deadline)
-
 	for {
-		if now := time.Now(); deadline.Sub(now) < (r.maxSafetyTimeout / 2) {
-			deadline = now.Add(r.maxSafetyTimeout)
-			conn.SetReadDeadline(deadline)
-		}
+		conn.SetReadDeadline(time.Now().Add(r.maxSafetyTimeout))
 
 		if msg, err = batch.ReadMessage(); err != nil {
 			batch.Close()
--- a/reader.go
+++ b/reader.go
@@ -400,8 +400,7 @@ type ReaderConfig struct {
 	// Default: 10s
 	MaxWait time.Duration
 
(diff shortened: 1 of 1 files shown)


reader.go
```diff
@@ -400,6 +400,11 @@ type ReaderConfig struct {
 	// Default: 10s
 	MaxWait time.Duration
 
+	// MaxSafetyTimeout amount of time to wait to fetch message from kafka messages batch.
+	//
+	// Default: 10s
+	MaxSafetyTimeout time.Duration
```
achille-roussel, line 406: What do you think about renaming this to `ReadBatchTimeout` since it's controlling the time we allocate to read batches off of the network connection?

reader.go
```diff
@@ -1514,15 +1525,8 @@ func (r *reader) read(ctx context.Context, offset int64, conn *Conn) (int64, err
...
-	const safetyTimeout = 10 * time.Second
-	deadline := time.Now().Add(safetyTimeout)
-	conn.SetReadDeadline(deadline)
-
 	for {
-		if now := time.Now(); deadline.Sub(now) < (safetyTimeout / 2) {
-			deadline = now.Add(safetyTimeout)
-			conn.SetReadDeadline(deadline)
-		}
+		conn.SetReadDeadline(time.Now().Add(r.maxSafetyTimeout))
```
achille-roussel, line 1529: Maybe we could keep this simpler by moving this call before the loop? That way we only pay the cost of setting the deadline once per batch instead of once per message?

The code we had before was a bit too complex but attempted to reduce the cost of these `SetReadDeadline` calls.

reader.go
moogacs: I like it it's more descriptive, happy to change

reader.go
moogacs: I agree, what I was thinking keeping it in every message, so that if it was small though, the batch still can continue. the problem I am afraid is that one don't know how big is the batch and it would be harder to give it the `SetReadDeadline ` a correct value? what do you think? correct me if i miss something

achille-roussel (approved):
Thanks for your contribution!

PR #991, opened 2022-09-19
Allows users to configure the GroupInstanceID

huszkacs:
GroupInstanceID can be configured by setting ReaderConfig.GroupInstanceID which will be passed to the joinGroup and heartbeat requests. If the GroupInstanceID has a value the kafka server will set it as a static member.

### DIFF
--- a/consumergroup.go
+++ b/consumergroup.go
@@ -73,6 +73,9 @@ type ConsumerGroupConfig struct {
 	// for more complex use cases.
 	Topics []string
 
+	// The unique identifier of the consumer instance provided by end user.
+	GroupInstanceID string
+
 	// A transport used to send messages to kafka clusters.
 	//
 	// If nil, DefaultTransport will be used.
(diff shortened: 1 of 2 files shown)


rhansen2:
Hi @huszkacs,

Thanks for submitting this PR!. Just wanted to clarify if this PR is meant to fully support the behavior described in https://cwiki.apache.org/confluence/display/KAFKA/KIP-345:+Introduce+static+membership+protocol+to+reduce+consumer+rebalances#KIP345:Introducestaticmembershipprotocoltoreduceconsumerrebalances-ClientSideChanges or if you're just looking to support passing the GroupInstanceID?

huszkacs:
@rhansen2 it is meant to fully support the behaviour. Thanks to the new structure, there is no need for any other change. Previously I had more complicated PR.

(sorry for the late answer)

rhansen2:
It looks like https://cwiki.apache.org/confluence/display/KAFKA/KIP-345:+Introduce+static+membership+protocol+to+reduce+consumer+rebalances describes some changes to the client behavior, specifically `For the effectiveness of the KIP, consumer with `group.instance.id` set will not send leave group request when they go offline` and some changes to the range balancer. Are we already compliant with those behaviors?

huszkacs:
that is a good question, I take a look

rhansen2:
Closed due to inactivity as well as some of the changes this depended on being reverted.

PR #1001, opened 2022-09-30
Refine err msg in SetOffsetAt()

haorenfsa:


reader.go
```diff
@@ -1090,7 +1093,7 @@ func (r *Reader) SetOffsetAt(ctx context.Context, t time.Time) error {
 
 		return r.SetOffset(offset)
 	}
-	return fmt.Errorf("error setting offset for timestamp %+v", t)
+	return fmt.Errorf("error dailing all brokers, one of the error: %w", err)
```
achille-roussel, line 1096:
```suggestion
	return fmt.Errorf("error dialing all brokers, one of the errors: %w", err)
```

reader.go
```diff
@@ -1073,13 +1073,16 @@ func (r *Reader) SetOffsetAt(ctx context.Context, t time.Time) error {
 		return io.ErrClosedPipe
 	}
 	r.mutex.Unlock()
-
+	if len(r.config.Brokers) < 1 {
+		return errors.New("no broker in config")
```
rhansen2, line 1077:
```suggestion
		return errors.New("no brokers in config")
```

haorenfsa:
There seems to be some problem with the CI, it may fail on master

rhansen2 (approved):
Thanks for the contribution!

PR #1003, opened 2022-10-06
Change return type of reader.Close() to void

OscBacon:


achille-roussel:
Hello @OscBacon, thanks for your contribution!

The reason the method signature is `Close() error` is because it follows the signature of the `io.Closer` interface.

The change you suggest might cause compilation to break in programs that either check the error value or convert `*kafka.Reader` to an interface like `io.Closer`, so we need a very strong reason to justify the engineering time that will be spent updating applications.

Would you be able to provide an explanation of what motivated this change?

OscBacon:
Hi @achille-roussel, my pleasure!
I was refactoring my code, and was trying to figure out if an error returned by reader.Close() should be fatal for my program. After going through the function definition, I realized that no error can be thrown.
With this in mind, I think it would be clearer for users that there is no need for error checking!
Perhaps the doc for reader.Close could mention that no error is thrown?

achille-roussel:
Actually, I don't think we should document that there will be no errors either, otherwise we may have to break this promise in the future if we ever need to return an error.

The intent here is to leave it up to the application to decide what to do: and error when closing might be considered a fatal condition which requires aborting the program, or it might be better to record the error and continue the execution... this is context-dependent and I don't believe kafka-go should dictate what the right decision would be.

Let me know if this answer is useful!

PR #1008, opened 2022-10-14
change writer retries stat from summary to counter

### LINKED ISSUE #1005: retries stat should be a counter
**Describe the solution you would like**

I have been going through the stats that you are exposing for the `reader` and `writer`. The `retries` stat on the `writer` doesn't make sense to me as it's a summary stat but it acts as a counter:
```
type writerStats struct {
        ...
	retries        summary
}
...
if attempt != 0 {
	stats.retries.observe(1)
        ...
}      
```
This means `min`,`max` and `avg` will always equal 1, which basically makes the stat useless. This should be changed to a `counter` instead or is there some other way of getting the actual number of retries out?

achille-roussel:
Fixes #1005 

Technically a breaking change but it seems simple enough for applications to address, any use of the `retries` metric prior to this change would have likely been erroneous anyways.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -356,7 +356,7 @@ type WriterStats struct {
 	BatchTime  DurationStats `metric:"kafka.writer.batch.seconds"`
 	WriteTime  DurationStats `metric:"kafka.writer.write.seconds"`
 	WaitTime   DurationStats `metric:"kafka.writer.wait.seconds"`
-	Retries    int64         `metric:"kafka.writer.retries.count"`
+	Retries    int64         `metric:"kafka.writer.retries.count" type:"counter"`
 	BatchSize  SummaryStats  `metric:"kafka.writer.batch.size"`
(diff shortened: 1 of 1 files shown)


writer.go
```diff
@@ -356,7 +356,7 @@ type WriterStats struct {
 	BatchTime  DurationStats `metric:"kafka.writer.batch.seconds"`
 	WriteTime  DurationStats `metric:"kafka.writer.write.seconds"`
 	WaitTime   DurationStats `metric:"kafka.writer.wait.seconds"`
-	Retries    SummaryStats  `metric:"kafka.writer.retries.count"`
+	Retries    int64         `metric:"kafka.writer.retries.count"`
```
nlsun, line 359: the other lines have `type:"counter"`, not sure what it does though

nlsun (approved):
1 comment otherwise lgtm

PR #1014, opened 2022-10-22
compress/snappy: configurable compression level

chriso:
There are three snappy compression routines provided by https://github.com/klauspost/compress. At the moment kafka-go is hard-coded to use `snappy.Encode` aka. `s2.EncodeSnappyBetter`.

This PR makes the compression level configurable, allowing you to choose either a faster compression routine (`s2.EncodeSnappy`) or a slower one that compresses better (`s2.EncodeSnappyBest`).

There appears to be a global set of codecs shared by all producers. To set the level for all producers configured to compress with snappy, you'd need to do something like:

```go
compress.SnappyCodec.Compression = snappy.FasterCompression
```

This is the same approach taken by https://github.com/segmentio/kafka-go/blob/main/compress/gzip/gzip.go. Ideally we'd be able to set the compression level at the producer level, but it looks like a much larger refactor would be required.

### DIFF
--- a/compress/snappy/snappy.go
+++ b/compress/snappy/snappy.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"sync"
 
+	"github.com/klauspost/compress/s2"
 	"github.com/klauspost/compress/snappy"
 )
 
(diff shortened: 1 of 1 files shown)


achille-roussel (approved):
Nice addition, thanks for the contribution!

PR #1017, opened 2022-10-27
use useSyncCommits

wsy6543:
Unified use of useSyncCommits to judge


achille-roussel (approved):
Thanks for the contribution!

Issue #1021, opened 2022-11-02
Multiple consumer group offset problems with 0.4.36


dselans:
**Describe the bug**

Several consumer group issues with `ReadMessage` and `FetchMessage` when using lib version `0.4.36`:

1. Offsets are not committed at all when `CommitMessages` is called manually after a successful `FetchMessage`
2. Offsets are partially committed when using ReadInterval + CommitInterval set to 0
3. Downgrading to `0.4.35` fixes both of the issues

**Kafka Version**

2.4.1.1

**To Reproduce**

```
groupID := "foo"

readerConfig := kafka.ReaderConfig{
	Brokers: k.Options.Brokers,
	GroupID: groupID,
	Topic:   topic,
	MaxWait: k.Options.ReaderMaxWait,
	Dialer:  k.Dialer,
	CommitInterval: 0,
}

reader := kafka.NewReader(readerConfig) 

// Try to read some messages
m, err := reader.ReadMessage(context.Background())
if err != nil {
    // err
}

// Close reader, create new one w/ same GroupID
m, err := reader.ReadMessage(context.Background())
if err != nil {
    // err
}

// A dupe read will occur

// If you view kafdrop or use another utility to view lag - the consumer group will likely have lag as well.

// When using FetchMessage and calling CommitMessages - offsets do not appear to be committed at all.
```

Thank you!

moogacs:
I am suspecting this PR https://github.com/segmentio/kafka-go/pull/947

rhansen2:
Hi @dselans,

Thanks for reporting these issues. Based on your Kafka version it looks like you're using MSK is that correct? If so are you able to reproduce your issues against non-MSK clusters?

I'm struggling to reproduce this behavior, are you able to provide logs from your readers? They may be helpful in determining what's going on.

https://pkg.go.dev/github.com/segmentio/kafka-go#readme-logging

Thanks!

RO-29:
I am facing partially the same issue and it caused a huge problem on our side due to duplicates it generated

offsets are committed partially 


If it helps I started receiving multiple warnings also after this upgrade

>I am suspecting this PR https://github.com/segmentio/kafka-go/pull/947
I agree on this aspect too as I see memberID being introduced in this PR but not entirely sure of the change

```
 Failed to join group identifier-ingester: [79] Member ID Required: the group member needs to have a valid member id before actually entering a consumer group
```

dselans:
[...]

PR #1022, opened 2022-11-03
Fixes for consumer group

rhansen2:
Moving the consumer group to the Client has introduced a number of bugs. This fixes a couple of them.


rhansen2:
Potentially helps with some of https://github.com/segmentio/kafka-go/issues/1021

achille-roussel (approved):
👏

Issue #1061, opened 2023-01-12
Memory leak in reflect package


mhmtszr:
I realized memory increase in our application and I dumped the heap usage and saw that kafka-go library allocates a lot of heap memory.
![image](https://user-images.githubusercontent.com/35079195/212180996-45fcf7ff-a568-4efe-9019-cf6ef102700b.png)

In reflect.go file:
```
func makeArray(t reflect.Type, n int) array {
	return array{val: reflect.MakeSlice(reflect.SliceOf(t), n, n)}
}

func bytesToString(b []byte) string { return string(b) }
```

- Every makeArray function call allocates a new slice, we may use sync.pool here to decrease the allocation.
- bytesToString makes an unnecessary string allocation, we may use unsafe pointer to avoid allocation.

shivansh-bb:
I have observed similar issue and this memory is not released even after closing the writer and the transport.

mhmtszr:
@shivansh-bb you can look at https://github.com/search?q=repo%3Asegmentio%2Fkafka-go+unsafe&type=code building application with unsafe tag, reflect_unsafe.go may reduce your memory usage.

shivansh-bb:
Its using unsafe_NewArray yet allocating huge memory to objects that is not getting cleared on closing the writer.
PFA Attached memory profile

rhansen2:
Hi,

@mhmtszr Did building with unsafe solve your issue?

@shivansh-bb Is your memory profile of inuse objects or total allocated objects?

mhmtszr:
Hi @rhansen2 ,

I did build with unsafe but still there are huge amount of allocations.

rhansen2:
@mhmtszr, Given that the library allocates based on the amount of data being produced or consumed it's hard to say how much it should be allocating. How much memory are you seeing allocated relative to the amount of data being consumed or produced?

shivansh-bb:
> @shivansh-bb Is your memory profile of inuse objects or total allocated objects?
Hi @rhansen2 
profile was for the allocated inuse_space.

mhmtszr:
@rhansen2 not much, we are not consuming a lot of messages. Memory usage is being increased constantly.

rhansen2:
@mhmtszr @shivansh-bb Thanks for the additional context!

[...]

PR #1062, opened 2023-01-12
Avoid unnecessary string allocation while decoding

mhmtszr:
https://github.com/segmentio/kafka-go/issues/1061

I realized memory increase in our application and I dumped the heap usage and saw that kafka-go library allocates a lot of heap memory. 

I found a solution for bytesToString memory leak, by using the unsafe package we can convert byte to string or string to byte without extra allocation.

### DIFF
--- a/protocol/reflect.go
+++ b/protocol/reflect.go
@@ -5,6 +5,7 @@ package protocol
 
 import (
 	"reflect"
+	"unsafe"
 )
 
 type index []int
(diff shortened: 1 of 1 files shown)


achille-roussel:
Hello @mhmtszr thanks for opening the PR!

The feature you are looking for is actually already available by adding the `unsafe` build tag (e.g. with `go build -tags unsafe ...`).

We make it an opt-in feature to prioritize safety and correctness over performance, there is always more risks of introducing defects when unsafe is involved.

Maybe we could add a section to the README to document this feature?

mhmtszr:
I will try it, thanks @achille-roussel. It would be great if this would be in README, seems great feature.

PR #1063, opened 2023-01-15
Use buffer pool for decompressed buffer

ashishkf:
Directly using batch.Read or through the Reader interface requires creating new batch using `conn.ReadBatchWith`. This causes the decompressed buffer to be allocated fresh and requires repeated calls to Grow and then it is discarded on batch.Close.

To avoid this, this patch acquires new buffer when creating newMessageSetReader and releases it in batch.Close.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -81,6 +81,7 @@ func (batch *Batch) close() (err error) {
 	batch.lock = nil
 	if batch.msgs != nil {
 		batch.msgs.discard()
+		releaseBuffer(batch.msgs.decompressed)
 	}
 
 	if err = batch.err; errors.Is(batch.err, io.EOF) {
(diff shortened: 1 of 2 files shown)


batch.go
```diff
@@ -81,6 +81,7 @@ func (batch *Batch) close() (err error) {
 	batch.lock = nil
 	if batch.msgs != nil {
 		batch.msgs.discard()
+		releaseBuffer(batch.msgs.decompressed)
```
achille-roussel, line 84: Could we change this to set the decompressed field to `nil` after releasing it? Otherwise there's a risk to releasing the buffer again if Close is called multiple times.

```go
if batch.msgs.decompressed != nil {
   releaseBuffer(batch.msgs.decompressed)
  batch.msgs.decompressed = nil
}
```

ashishkf: Good catch. Fixed now.

ashishkf:
friendly reminder for review.

achille-roussel (approved):
Thanks for the improvement! 🚀

PR #1070, opened 2023-02-01
Update dialer.go

dxasu:
"errch <- UnknownTopicOrPartition" unable reachable, It will ReadPartitions loop until context timeout

### DIFF
--- a/dialer.go
+++ b/dialer.go
@@ -193,6 +193,7 @@ func (d *Dialer) LookupPartition(ctx context.Context, network string, address st
 					return
 				}
 			}
+			break
 		}
 
 		errch <- UnknownTopicOrPartition


dxasu:
"errch <- UnknownTopicOrPartition" unable reachable, It will ReadPartitions loop until context timeout

dialer.go
```diff
@@ -193,6 +193,7 @@ func (d *Dialer) LookupPartition(ctx context.Context, network string, address st
 					return
 				}
 			}
+			break
 		}
 
 		errch <- UnknownTopicOrPartition
```
achille-roussel, line 199: Could we remove this line instead of adding a `break` ? That way we don't create new conditions in the code that have never been exercised?

PR #1077, opened 2023-02-14
Change comment on MaxAttempts

C-Pro:
Comment on `ReaderConfig.MaxAttempts` is confusing: word "deliver" in context of a kafka driver can be understood like "deliver over network", and it did not specify what exactly will be attempted.


achille-roussel (approved):
Thanks for the update! 👏

PR #1086, opened 2023-02-26
Remove runtime.Gosched

sejin-P:
After [go 1.14](https://go.dev/doc/go1.14), go scheduler can interrupt without function call and can ensure that goroutines get fair access to the processor.  Moreover, it could decrease performance because of intentional goroutine switching.


nlsun (approved):
Thanks!

PR #1095, opened 2023-03-13
Add DeleteGroups function to Client

bgranvea:
Implement DeleteGroups protocol and add it to Client.

### DIFF
--- a/deletegroups.go
+++ b/deletegroups.go
@@ -53,11 +53,7 @@ func (c *Client) DeleteGroups(
 	}
 
 	for _, t := range r.Responses {
-		if t.ErrorCode == 0 {
-			ret.Errors[t.GroupID] = nil
-		} else {
-			ret.Errors[t.GroupID] = Error(t.ErrorCode)
-		}
+		ret.Errors[t.GroupID] = makeError(t.ErrorCode, "")
 	}
 
 	return ret, nil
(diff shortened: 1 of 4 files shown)


protocol/deletegroups/deletegroups_test.go
```diff
@@ -0,0 +1,33 @@
...
+	for _, version := range []int16{0, 1, 2} {
+		prototest.TestResponse(t, version, &deletegroups.Response{
+			Responses: []deletegroups.ResponseGroup{
+				{
+					GroupID:   "group1",
+					ErrorCode: 1,
+				},
+				{
+					GroupID:   "group2",
+					ErrorCode: 1,
```
nlsun, line 28: ErrorCode 1 is a failure, could you write a test case that succeeds as well

deletegroups.go
```diff
@@ -0,0 +1,64 @@
...
+		Throttle: makeDuration(r.ThrottleTimeMs),
+		Errors:   make(map[string]error, len(r.Responses)),
+	}
+
+	for _, t := range r.Responses {
+		if t.ErrorCode == 0 {
+			ret.Errors[t.GroupID] = nil
+		} else {
+			ret.Errors[t.GroupID] = Error(t.ErrorCode)
+		}
```
nlsun, line 60: use the `makeError` function instead of `Error` (see other examples for usage)

also don't need the if/else statement as `makeError` takes care of that too

bgranvea: thanks for the review, I've made the changes.

PR #1096, opened 2023-03-14
Not omit error in logger

itechdima:
Add explicit errors as arguments so anyone can handle them in the logger implementation.

We stumbled across issue that during Kafka rebalancing we have plenty of different error logs (EOF, Unexpected EOF, not a leader, etc.). The only possible way to handle that noise is to add some middleware logic into our implementation of logger. So, to have better backward compatibility it's better to not check against string values, but against errors explicitly.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -238,7 +238,7 @@ func (r *Reader) commitLoopInterval(ctx context.Context, gen *Generation) {
 
 	commit := func() {
 		if err := r.commitOffsetsWithRetry(gen, offsets, defaultCommitRetries); err != nil {
-			r.withErrorLogger(func(l Logger) { l.Printf(err.Error()) })
+			r.withErrorLogger(func(l Logger) { l.Printf("%v", err) })
 		} else {
 			offsets.reset()
 		}
@@ -311,7 +311,7 @@ func (r *Reader) run(cg *ConsumerGroup) {
 			}
 			r.stats.errors.observe(1)
 			r.withErrorLogger(func(l Logger) {
-				l.Printf(err.Error())
+				l.Printf("%v", err)
 			})
 			// Continue with next attempt...
 		}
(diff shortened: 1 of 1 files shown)


velanse (approved):

marthjod:
Our goal here is to reduce alert noise for our alerts which refer to logged errors. A lot of errors we see are transient and can be treated with lower criticality in our case. As of now, we have no way of differentiating between errors we care about and those transient ones. To get a better handle on the different kinds of errors without hacky string matching, we would like to be able to check and filter via the actual error types in our middleware. This way, we can more cleanly filter and decide what to bubble up and what to treat as intermittent failure.

itechdima:
@nlsun @achille-roussel 
could you please take a look?

itechdima:
@rhansen2 
Maybe you can facilitate here? :)

reader.go
```diff
@@ -1346,7 +1346,7 @@ func (r *reader) run(ctx context.Context, offset int64) {
 
 			case errors.Is(err, UnknownTopicOrPartition):
 				r.withErrorLogger(func(log Logger) {
-					log.Printf("failed to read from current broker for partition %d of %s at offset %d, topic or parition not found on this broker, %v", r.partition, r.topic, toHumanOffset(offset), r.brokers)
+					log.Printf("failed to read from current broker for partition %d of %s at offset %d, topic or parition not found on this broker, %v: %v", r.partition, r.topic, toHumanOffset(offset), r.brokers, err)
```
rhansen2, line 1349: It looks like with these new formats we'll have 
`failed to read from current broker for partition %d of %s at offset %d, topic or parition not found on this broker, %v: Unknown Topic Or Partition`

Could we possible reformat these a bit so we don't log redundant messages?

Potentially like
`
failed to read from current broker for partition %d of %s at offset %d, Unknown Topic Or Partition, %v`

rhansen2 (approved):
Thanks for the contribution!

PR #1108, opened 2023-04-01
Update SCRAM dependency

SoMuchForSubtlety:
github.com/xdg/scram was moved to github.com/xdg-go/scram, the original no longer gets updated. 

See https://github.com/xdg/scram#readme


rhansen2:
Thanks for the contribution. CircleCI is having trouble with the build. I'm going to try closing and reopening to trigger a build.

PR #1110, opened 2023-04-05
Don't invoke r.log() if not r.debug

colega:
Profiling shows that up to 40% of our allocations happen in r.markRead(): just to allocate a slice to pass the "remain" arg, which will be discarded because we don't run with r.debug=true.

```
  Total:    64259028   64259028 (flat, cum) 41.57%
    350            .          .           	if r.count == 0 { 
    351            .          .           		panic("markRead: negative count") 
    352            .          .           	} 
    353            .          .           	r.count-- 
    354            .          .           	r.unwindStack() 
    355     64259028   64259028           	r.log("Mark read remain=%d", r.remain) 
    356            .          .           } 
    357            .          .            
    358            .          .           func (r *messageSetReader) unwindStack() { 
    359            .          .           	for r.count == 0 { 
    360            .          .           		if r.remain == 0 { 
```

This makes the code slightly verbose, having to check the debug mode before invoking r.log(), but it's a huge difference in high-throughput applications.

### DIFF
--- a/message_reader.go
+++ b/message_reader.go
@@ -159,7 +159,9 @@ func (r *messageSetReader) readMessageV1(min int64, key readBytesFunc, val readB
 		if codec, err = r.header.compression(); err != nil {
 			return
 		}
-		r.log("Reading with codec=%T", codec)
+		if r.debug {
+			r.log("Reading with codec=%T", codec)
+		}
 		if codec != nil {
 			// discard next four bytes...will be -1 to indicate null key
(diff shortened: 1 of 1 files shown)


rhansen2 (approved):

colega:
:wave: Hey, we're running with this change, and this function doesn't appear in the profiles anymore. It has zero impact disabled. I guess it's not harmful to leave the debug logs in case someone needs to see them.

PR #1125, opened 2023-05-07
Fetch multiple messages at once

sergiusd:
I need to fetch messages immediately and process them with parallel worker processes. After processing, I call CommitMessages and move the offset.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -814,14 +814,7 @@ func (r *Reader) FetchMessage(ctx context.Context) (Message, error) {
 	r.activateReadLag()
 
 	for {
-		r.mutex.Lock()
-
-		if !r.closed && r.version == 0 {
-			r.start(r.getTopicPartitionOffset())
-		}
-
-		version := r.version
-		r.mutex.Unlock()
+		version := r.fetchMessageStartAndGetVersion()
 
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


sergiusd:
Hi @rhansen2
What do you think of this feature? Does this make sense or should I use something else?
And let me know please, did I add the test in the wrong place or is my code not finished yet?

rhansen2:
I'm not sure I understand the need for this to be withing kafka-go, it looks like this is equivalent to just calling FetchMessage in a loop. Please let me know if I'm not understanding.

sergiusd:
This may be needed in a scenario where you need to read a batch of messages, process and commit an offsett after processing.
An example of such scenarios:
1. To utilize 10 processors, we read 10+ messages and process them in parallel.
2. To speed up the processing of messages, during which requests are made to an external resource, for example, lasting 1 second - we read the batch, send requests in parallel, which reduces the response time.
What do you think about it?

rhansen2:
We often use that same pattern but batch the messages ourselves. 

Unless I'm misunderstanding this PR seems similar to just doing something like

```go
r := kafka.NewReader(...)
const n = 10
var msgs []kafka.Message
waitTimeout := time.Second
for {
 ctx, cancel := context.WithTimeout(context.Background(), waitTimeout)
 msg, err := r.FetchMessage(ctx)
 cancel()
 if errors.Is(err, context.DeadlineExceeded) {
     // break the loop
 }
 // handle other errors
 msgs = append(msgs, msg)
 if len(msgs) == n {
   // break the loop 
 }

}
```

sergiusd:
Yes, you are right, it can be implemented on the application side, so then I will do it.

zkqiang:
@rhansen2  Why not add a feature that fetch multiple messages in network at once?

crossRT:
Hi @rhansen2 @sergiusd 
I have a scenario that I need to read a batch of 2000 messages and submit them to OpenSearch via a single Bulk API call.
How would you recommend to do that?

My understanding is that I need to:
1. for loop `r.FetchMessage()` with 2000 times.
2. append the messages into an array.
3. submit the 2000 messages to OpenSearch via single Bulk API.
4. commit 2000 messages with `r. CommitMessages()`
5. clean up the messages array.
6. continue the next for loop

Is this correct?
Could you share some example code to do this?

PR #1129, opened 2023-05-18
Fix doc example for break from retryer on success

lanzay:


rhansen2 (approved):
Thanks for the contribution!

Issue #963, opened 2022-08-02
writer.WriteMessages(...) method generates "Denied Operation = Describe" events on non-related topics


chinniehendrix:
**Describe the bug**

> When producing keyed messages to a specific topic using the writer.WriteMessages(...) method with mutual TLS and ACLs enabled, the client tries to access other topics (for which it does not have permissions) generating lots of "Denied Operation = Describe" events. This behavior, when shared simultaneously by hundreds of clients, can have a significant impact on the performance of the Kafka cluster.

**Kafka Version**

> * What version(s) of Kafka are you testing against?
v2.4.1

> * What version of kafka-go are you using?
v0.4.33

**To Reproduce**

> All resources can be found in this public repository: https://github.com/chinniehendrix/local-kafka-poc/tree/feat/demo-kafka-go-describe-issue/strimzi/tls-clients

```go
package main

import (
	"context"
	"crypto/tls"
	"crypto/x509"
	"fmt"
	"io/ioutil"
	"time"

	log "github.com/sirupsen/logrus"

	"github.com/segmentio/kafka-go"
)

func tlsConfig() *tls.Config {
	userCert, errCert := ioutil.ReadFile("/client-certs/user.crt")
	if errCert != nil {
		log.Fatal(errCert)
	}

	userKey, errKey := ioutil.ReadFile("/client-certs/user.key")
	if errKey != nil {
		log.Fatal(errKey)
	}

	caPEM, errCA := ioutil.ReadFile("/ca-certs/ca.crt")
	if errCA != nil {
		log.Fatal(errCA)
	}

	cert, err := tls.X509KeyPair(userCert, userKey)
	if err != nil {
		log.Fatal(err.Error())
	}

	caCertPool := x509.NewCertPool()
	caCertPool.AppendCertsFromPEM([]byte(caPEM))

	config := &tls.Config{
		MinVersion:   tls.VersionTLS12,
		Certificates: []tls.Certificate{cert},
		RootCAs:      caCertPool,
	}

	fmt.Println("Returning TLS config")
	return config
}

func main() {
	now := time.Now()
	fmt.Println("Go client started")
	fmt.Println(now)
	logger := log.New()

	w := &kafka.Writer{
		Addr:        kafka.TCP("my-cluster-kafka-brokers:9093"),
		Topic:       "my-topic",
		Compression: kafka.Lz4,
		Transport: &kafka.Transport{
			TLS: tlsConfig(),
		},
[...]

PR #1131, opened 2023-05-18
Fix unnecessary topic metadata request

mhmtszr:
**Problem**
As mentioned [here](https://github.com/segmentio/kafka-go/issues/963#issuecomment-1279253873), metadata information was coming for all the topics in the cluster, in large clusters this number is quite high. We needed a config where we could give the topic names in order to get the metadata information only for the topics we were interested in.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -81,6 +81,10 @@ type Transport struct {
 	// Default to 6s.
 	MetadataTTL time.Duration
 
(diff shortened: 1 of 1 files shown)


mhmtszr:
@achille-roussel @rhansen2 could you please review it? This affects our resource consumption.

rhansen2 (approved):

rhansen2:
Thanks for the contribution!

PR #1132, opened 2023-05-19
Writer: ensure batching logic factors in message headers

### LINKED ISSUE #1050: Message accounting does not include size of headers.
**Describe the bug**

Message accounting does not include size of headers. See:

https://github.com/segmentio/kafka-go/blob/v0.4.38/message.go#L43-L45

When messages get close to the bytes limit, they may end up going over due to headers. This results in a tcp rst from Azure Event Hub.

**Kafka Version**

Azure Event Hubs

**To Reproduce**

Send a batch messages with about 1MB of _payload_. Add headers. Boom.

**Expected Behavior**

Success.

**Observed Behavior**

**Additional Context**

dominicbarnes:
Currently, the decision to include a message in a batch does not consider headers, so you can get payloads that are larger than intended if you are close to your upper limit and then include headers. This adds a new test for batching with headers that failed and then a fix was added to correct the failing test. This also updates another test that began failing once we included the header length in this calculation.

This closes #1050

### DIFF
--- a/message.go
+++ b/message.go
@@ -49,6 +49,17 @@ func (msg *Message) size() int32 {
 	return 4 + 1 + 1 + sizeofBytes(msg.Key) + sizeofBytes(msg.Value) + timestampSize
 }
 
+func (msg *Message) headerSize() int {
+	return varArrayLen(len(msg.Headers), func(i int) int {
+		h := &msg.Headers[i]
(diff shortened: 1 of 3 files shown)


dominicbarnes:
@kung-foo would you be willing to use this branch of kafka-go to see if it actually fixes the error you reported in #1050 ?

rhansen2 (approved):

PR #1165, opened 2023-07-17
Adds attempt number to log message when failing to write messages

jcarter3:
This is done to make it easier to write better error logger implementations. For instance, when using IAM auth for MSK, the first write will fail after credentials rotate. The connection is re-negotiated and the retry is successful, but the logs are filled with errors. This lets us ignore the error of the first attempt, but log subsequent ones.


rhansen2 (approved):
Thanks for the contribution!

PR #1119, opened 2023-04-19
KIP-546: Add Client Quota APIs

petedannemann:
KIP: https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client

API docs:
https://kafka.apache.org/protocol#The_Messages_DescribeClientQuotas

https://kafka.apache.org/protocol#The_Messages_AlterClientQuotas

I followed a similar code style to what was used for [alterconfigs](https://github.com/segmentio/kafka-go/blob/main/alterconfigs.go). Let me know if there is a better code style to use for this.

If this PR is too large to review I could split it into three pieces:
1. Adding support for float64 data types used in Kafka's protocol (quotas appear to be the only APIs that use this data type)
2. Add createquotas APIs
3. Add describequotas APIs

PR #1166, opened 2023-07-18
Support describeacls

petedannemann:
Support describeacls API

[API Docs](https://kafka.apache.org/protocol#The_Messages_DescribeAcls)

I based this PR on similar work done here https://github.com/segmentio/kafka-go/pull/1119

Additionally I fixed the tags for createacls protocol as I noticed those did not match the [API docs](https://kafka.apache.org/protocol#The_Messages_CreateAcls)


protocol/createacls/createacls.go
```diff
@@ -21,29 +21,37 @@ func (r *Request) Broker(cluster protocol.Cluster) (protocol.Broker, error) {
...
 
 type RequestACLs struct {
-	ResourceType        int8   `kafka:"min=v0,max=v2"`
-	ResourceName        string `kafka:"min=v0,max=v2"`
-	ResourcePatternType int8   `kafka:"min=v0,max=v2"`
-	Principal           string `kafka:"min=v0,max=v2"`
-	Host                string `kafka:"min=v0,max=v2"`
-	Operation           int8   `kafka:"min=v0,max=v2"`
-	PermissionType      int8   `kafka:"min=v0,max=v2"`
+	// We need at least one tagged field to indicate that v2+ uses "flexible"
```
rhansen2, line 24: Could you add a test for these using prototest?

createacl_test.go
```diff
@@ -41,9 +42,43 @@ func TestClientCreateACLs(t *testing.T) {
...
 	}
 
-	for _, err := range res.Errors {
+	for _, err := range createRes.Errors {
 		if err != nil {
 			t.Error(err)
 		}
 	}
+
+	describeResp, err := client.DescribeACLs(context.Background(), &DescribeACLsRequest{
```
petedannemann, line 51: Move this into describeacls_test

createacl_test.go
```diff
@@ -5,17 +5,18 @@ import (
 	"testing"
 
 	ktesting "github.com/segmentio/kafka-go/testing"
+	"github.com/stretchr/testify/assert"
```
petedannemann, line 8: Rename this file to createacls_test.go

createacl_test.go
```diff
@@ -5,17 +5,18 @@ import (
...
 	}
 
 	client, shutdown := newLocalClient()
 	defer shutdown()
 
-	res, err := client.CreateACLs(context.Background(), &CreateACLsRequest{
+	createRes, err := client.CreateACLs(context.Background(), &CreateACLsRequest{
 		ACLs: []ACLEntry{
 			{
 				Principal:           "User:alice",
```
petedannemann, line 22: Generate a unique principal name to avoid test collisions

protocol/createacls/createacls.go
petedannemann: Good call, I just added these

createacls_test.go
petedannemann, line 1: I renamed this file to match the file name convention of the corresponding source code file https://github.com/segmentio/kafka-go/blob/main/createacls.go

createacls_test.go
```diff
@@ -15,15 +15,18 @@ func TestClientCreateACLs(t *testing.T) {
 	client, shutdown := newLocalClient()
 	defer shutdown()
 
-	res, err := client.CreateACLs(context.Background(), &CreateACLsRequest{
+	topic := makeTopic()
```
petedannemann, line 18: I made these non-deterministic to avoid collisions with other tests

describeacls.go
```diff
@@ -0,0 +1,106 @@
...
+	Addr net.Addr
+
+	// Filter to filter ACLs on.
+	Filter ACLFilter
+}
+
+type ACLFilter struct {
+	ResourceTypeFilter        ResourceType
+	ResourceNameFilter        string
+	ResourcePatternTypeFilter PatternType
```
petedannemann, line 25: I'm not really sure how to handle this field since it was added in v1. We could use a pointer to make this nullable or have separate data structures and code paths for different versions. WDYT @rhansen2 ?

rhansen2: I think this works as is but we should probably have a comment explaining that it won't be used if the only supported API version is v0. I think the encoding will leave it out as appropriate based on the api version being used.

petedannemann: Sounds good, I added a comment

rhansen2 (approved):

Issue #1156, opened 2023-07-05
ProtocolMetadata v1 without OwnedPartitions


krvladislav:
**Describe the bug**

JoinGroup request sends ProtocolMetadata marked as version v1 
https://github.com/segmentio/kafka-go/blob/main/consumergroup.go#L998
but it never provides OwnedPartitions field which is expected by a protocol 
https://github.com/apache/kafka/blob/trunk/clients/src/main/resources/common/message/ConsumerProtocolSubscription.json#L34

This behavior can affect migration from kafka-go to another kafka client. Sarama provides guard 
https://github.com/Shopify/sarama/blob/main/consumer_group_members.go#L43 to skip this inconsistence silently. But franz-go throws an error which slowdown rebalance process.

**Kafka Version**

> * What version(s) of Kafka are you testing against?
2.6.0.0
> * What version of kafka-go are you using?
v0.4.40

**Expected Behavior**

ProtocolMetadata must contains OwnedPartitions or sets Version as v0

PR #1172, opened 2023-07-28
work around incorrect v1 metadata

rhansen2:
Potential fix for #1156

### DIFF
--- a/joingroup.go
+++ b/joingroup.go
@@ -2,7 +2,6 @@ package kafka
 
 import (
 	"bufio"
-	"bytes"
 	"context"
 	"fmt"
 	"net"
@@ -189,43 +188,6 @@ func (c *Client) JoinGroup(ctx context.Context, req *JoinGroupRequest) (*JoinGro
 	return res, nil
 }
 
-type groupMetadata struct {
-	Version  int16
-	Topics   []string
-	UserData []byte
-}
-
-func (t groupMetadata) size() int32 {
-	return sizeofInt16(t.Version) +
-		sizeofStringArray(t.Topics) +
-		sizeofBytes(t.UserData)
-}
-
-func (t groupMetadata) writeTo(wb *writeBuffer) {
-	wb.writeInt16(t.Version)
-	wb.writeStringArray(t.Topics)
-	wb.writeBytes(t.UserData)
-}
-
-func (t groupMetadata) bytes() []byte {
-	buf := bytes.NewBuffer(nil)
-	t.writeTo(&writeBuffer{w: buf})
-	return buf.Bytes()
-}
-
-func (t *groupMetadata) readFrom(r *bufio.Reader, size int) (remain int, err error) {
-	if remain, err = readInt16(r, size, &t.Version); err != nil {
-		return
-	}
-	if remain, err = readStringArray(r, remain, &t.Topics); err != nil {
-		return
-	}
-	if remain, err = readBytes(r, remain, &t.UserData); err != nil {
-		return
-	}
-	return
-}
-
 type joinGroupRequestGroupProtocolV1 struct {
 	ProtocolName     string
 	ProtocolMetadata []byte
@@ -237,6 +199,7 @@ func (t joinGroupRequestGroupProtocolV1) size() int32 {
 }
 
 func (t joinGroupRequestGroupProtocolV1) writeTo(wb *writeBuffer) {
+	fmt.Println("writing", t.ProtocolName, t.ProtocolName)
 	wb.writeString(t.ProtocolName)
 	wb.writeBytes(t.ProtocolMetadata)
 }
--- a/joingroup_test.go
+++ b/joingroup_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/segmentio/kafka-go/protocol/consumer"
 	ktesting "github.com/segmentio/kafka-go/testing"
 )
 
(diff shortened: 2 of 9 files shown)


joingroup.go
```diff
@@ -237,6 +199,7 @@ func (t joinGroupRequestGroupProtocolV1) size() int32 {
 }
 
 func (t joinGroupRequestGroupProtocolV1) writeTo(wb *writeBuffer) {
+	fmt.Println("writing", t.ProtocolName, t.ProtocolName)
```
petedannemann, line 202: Can we remove this log?

joingroup_test.go
```diff
@@ -130,12 +131,20 @@ func TestSaramaCompatibility(t *testing.T) {
 		//
 		// See consumer_group_members_test.go
 		//
+		groupMemberMetadataBad = []byte{
```
petedannemann, line 134: Is "bad" here just mean it's missing ownedpartitions and is the backwards compatible version? I wonder if there's a different name we can use here and in the test to make this more clear

protocol/consumer/consumer_test.go
```diff
@@ -33,8 +43,36 @@ func TestSubscription(t *testing.T) {
...
+	if s.Version != 1 {
+		t.Fatalf("expected version to be 1 got: %d", s.Version)
+	}
+
+	if !reflect.DeepEqual(s.Topics, []string{"one", "two"}) {
+		t.Fatalf("expected topics to be one [one two] got: %v", s.Topics)
+	}
+
+	if !bytes.Equal(s.UserData, []byte{0x01, 0x02, 0x03}) {
+		t.Fatalf(`expected user datea to be one [1 2 3] got: %v`, s.UserData)
```
petedannemann, line 76:
```suggestion
		t.Fatalf(`expected user data to be [1 2 3] got: %v`, s.UserData)
```

protocol/consumer/consumer_test.go
```diff
@@ -33,8 +43,36 @@ func TestSubscription(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if s.Version != 1 {
+		t.Fatalf("expected version to be 1 got: %d", s.Version)
+	}
+
+	if !reflect.DeepEqual(s.Topics, []string{"one", "two"}) {
+		t.Fatalf("expected topics to be one [one two] got: %v", s.Topics)
```
petedannemann, line 72:
```suggestion
		t.Fatalf("expected topics to be [one two] got: %v", s.Topics)
```

petedannemann (commented):
Could you do a functional test to verify if this fixes the issue we want it to?

protocol/consumer/consumer.go
```diff
@@ -1,21 +1,85 @@
...
+	return protocol.Unmarshal(b, 1, s)
+}
+
+func (s *Subscription) FromBytes(b []byte) error {
+	if len(b) < 2 {
+		return io.ErrUnexpectedEOF
+	}
+	version := readInt16(b[0:2])
+	err := protocol.Unmarshal(b, version, s)
+	if err != nil && version == 1 && errors.Is(err, io.ErrUnexpectedEOF) {
```
petedannemann, line 41: Should we use version >= 1 for consistency with sarama? https://github.com/IBM/sarama/commit/df0fb3d1f471b0177682afe4903e8ab5c99e1e65#diff-9b4870237f585e4fc1dfe4c35dd6d25ac430a50399762234c1c4f5097e012702R38

protocol/consumer/consumer.go
```diff
@@ -1,21 +1,85 @@
...
+type subscriptionBackwardsCompat struct {
+	Version  int16    `kafka:"min=v0,max=v1"`
+	Topics   []string `kafka:"min=v0,max=v1"`
+	UserData []byte   `kafka:"min=v0,max=v1,nullable"`
+}
+
+func (s *subscriptionBackwardsCompat) FromBytes(b []byte) error {
+	// This type is only intended to maintain backwards compatibility with
+	// this library and support other clients in the wild sending
+	// version 1 supscription data without OwnedPartitionsy
```
petedannemann, line 31:
```suggestion
	// version 1 subscription data without OwnedPartitions
```

protocol/consumer/consumer.go
```diff
@@ -1,21 +1,85 @@
...
+	Version         int16            `kafka:"min=v0,max=v3"`
+	Topics          []string         `kafka:"min=v0,max=v3"`
+	UserData        []byte           `kafka:"min=v0,max=v3,nullable"`
+	OwnedPartitions []TopicPartition `kafka:"min=v1,max=v3"`
+	GenerationID    int32            `kafka:"min=v2,max=v3"`
+	RackID          string           `kafka:"min=v3,max=v3,nullable"`
+}
+
+type subscriptionBackwardsCompat struct {
+	Version  int16    `kafka:"min=v0,max=v1"`
```
petedannemann, line 23: Style wise I'm on the fence of adding an addition type for this vs. handling this in Subscriptions's FromBytes method. What do you think?

protocol/consumer/consumer.go
```diff
@@ -1,21 +1,85 @@
...
+	return protocol.Unmarshal(b, 1, s)
+}
+
+func (s *Subscription) FromBytes(b []byte) error {
+	if len(b) < 2 {
+		return io.ErrUnexpectedEOF
+	}
+	version := readInt16(b[0:2])
+	err := protocol.Unmarshal(b, version, s)
+	if err != nil && version >= 1 && errors.Is(err, io.ErrUnexpectedEOF) {
```
petedannemann, line 41: Do you think we could either move the comment about backwards capability above here or make these checks in the if statement a separate function like `isBackwardsCompatible`? I had a bit of a hard time understanding it the way the code is structured now

PR #1174, opened 2023-07-28
Deleteacls support

petedannemann:
Support deleteacls API

[API Docs](https://kafka.apache.org/protocol#The_Messages_DeleteAcls)

### DIFF
--- a/deleteacls_test.go
+++ b/deleteacls_test.go
@@ -1,0 +1,91 @@
+package kafka
+
+import (
+	"context"
+	"testing"
+
+	ktesting "github.com/segmentio/kafka-go/testing"
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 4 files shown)


deleteacls_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"context"
+	"testing"
+
+	ktesting "github.com/segmentio/kafka-go/testing"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestClientDeleteACLs(t *testing.T) {
```
hhahn-tw, line 11: Is there any value in querying for a deleted ACL and expecting an empty/"not found" response in addition to checking for non-error in deleteResp?

petedannemann: Good idea, I just added this

Issue #1188, opened 2023-09-02
ErrNoProgress potentially misfiring?


zachxu42:
**Describe the bug**

We're using the library across hundreds of instances reading from an Kafka cluster. And the rate of this ErrNoProgress errors is alarmingly high, at around 10/s across all the components. I believe they're all from this [bit of logic](https://github.com/segmentio/kafka-go/blob/f4ca0b48296538eb5639027cd83f40a3828caf7d/conn.go#L1408-L1415) and I noticed a change was made to [handle this particular error more gracefully](https://github.com/segmentio/kafka-go/commit/8f063ce560a7dc30cf6e8e6e3cdb9b1b235a7698). But nonetheless it'd still close the connection and open a new one. Closing and opening tens of connections per second can add significant load to the Kafka cluster and impact performance.

I wonder why the logic would consider the connection to be corrupt so often. Is there potentially some concurrency issue/bug that's causing it to be false positive sometimes? 

**Kafka Version**

> * What version(s) of Kafka are you testing against?
2.5.1
> * What version of kafka-go are you using?
v0.4.12

**To Reproduce**
Run many consumers and observe the increase of reader errors or, in the older version of the library, the log `the kafka reader got an unknown error reading...multiple Read calls return no data or error`

**Expected Behavior**
Very few to zero ErrNoProgress due to corrupt connections even when there are many consumers.

**Observed Behavior**
Tens of reader errors due to ErrNoProgress which lead to frequent reader reconnections.

zachxu42:
Spent a day poking around, and found the most promising theory to be the following

* A goroutine couldn't finish reading all the bytes meant for it; Leaving unfinished bytes in the buffer would cause data corruption for the next reader goroutine, causing the ErrNoProgress error [here](https://github.com/segmentio/kafka-go/blob/main/conn.go#L1413).
* The first goroutine didn't finish consuming all the bytes in the read buffer because it hit `r.empty=true` [here](https://github.com/segmentio/kafka-go/blob/f4ca0b48296538eb5639027cd83f40a3828caf7d/message_reader.go#L125-L128)
* r.empty was set to true because `highWaterMark == offset` [here](https://github.com/segmentio/kafka-go/blob/main/conn.go#L862)
* When `highWaterMark == offset`, I tried logging `size` and `remaining`, and see that there's clearly bytes in the buffer. So this goroutine that's currently holding the read lock should not just give up. 
[...]

PR #1192, opened 2023-09-08
check remaining bytes when marking batch as empty

zachxu42:
Fixes [issue-1188](https://github.com/segmentio/kafka-go/issues/1188) and drastically reduce the number of reader errors and dials.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -868,7 +868,7 @@ func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
 
 	var msgs *messageSetReader
 	if err == nil {
-		if highWaterMark == offset {
+		if highWaterMark == offset && remain == 0 {
 			msgs = &messageSetReader{empty: true}
 		} else {
 			msgs, err = newMessageSetReader(&c.rbuf, remain)


petedannemann:
@zachxu42 this seems like a plausible fix but given that we've broken kafka-go before trying to fix this issue I would feel a lot before if we first found some way to:
1. Reproduce this issue in a test
2. Add test coverage to prevent us from breaking this in a similar way as we did in https://github.com/segmentio/kafka-go/pull/788

petedannemann:
We spent some more time comparing this to https://github.com/segmentio/kafka-go/pull/1177 and we think that is more of a root cause fix, but I think we'll still need to make some testing investments prior to proceeding with it

zachxu42:
Thanks @petedannemann for the comment. I believe #1177 is a different case, when the error is **not** nil. In this PR I could change it to do the same 

```
if highWaterMark == offset {
    msgs = &messageSetReader{empty: true}
    if remain > 0 {
		c.rbuf.Discard(remain)
    }
} else {
    msgs, err = newMessageSetReader(&c.rbuf, remain)
}
```

I think the issue is, whether there's an error or not, if we're about to return an empty message set, we need to make sure we clear the buffer if there's any remaining bytes, so the next reader doesn't see partial/corrupt data.

abitduck:
I have the same EOF issue. below is the tcpdump of the Kafka stream:

* Expected: The response **Message Set** not empty
* Problem: The **Message Set** is empty, and **EOF** occurs.

Request:
image

Response:
image

PR #1202, opened 2023-10-03
Fix config related stats for a default writer config

sharpyfox:
Looks like there is a mistake in getting some config parameters for stats in writer. Writer is using getters like `writeBackoffMin()` to get actual data, but stats are using direct config values with `w.WriteBackoffMin` which means that some stats are `0` by default


petedannemann (approved):
Thanks for your contribution @sharpyfox !

PR #1203, opened 2023-10-09
Implementation of ListPartitionReassignments API

bgranvea:


listpartitionreassignments.go
```diff
@@ -0,0 +1,135 @@
...
+	"context"
+	"net"
+	"time"
+
+	"github.com/segmentio/kafka-go/protocol/listpartitionreassignments"
+)
+
+// ListPartitionReassignmentsRequest is a request to the ListPartitionReassignments API.
+type ListPartitionReassignmentsRequest struct {
+	// Address of the kafka broker to send the request to.
```
petedannemann, line 13: I think we need to add tags to these structs like we've done elsewhere
https://github.com/segmentio/kafka-go/blob/f48706ef46bbf987630e4b8d728fa7f1a2f831b0/protocol/alterclientquotas/alterclientquotas.go#L11

protocol/listpartitionreassignments/listpartitionreassignments.go
```diff
@@ -0,0 +1,58 @@
...
+	// We need at least one tagged field to indicate that this is a "flexible" message
+	// type.
+	_ struct{} `kafka:"min=v0,max=v0,tag"`
+
+	TimeoutMs int32          `kafka:"min=v0,max=v0"`
+	Topics    []RequestTopic `kafka:"min=v0,max=v0,nullable"`
+}
+
+type RequestTopic struct {
+	Name             string  `kafka:"min=v0,max=v0"`
```
petedannemann, line 21: I think we need tags on these nested data structures in this file as well

petedannemann (commented):
This looks good, left one small comment

protocol/listpartitionreassignments/listpartitionreassignments.go
bgranvea: I must also add it on all response structures, right?

```
type ResponseTopic struct {
	Name       string              `kafka:"min=v0,max=v0"`
	Partitions []ResponsePartition `kafka:"min=v0,max=v0"`
}

type ResponsePartition struct {
	PartitionIndex   int32   `kafka:"min=v0,max=v0"`
	Replicas         []int32 `kafka:"min=v0,max=v0"`
	AddingReplicas   []int32 `kafka:"min=v0,max=v0"`
	RemovingReplicas []int32 `kafka:"min=v0,max=v0"`
}

```

petedannemann: Yes, those as well as `RequestTopic`

petedannemann (approved):

PR #1218, opened 2023-10-30
add methods for parsing and stringifying acl related resources

petedannemann:
Add methods for parsing ACL related resources and converting them between their kafka-go and human readable versions

This is taken from Sarama https://github.com/IBM/sarama/blob/3ca69a884964f10c228a382d184b0d5005c342e1/acl_types.go#L35

### DIFF
--- a/createacls_test.go
+++ b/createacls_test.go
@@ -50,3 +50,37 @@ func TestClientCreateACLs(t *testing.T) {
 		}
 	}
 }
+
+func TestACLPermissionTypeMarshal(t *testing.T) {
+	for i := ACLPermissionTypeUnknown; i <= ACLPermissionTypeAllow; i++ {
+		text, err := i.MarshalText()
+		if err != nil {
+			t.Errorf("couldn't marshal %d to text: %s", i, err)
+		}
+		var got ACLPermissionType
+		err = got.UnmarshalText(text)
+		if err != nil {
+			t.Errorf("couldn't unmarshal %s to ACLPermissionType: %s", text, err)
+		}
+		if got != i {
+			t.Errorf("got %d, want %d", got, i)
+		}
+	}
+}
+
+func TestACLOperationTypeMarshal(t *testing.T) {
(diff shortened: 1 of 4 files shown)


createacls_test.go
```diff
@@ -50,3 +50,37 @@ func TestClientCreateACLs(t *testing.T) {
 		}
 	}
 }
+
+func TestACLPermissionTypeMarshal(t *testing.T) {
+	for i := ACLPermissionTypeUnknown; i <= ACLPermissionTypeAllow; i++ {
```
hhahn-tw, line 55: It feels a little weird to hard code the end of the range (`ACLPermissionTypeAllow`) here rather than, say, max(mapping). Is this list not expected to expand?

resource.go
```diff
@@ -15,6 +20,48 @@ const (
 	ResourceTypeDelegationToken ResourceType = 6
 )
 
+func (rt ResourceType) String() string {
+	mapping := map[ResourceType]string{
+		ResourceTypeUnknown:         "Unknown",
+		ResourceTypeAny:             "Any",
+		ResourceTypeTopic:           "Topic",
+		ResourceTypeGroup:           "Group",
```
hhahn-tw, line 28: Are you missing ResourceTypeBroker here?

petedannemann: `ResourceTypeBroker` and `ResourceTypeCluster` have the same value associated with them and a map cannot have duplicate keys
https://segment.atlassian.net/browse/DP-1858

createacls_test.go
petedannemann: Yeah it's not ideal, but AFAIK there is no clear way to get the minimum and maximum values of an "enum" in Go. This test was taken from Sarama https://github.com/IBM/sarama/blob/82f0e48b0e2b6dfefc3088fe5e45195f6cc461d8/acl_types_test.go#L8

resource.go
hhahn-tw: 👍 maybe a comment to that effect for future code reviewers

hhahn-tw (approved):

PR #1228, opened 2023-11-10
panic issue

zmz-kevin:


zmz-kevin:
The kafka version is 2.4.1, and we are using compact topic. we have five partition, but now only one partition has data, and when we start to consume data  from kafka, I set the startoffset to FirstOffset, then panic

PR #1232, opened 2023-11-15
Add Chunk Size to RR Balancer (Increased Batching Ability)

erushing:
The motivation here is to try to get Kafka-Go's RR Balancer to batch more aggressively.  The RR Balancer naively iterates through the available partitions, putting a single message on each of them until the batch timeout.  If it could put x messages on each partition before moving onto the next one, it would still be evenly distributed, but batch better.

The coding approach was to bring in the same chunking mechanism our internal Bulrush library uses, but simplify it and adapt it for the code-style of Kafka-Go.

### DIFF
--- a/balancer.go
+++ b/balancer.go
@@ -4,6 +4,7 @@ import (
 	"hash"
 	"hash/crc32"
 	"hash/fnv"
+	"math"
 	"math/rand"
 	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


balancer.go
```diff
@@ -54,6 +55,35 @@ func (rr *RoundRobin) balance(partitions []int) int {
...
+
+	mutex   sync.RWMutex
+	counter *uint64
+}
+
+// Balance satisfies the Balancer interface.
+func (rr *ChunkedRoundRobin) Balance(msg Message, partitions ...int) int {
+	rr.mutex.RLock()
+	if rr.counter == nil {
+		var nextToInsert uint64 = math.MaxUint64 // so that first increment overflows it back to 0
```
petedannemann, line 72: We need to lock the mutex for writes within this if statement

petedannemann: I think it might make more sense to just use `rr.mutex.Lock()` above and `rr.mutex.Unlock()` below

balancer.go
```diff
@@ -54,6 +55,35 @@ func (rr *RoundRobin) balance(partitions []int) int {
 	return partitions[offset%length]
 }
 
+// ChunkedRoundRobin is a Balancer implementation that equally distributes messages
+// across all available partitions, but puts greater emphasis on batching by a chunk size
+// within a shorter time period than is possible via the regular RoundRobin Balancer.
+type ChunkedRoundRobin struct {
+	chunkSize int
```
petedannemann, line 62: I think we want `chunkSize` to be public so that users can configure it

balancer.go
erushing: Yeah, this is where I was thinking copying bulrush was a mistake.  Does it make sense to do the same locking as LeastBytes?
`
lb.mutex.Lock()

defer lb.mutex.Unlock()
`

petedannemann: Yeah adding the defer in for unlocks is probably the safest option. It will add some additional locking time but that won't be meaningful

balancer.go
erushing: Yeah, this is where I thought the test showed you could pass in a value to the struct, but I would find this out for real when I tried to actually test this with a kafka-go based service.  I see that Bulrush uses a setter for this, so I'm sure I'm off base with the way I did this.

I had this worker in mind where I want to be able to pass in a chunk size as they pass in RR as a balancer.  
https://github.com/segmentio/identity/blob/2d04b8f5a16d235453c922e1e9d1ff7ca1b2a92b/identity-resolver/worker.go#L290C21-L290C21

petedannemann: The test is in the same package as the balancer so it is able to reference private fields. That won't be the case for users of the balancer. We don't use setters elsewhere in this kafka-go package and just use public fields for fields that we want to expose to users so I'd be inclined to stick with that convention

erushing:
This was tested on a live application and showed increased batching, despite fairly low throughput through each producer and a low batch timeout value (10ms)
Screen Shot 2023-11-27 at 11 39 26 AM
Screen Shot 2023-11-27 at 11 39 36 AM

PR #1233, opened 2023-11-16
Add RawProduce API

apacker:
This PR adds a new `RawProduce` API to the client. The background/context for this API is: we have an internal system that does buffering of Kafka messages on disk, and needs to flush that data to Kafka without having to decode/re-encode the data and incur a 2x cost of compression/encoding. In this context, it would be optimal to be able to pass a raw/pre-encoded `RecordSet` to the Produce API, hence the introduction of a new `RawProduce` interface.

Notes:
* The existing `ProduceResponse` type is reused as there's no reason to introduce a new response type.
* Since request/response types and their resulting encodings are linked directly to an API key, I've introduced a mechanism to be able to override types via a `OverrideTypeMessage` interface.


rawproduce.go
```diff
@@ -0,0 +1,103 @@
...
+// to a topic partition. The request contains a pre-encoded/raw record set.
+type RawProduceRequest struct {
+	// Address of the kafka broker to send the request to.
+	Addr net.Addr
+
+	// The topic to produce the records to.
+	Topic string
+
+	// The partition to produce the records to.
+	Partition int
```
petedannemann, line 24: I know our existing `ProduceRequest` only supports producing to one partition at a time, but should we take this opportunity to add support for producing to multiple partitions in one request with `RawProduceRequest`? This could result in a large performance improvement and reduction in load on Kafka clusters for clients that can accommodate for it

petedannemann (commented):
🚀 great work, two small comments for now. I'll think this over a bit more and re-review later this week

rawproduce.go
petedannemann: If that is too much unrelated work for this feature we could always add some kind of `MultiRawProduceRequest` later on if needed

protocol/protocol.go
```diff
@@ -213,6 +213,37 @@ func Register(req, res Message) {
...
+
+const (
+	RawProduceOverride OverrideTypeKey = 0
+)
+
+var overrideApiTypes [numApis]map[OverrideTypeKey]apiType
+
+func RegisterOverride(req, res Message, key OverrideTypeKey) {
+	k1 := req.ApiKey()
+	k2 := req.ApiKey()
```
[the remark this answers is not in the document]
apacker: Yes, good catch. Fixed.

petedannemann (approved):

PR #1239, opened 2023-11-21
Optimize Conn.SetRequiredAcks parameter type

ACV-er:
Better readability and more intelligent code completion

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1269,9 +1269,9 @@ func (c *Conn) writeCompressedMessages(codec CompressionCodec, msgs ...Message) 
 
 // SetRequiredAcks sets the number of acknowledges from replicas that the
 // connection requests when producing messages.
-func (c *Conn) SetRequiredAcks(n int) error {
+func (c *Conn) SetRequiredAcks(n RequiredAcks) error {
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -1269,9 +1269,9 @@ func (c *Conn) writeCompressedMessages(codec CompressionCodec, msgs ...Message)
 
 // SetRequiredAcks sets the number of acknowledges from replicas that the
 // connection requests when producing messages.
-func (c *Conn) SetRequiredAcks(n int) error {
+func (c *Conn) SetRequiredAcks(n RequiredAcks) error {
 	switch n {
-	case -1, 1:
+	case RequireNone, RequireOne, RequireAll:
```
vasily-kirichenko, line 1274: Having `RequiredAcks` declared as
```go
const (
	RequireNone RequiredAcks = 0
	RequireOne  RequiredAcks = 1
	RequireAll  RequiredAcks = -1
)
```
I don't understand why you added `RequireNone` case here? I mean, to preserve behavior, the code should be
```go
case RequireAll, RequireOne:
...
```

PR #1247, opened 2023-12-07
logger: ensure compatibility with new log/slog package

kevinburkesegment:
The API for the logger changes only slightly and should continue to be backwards compatible for any existing code.


kevinburkesegment:
I've fixed everything with CircleCI except the linters; the error message on the linter is incomprehensible.

go.mod
```diff
@@ -1,13 +1,22 @@
 module github.com/segmentio/kafka-go
 
-go 1.15
+go 1.18
```
petedannemann, line 3: I think we need a separate PR for upgrading the Go version. Since kafka-go is so widely used this will be a breaking change for some users and the more explicit we can make that the less of a surprise it will be to users

.circleci/config.yml
```diff
@@ -2,19 +2,19 @@ version: 2
...
   # The kafka 0.10 tests are maintained as a separate configuration because
   # kafka only supported plain text SASL in this version.
   kafka-010:
-    working_directory: &working_directory /go/src/github.com/segmentio/kafka-go
+    working_directory: &working_directory /home/circleci/src/github.com/segmentio/kafka-go
     environment:
       KAFKA_VERSION: "0.10.1"
     docker:
-    - image: circleci/golang
+    - image: cimg/go:1.21
```
petedannemann, line 17: I think it may be best to test against the same Go version that is in our `go.mod` file

petedannemann:
> I've fixed everything with CircleCI except the linters; the error message on the linter is incomprehensible.

If you'd prefer I can handle the Go version upgrade and linting fixes. Then this PR will just be the change you actually want to make. Let me know what's best for you

kevinburkesegment:
> If you'd prefer I can handle the Go version upgrade and linting fixes. 

Sure, if you have time that would be great.

PR #1270, opened 2024-03-05
Allow null replicas in RequestPartition

leizor:
The `replicas` slice was declared as an empty slice which precludes passing a nil slice in the case of wanting to cancel an ongoing assignment.


leizor:
The `TestReaderReadCompactedMessage` test is failing but it passes on my local machine and I think is unrelated to the change in this PR.

Issue #615, opened 2021-03-09
A way to retrieve what partitions where consumer is signed


romani-felipe:
**Describe the solution you'd like**

Some way to provide information of what partitions this client is signed.

I think can be a method inside ReaderKafka like this:
```go
r := kafka.ReaderConfig{
		Brokers: []string{"localhost:9092"},
		Topic:   "topic.test",
		GroupID: "test-group",
	}
r.SignedPartitions(func(partitions) {
     fmt.Println(partitions) // partitions parameter is a slice []int{} because a consumer can sign more than one partition
}) 
// It´s can be a closure function hook for when occur a rebalance it can do some action again, print, log ou something else
```

This is helpful for consumer groups to debug who is assigned to what partition.

The Java client provides this kind of information.

achille-roussel:
Hello @romani-felipe, thanks for opening this issue!

The change you're suggesting sounds interesting, would you be able to submit a pull request implementing the feature?

Happy to discuss implementation details in this issue as well if you have questions about how to go about making the change.

romani-felipe:
@achille-roussel  Sure!

In next days i will start the implementation and will update this issue.

PR #1305, opened 2024-07-02
#615 - listener for assignments

nachogiljaldo:
Closes https://github.com/segmentio/kafka-go/issues/615

Adds a listener to send re-assignments to a listener.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -304,6 +304,23 @@ func (r *Reader) run(cg *ConsumerGroup) {
 		for attempt := 1; attempt <= r.config.MaxAttempts; attempt++ {
 			gen, err = cg.Next(r.stctx)
 			if err == nil {
+				if r.config.AssignmentListener != nil {
+					assignments := make([]GroupMemberTopic, 0, len(gen.Assignments))
+					for topic, partitions := range gen.Assignments {
+						assignedPartitions := make([]int, 0, len(partitions))
+						for _, partition := range partitions {
+							assignedPartitions = append(assignedPartitions, partition.ID)
+						}
+						sort.Slice(assignedPartitions, func(i, j int) bool {
+							return assignedPartitions[i] < assignedPartitions[j]
+						})
+						assignments = append(assignments, GroupMemberTopic{
+							Topic:      topic,
+							Partitions: assignedPartitions,
+						})
+					}
+					r.config.AssignmentListener(assignments)
+				}
 				break
 			}
 			if errors.Is(err, r.stctx.Err()) {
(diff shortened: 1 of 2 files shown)


nachogiljaldo:
@achille-roussel I saw you commented on https://github.com/segmentio/kafka-go/issues/615 . Is this something you could review?

isaacd9:
I don't think adding this is a good idea. Instead, you should probably use the [ConsumerGroup](https://pkg.go.dev/github.com/segmentio/kafka-go#ConsumerGroup) API to listen for a new generation and construct a new reader.

nachogiljaldo:
Thanks for the feedback @isaacd9 ! Could you elaborate?

I had a quick look at the [NewReader](https://github.com/segmentio/kafka-go/blob/main/reader.go#L629). Also it seems to use `run` which is private, so it's not possible to do it that way without duplicating even more code? Same for `getTopics()` while building the ConsumerGroupConfig.

Maybe, instead, a modification safe version of ConsumerGroup should be exposed (that only has Next but not Close) in the `Reader`? That allows to preserve the encapsulation of the logic to create Readers which IMHO seems important and avoids unnecessary duplication on all clients that would need this.

Something like:
```
type ConsumerGroupGenerationListener interface {
  Next(ctx context.Context) (*Generation, error)
}

...

func (r *Reader) ConsumerGroupGenerationListener(ctx context.Context) (ConsumerGroupGenerationListener, error) {
	if !r.useConsumerGroup() {
		return nil, errOnlyAvailableWithGroup
	}
        return r.consumerGroupGenerationListener, nil
}
```

nachogiljaldo:
@petedannemann , sorry for the direct ping but you seem to have reviewed some of the last PRs. I wonder if this is something you could provide feedback on the approach (or the alternatives provided by Isaac or my counter proposal).

nachogiljaldo:
heya @erikdw / @jkoelker apologies for the direct ping, but this has been laying around for a while and I would need some guidance (as you seem to have approved the last merged PRs)

If you're indeed maintainers of this library, are you ok with this approach? Would you prefer the one outlined in https://github.com/segmentio/kafka-go/pull/1305#issuecomment-2205206085 or a different one? Happy to take on those suggestions

nachogiljaldo:
@jkoelker thanks for the prompt reply!

Fair points, I updated the PR description and first commit message to be more informative, I hope that makes sense.

Also removed the sorting as you suggested.

As per the specifics of the implementation, since you said you're not familiar with the code or kafka, would you know who could help providing feedback on the approach / review this PR?

PR #1309, opened 2024-07-11
Makefile: use docker compose not docker-compose

kevinburkesegment:
As part of the Docker binary this will be updated more often and has
better support and newer features than the separate "docker-compose"
binary.


erikdw (approved):
LGTM.  Those build failures have been there for awhile AFAICR

Issue #1307, opened 2024-07-08
Closing kafka Writer during WriteMessages causes a potential hang


ionutboangiu:
**Describe the bug**

Closing the kafka.Writer while attempting to write a message can cause the process to hang indefinitely. This issue is resolved by closing the writer again.

**Kafka Version**

- **Kafka Version:** 3.7.0
- **kafka-go Version:** current [main branch](https://github.com/segmentio/kafka-go/tree/7b9c99dd6562f80bd3fc447305115029a7154382)

**To Reproduce**

```go
package main

import (
	"context"
	"log"
	"net"
	"strconv"
	"time"

	"github.com/segmentio/kafka-go"
)

func main() {
	brokerURL := "localhost:9092"
	topic := "test"
	createTopic(topic, brokerURL)
	writer := &kafka.Writer{
		Addr:   kafka.TCP(brokerURL),
		Topic:  topic,
		Logger: log.Default(),

		Transport: &kafka.Transport{
			Dial: (&net.Dialer{
				Timeout: 3 * time.Second,
			}).DialContext,
		},
	}
	done := make(chan struct{})
	go func() {
		time.Sleep(1 * time.Millisecond)
		writer.Close()
		close(done)
	}()
	if err := writer.WriteMessages(context.Background(), kafka.Message{
		Value: []byte("payload"),
	}); err != nil {
		log.Fatal(err)
	}

	<-done
	log.Print("does not reach this point")
}

func createTopic(name, brokerURL string) {
	conn, err := kafka.Dial("tcp", brokerURL)
	if err != nil {
		panic(err)
	}
	defer conn.Close()

	controller, err := conn.Controller()
	if err != nil {
		panic(err)
	}
	var controllerConn *kafka.Conn
	controllerConn, err = kafka.Dial("tcp", net.JoinHostPort(controller.Host, strconv.Itoa(controller.Port)))
	if err != nil {
		panic(err)
	}
	defer controllerConn.Close()

	err = controllerConn.CreateTopics(kafka.TopicConfig{
		Topic:             name,
		NumPartitions:     1,
		ReplicationFactor: 1,
	})
	if err != nil {
		panic(err)
	}
}
```

**Expected Behavior**

Expecting `kafka.Writer.Close()` not to hang.

**Observed Behavior**

Hangs here: [w.group.Wait()](https://github.com/segmentio/kafka-go/blob/7b9c99dd6562f80bd3fc447305115029a7154382/writer.go#L573)

[...]

PR #1312, opened 2024-07-23
Fix: Closing kafka Writer during WriteMessages causes a potential hang #1307

iamgoroot:
Fixes issue #1307 with closing Writer while WriteMessages() is ongoing

https://github.com/segmentio/kafka-go/issues/1307

Cause of the issue:
* Close takes the lock and goes through writers created in batchMessages closing them in a loop
* Close hangs in wait while ongoing messages are processed
* meanwhile in WriteMessages there's only one check for writer being closed. If WriteMessages is past that check while closing new writers are created that will not be released in Close since it's already in a wait

Solution: 
WriteMessages should fail with error in case the writer is closed upon acquiring mutex second time and before creating new writers (or map of writers) that won't be closed.


nachogiljaldo (approved):
LGTM

Issue #1308, opened 2024-07-08
Rebalances can lead to wrong lag being reported


nachogiljaldo:
**Describe the bug**

When commits are not immediate and/or some events can have a relatively high processing time and topics substain a low traffic, it can lead to lag being reported even if events were processed and committed.

**Kafka Version**

3.6.x

**To Reproduce**

This test reprocuces the behavior:
https://github.com/nachogiljaldo/kafka-go/blob/do_not_commit_offset_of_not_owned_partitions/reader_test.go#L1184

The situation is:
- consumer A has partitions 1 and 2
- consumer A receives msg 1 for partition 2
- consumer B appears and gets assigned partition 1
- consumer B receives msg 1 and 2 and commits them
- consumer A finishes consuming msg 1 and commits it --> offset goes back to 1, but consumer B does not know it, so it keeps on asking for offset 2

**Expected Behavior**

There are 2 things I would expect:
- committing a message for a partition that is not assigned to the current consumer fails
- if that happens, we read all events that now became uncommited but that doesn't happen either

**Observed Behavior**

We do not get the missing message which leads to a fake lag when the traffic has little traffic.

nachogiljaldo:
Ok, I believe I found the culprit and the solution could come from 2 places.

There's 2 problems the way I see it:
a) reader does not verify the partitions that are going to be committed against the generation's Assignments. This opens the gate for a race-condition between the consumer that received a partition and commits pending to be done in the old generation.
b) That would not be a big deal if the offset was not cached at the connection level. But because it keeps as offset the last offset it has seen, if the offset "goes back in time" due to a) the connection is not aware of and it does not re-read the messages, which appear as lagging until new events are sent / processed and committed.

PR #1330, opened 2024-09-15
Do not commit offsets for past generations if partition not owned

nachogiljaldo:
This is a skeleton PR aiming for discussion related to https://github.com/segmentio/kafka-go/issues/1308

In that issue, I describe how rebalances can lead to a situation where the consumer receiving the partition does not read messages in spite of lag existing (because conn's offset is > than the broker's current offset and there's no new messages coming).

I was aiming to fix it by avoiding commits for past generations, to do so, I add the generation id to the message and when committing, I log an error instead of adding it to the stash if it belongs to a generation < than current.

This has the risk of losing valid commits when the new generation comes (as its associated generation < latest generation.id), however it works because a new generation ends up creating new connections which means uncommitted events are duplicated. It is not that I love it because it works incidentally but it does work as my test shows.

If you have a different / better approach I am happy to explore it if you do explain it a bit.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -65,6 +65,8 @@ type Conn struct {
 	apiVersions atomic.Value // apiVersionMap
 
 	transactionalID *string
+
+	generationId int32
 }
 
 type apiVersionMap map[apiKey]ApiVersion
@@ -182,6 +184,7 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 		offset:          FirstOffset,
 		requiredAcks:    -1,
 		transactionalID: emptyToNullable(config.TransactionalID),
+		generationId:    -1,
 	}
 
 	c.wb.w = &c.wbuf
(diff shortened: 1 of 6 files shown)


conn.go
```diff
@@ -17,6 +17,8 @@ import (
 var (
 	errInvalidWriteTopic     = errors.New("writes must NOT set Topic on kafka.Message")
 	errInvalidWritePartition = errors.New("writes must NOT set Partition on kafka.Message")
+
+	undefinedGenerationId int32 = -1
```
logrusorgru, line 21: It should be `const`, not `var`.

nachogiljaldo: thanks for the feedback, I applied the suggestion, I would also appreciate feedback on the general approach as well :pray:

logrusorgru:
@nachogiljaldo , please, if it’s not a problem for you, apply this patch https://github.com/nachogiljaldo/kafka-go/pull/1 . (No changes in logic).

M0rdecay:
Hello! Any chances new release will be published with this patch?

PR #1287, opened 2024-05-02
fix: avoid return io.UnexpectedEOF error while remain is zero

acynothia:
A lot of unexpected EOF error when reading from partition cause by `newMessageSetReader` but remain is zero.

resolves #1286

PR #1342, opened 2024-10-20
fix: A lot of unexpected EOF error when reading from partition

jarvanstack:
error log like: 
> no messages received from kafka within the allocated time for partition 0 of cysj2-UpdateUsername at offset 472919: [7] Request Timed Out: the request exceeded the user-specified time limit in the request


apaatsio:
Duplicate of https://github.com/segmentio/kafka-go/pull/1287

PR #1349, opened 2024-11-15
Fixes some flaky tests in the build as well as the case when tests start before kafka is ready

nachogiljaldo:
This PR fixes at least 1 flaky test and a scenario when kafka is slow to start and the first tests in the suite just fail.

The cases are various but we have:
- topic creating not waiting before sending and reading
- timeouts
- kafka not started

I don't think this has solved them all, but I managed to get a local build passing as well as in CI, which I could not achieve without these changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,9 @@ jobs:
         paths:
         - /go/pkg/mod
     - run:
+        name: Wait for kafka
+        command: ./scripts/wait-for-kafka.sh
+    - run:
         name: Test kafka-go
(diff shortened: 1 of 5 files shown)


nachogiljaldo:
@jkoelker / @erikdw apologies for pinging you directly, but I see you reviewed the last 4 PRs that were merged.

I found that it was impossible to get a green build while working on ht ttps://github.com/segmentio/kafka-go/pull/1330 (which would also need some review) and decided to add this PR to address those flaky tests.

I see `main` is currently broken, so hopefully this would help with it.

Issue #656, opened 2021-05-07
Question: How to unit tests my Golang with Kafka-Go


iqDF:
### Description

As a background issue, I have a codebase rules where we aim for higher test coverage (Sonar Code Cov) pipeline. Now I'm trying to integrate kafka pub/sub as part of the solution but realise I cannot write a unit tests.

I wonder if you support doing mini kafka or mock kafka for unit testing? Would appreciate your help for unit testing Kafka.

dominicbarnes:
It sounds like you're asking for some sort of fake/mock server-side implementation of Kafka that can be used for unit testing. That's a pretty huge amount of lift that we don't have the bandwidth to commit to working on, especially because the benefits of such a system are only realized if we continue to keep up-to-date with every feature that Kafka adds over time, which becomes even more work.

That being said, you can follow a similar pattern to what we do for testing here, which is to use `docker-compose` to spin up a local Kafka and then operate on that directly. Depending on what exactly you're trying to isolate and test, this could be a reasonable approach to take.

If you're trying to avoid a full integration test, you could always create interfaces for your Kafka interactions and abstract behind that instead. Another possibility is to abstract your code for handling the messages themselves away from the book-keeping of fetching and committing messages, possibly using interfaces to also de-couple those components.

tl;dr: creating a fully-functional Kafka server mock is really hard, using interfaces in Go is probably the way to go

PR #1369, opened 2025-02-16
Create ConnIface interface and generate mock with mockgen

cpeliciari:
I took a look at https://github.com/segmentio/kafka-go/issues/656 and became curious as to why we don’t simply ship a generic mock generated with mockgen. In the future, if the community feels it’s necessary, we can improve this mock. It’s also very important to note that we already have a Fake for the library, which helps developers create their unit tests. If you don’t see the need for this PR, feel free to close this was just a suggestion for a mock.

### Changes:
- Created the ConnIface interface aggregating all public methods of *Conn.
- Generated a mock implementation for ConnIface using mockgen for easier testing.


petedannemann:
We prefer to test against an actual Kafka implementation. It's less code to maintain and gives us more reassurance on the accuracy. Thank you for the suggestion though

PR #1376, opened 2025-04-02
example groupID case fix

gam6itko:
In consumer code we use `groupID` env but in docker-compose `GroupID` passed.


petedannemann (approved):
Nice catch thank you!

PR #1379, opened 2025-04-30
Kafka 4.0 support

maxwolf8852:
Updated deprecated protocol fields for Apache Kafka 4.0 support:
 - **JoinGroup**: V1 -> V2
 - **CreateTopics**: V0 -> V2
 - **DeleteTopics** V0 -> V1
 
While `Conn` does not use `Client`, the "negotiateVersion" mechanism has been reused for backward compatibility.


consumergroup.go
```diff
@@ -555,7 +555,7 @@ func (g *Generation) partitionWatcher(interval time.Duration, topic string) {
 type coordinator interface {
 	io.Closer
 	findCoordinator(findCoordinatorRequestV0) (findCoordinatorResponseV0, error)
-	joinGroup(joinGroupRequestV1) (joinGroupResponseV1, error)
+	joinGroup(joinGroupRequestV1) (joinGroupResponse, error)
```
petedannemann, line 558: I think we should update joinGroupRequestV1 to be called joinGroupRequest since it actually is now not necessarily V1

createtopics.go
```diff
@@ -262,94 +262,146 @@ func (t createTopicsRequestV0Topic) writeTo(wb *writeBuffer) {
...
 	return
 }
 
 // See http://kafka.apache.org/protocol.html#The_Messages_CreateTopics
-type createTopicsResponseV0 struct {
-	TopicErrors []createTopicsResponseV0TopicError
+type createTopicsResponse struct {
+	v apiVersion
+
+	ThrottleTime int32 // v1+
```
petedannemann, line 347:
```suggestion
	ThrottleTime int32 // v2+
```

createtopics.go
```diff
@@ -262,94 +262,146 @@ func (t createTopicsRequestV0Topic) writeTo(wb *writeBuffer) {
...
+	TopicErrors  []createTopicsResponseTopicError
 }
 
-func (t createTopicsResponseV0) size() int32 {
-	return sizeofArray(len(t.TopicErrors), func(i int) int32 { return t.TopicErrors[i].size() })
+func (t createTopicsResponse) size() int32 {
+	sz := sizeofArray(len(t.TopicErrors), func(i int) int32 { return t.TopicErrors[i].size() })
+	if t.v >= v1 {
```
petedannemann, line 353:
```suggestion
	if t.v >= v2 {
```

createtopics.go
```diff
@@ -262,94 +262,146 @@ func (t createTopicsRequestV0Topic) writeTo(wb *writeBuffer) {
...
+		sz += sizeofInt32(t.ThrottleTime)
+	}
+	return sz
 }
 
-func (t createTopicsResponseV0) writeTo(wb *writeBuffer) {
+func (t createTopicsResponse) writeTo(wb *writeBuffer) {
+	if t.v >= v1 {
```
petedannemann, line 360:
```suggestion
	if t.v >= v2 {
```

createtopics.go
```diff
@@ -262,94 +262,146 @@ func (t createTopicsRequestV0Topic) writeTo(wb *writeBuffer) {
...
+		topic := createTopicsResponseTopicError{v: t.v}
+		if fnRemain, fnErr = (&topic).readFrom(r, size); fnErr != nil {
 			return
 		}
 		t.TopicErrors = append(t.TopicErrors, topic)
 		return
 	}
-	if remain, err = readArrayWith(r, size, fn); err != nil {
+	remain = size
+	if t.v >= v1 {
```
petedannemann, line 376:
```suggestion
	if t.v >= v2 {
```

createtopics_test.go
```diff
@@ -160,9 +160,9 @@ func TestClientCreateTopics(t *testing.T) {
 	}
 }
 
-func TestCreateTopicsResponseV0(t *testing.T) {
-	item := createTopicsResponseV0{
-		TopicErrors: []createTopicsResponseV0TopicError{
+func TestCreateTopicsResponseV1(t *testing.T) {
```
petedannemann, line 163:
```suggestion
func TestCreateTopicsResponse(t *testing.T) {
```

deletetopics.go
```diff
@@ -87,31 +87,47 @@ func (t deleteTopicsRequestV0) writeTo(wb *writeBuffer) {
...
+	}
+	return sz
 }
 
-func (t *deleteTopicsResponseV0) readFrom(r *bufio.Reader, size int) (remain int, err error) {
+func (t *deleteTopicsResponse) readFrom(r *bufio.Reader, size int) (remain int, err error) {
 	fn := func(withReader *bufio.Reader, withSize int) (fnRemain int, fnErr error) {
 		var item deleteTopicsResponseV0TopicErrorCode
-		if fnRemain, fnErr = (&item).readFrom(withReader, withSize); err != nil {
+		if fnRemain, fnErr = (&item).readFrom(withReader, withSize); fnErr != nil {
```
petedannemann, line 109: Interesting that we were checking the wrong error here before 😅

deletetopics.go
```diff
@@ -146,16 +162,24 @@ func (t deleteTopicsResponseV0TopicErrorCode) writeTo(wb *writeBuffer) {
 // deleteTopics deletes the specified topics.
 //
 // See http://kafka.apache.org/protocol.html#The_Messages_DeleteTopics
-func (c *Conn) deleteTopics(request deleteTopicsRequestV0) (deleteTopicsResponseV0, error) {
-	var response deleteTopicsResponseV0
-	err := c.writeOperation(
+func (c *Conn) deleteTopics(request deleteTopicsRequestV0) (deleteTopicsResponse, error) {
```
petedannemann, line 165: Seems like we should rename `deleteTopicsRequestV0` to `deleteTopicsRequest` since the version is not deterministic

joingroup_test.go
```diff
@@ -218,7 +218,9 @@ func TestMemberMetadata(t *testing.T) {
 }
 
 func TestJoinGroupResponseV1(t *testing.T) {
-	item := joinGroupResponseV1{
+	const version = v1
```
petedannemann, line 221: Could we add a similar test scenario for when the version is v0?

petedannemann (commented):
Thank you for this! The main feedback is it looks like ThrottleTime was actually added to CreateTopics responses in v2, not v1. I think if possible it would be good to ensure our tests cover both suported versions with this change. Additionally, we should add Kafka 4.0 to our circleci testing matrix to make sure the tests do actually succeed there

createtopics.go
```diff
@@ -262,94 +262,146 @@ func (t createTopicsRequestV0Topic) writeTo(wb *writeBuffer) {
...
 		return
 	}
 
 	return
 }
 
-func (c *Conn) createTopics(request createTopicsRequestV0) (createTopicsResponseV0, error) {
-	var response createTopicsResponseV0
+func (c *Conn) createTopics(request createTopicsRequest) (createTopicsResponse, error) {
+	version, err := c.negotiateVersion(createTopics, v0, v2)
```
petedannemann, line 389: I wonder if not having v1 here will be an issue. I'm not entirely sure what happens if we skip a supported version when using negotiateVersion

deletetopics.go
maxwolf8852: found and fixed two more places

petedannemann:
Looks like there are some additional issues with tests. I think these may have started with kafka 3.X and I started trying to dig into them here https://github.com/segmentio/kafka-go/pull/1381

petedannemann:
Thank you for addressing my feedback though, this PR looks good I just need to figure out the test issues before we can merge this

petedannemann:
What I am seeing in my PR is a lot of these errors in Kafka's logs that cause many requests to fail during tests
> Timeout failed due to Java heap space
I've tried adding JVM_OPTS for increasing the heap size and I am still seeing the errors even with a huge. I wonder if Kafka 4.0 is not stable yet, it seems like the heap size cannot be adjusted as expected and the memory usage is not predictable

maxwolf8852:
> What I am seeing in my PR is a lot of these errors in Kafka's logs that cause many requests to fail during tests
> 
> > Timeout failed due to Java heap space
> 
> I've tried adding JVM_OPTS for increasing the heap size and I am still seeing the errors even with a huge amount of memory given to Kafka. I wonder if Kafka 4.0 is not stable yet, it seems like the heap size cannot be adjusted as expected and the memory usage is not predictable

Yes, as you can see, such errors only occur when using `consumer group`. 
I also tried to reproduce these errors on a local device with more memory and they are still here. I'm currently trying to delve deeper into the problem.

petedannemann:
I ended up disabling the tests and made a few small alterations to this PR here https://github.com/segmentio/kafka-go/pull/1384. That is merged now so I will close this PR. Thank you for your work on this

PR #1386, opened 2025-05-14
fix: handle StartOffset when no consumer group set

### LINKED ISSUE #1368: Reader without consumer group read at last offset
**Describe the solution you would like**

Currently the implementation when spinning a consumer without consumer group will always start at the first offset: https://github.com/segmentio/kafka-go/blob/a558bb8629a8f5a920a8e1bb640064ad7efb7cd8/reader.go#L707

Would it be possible to support the `StartOffset` parameter instead?

The goal in here is to spin a consumer without a consumer group idea to get only the latest offsets

zekth:
fix: #1368

This handles startOffset when there's no consumer group set.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -483,8 +483,6 @@ type ReaderConfig struct {
 	// non-zero, it must be set to one of FirstOffset or LastOffset.
 	//
 	// Default: FirstOffset
-	//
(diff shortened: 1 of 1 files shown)


zekth:
@petedannemann could you please review? 🙏

PR #1390, opened 2025-06-08
Add method to fetch messages in batch

### LINKED ISSUE #123: Way to get batch messages and commit if the batch is successful
There is no way to implement the below flow 

1. Read batch messages from Kafka.
2. Process all the messages.
3. If all messages processing succesful - Commit the offset
4. If any of the message processing fails - Do not commit the offset and read the next batch again and try to process the messages and if successful commit the offset.

krsoninikhil:
Since FetchMessage is already reading messages from a fetched batch, this new method just hold the messages util the batchSize number of messages are read.

Fixes #123

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -912,6 +912,68 @@ func (r *Reader) CommitMessages(ctx context.Context, msgs ...Message) error {
 	}
 }
 
+// FetchMessageBatch fetches a batch of messages from the reader. It is similar to
+// FetchMessage, except it blocks until no. of messages read reaches batchSize.
+func (r *Reader) FetchMessageBatch(ctx context.Context, batchSize int) ([]Message, error) {
+	r.activateReadLag()
+	msgBatch := make([]Message, 0, batchSize)
+
+	var i int
+	for i <= batchSize {
+		r.mutex.Lock()
+
+		if !r.closed && r.version == 0 {
+			r.start(r.getTopicPartitionOffset())
+		}
+
+		version := r.version
+		r.mutex.Unlock()
+
+		select {
+		case <-ctx.Done():
+			return []Message{}, ctx.Err()
+
+		case err := <-r.runError:
+			return []Message{}, err
+
+		case m, ok := <-r.msgs:
+			if !ok {
+				return []Message{}, io.EOF
+			}
+
+			if m.version < version {
+				continue
+			}
+
+			r.mutex.Lock()
+
+			switch {
+			case m.error != nil:
+			case version == r.version:
+				r.offset = m.message.Offset + 1
+				r.lag = m.watermark - r.offset
+			}
+
+			r.mutex.Unlock()
+
+			if errors.Is(m.error, io.EOF) {
+				// io.EOF is used as a marker to indicate that the stream
+				// has been closed, in case it was received from the inner
+				// reader we don't want to confuse the program and replace
+				// the error with io.ErrUnexpectedEOF.
+				m.error = io.ErrUnexpectedEOF
+			}
+			if m.error != nil {
+				return nil, m.error
+			}
+
+			msgBatch = append(msgBatch, m.message)
+		}
+		i++
+	}
+	return msgBatch, nil
+}
+
 // ReadLag returns the current lag of the reader by fetching the last offset of
 // the topic and partition and computing the difference between that value and
 // the offset of the last message returned by ReadMessage.


reader.go
```diff
@@ -912,6 +912,68 @@ func (r *Reader) CommitMessages(ctx context.Context, msgs ...Message) error {
...
+			if m.error != nil {
+				return nil, m.error
+			}
+
+			msgBatch = append(msgBatch, m.message)
+		}
+		i++
+	}
+	return msgBatch, nil
+}
```
krsoninikhil, line 975: Duplication of the code can be avoided by calling this method in `FetchMessage`. I'll refactor if once the approach gets reviewed

ghaninia: What happens if the number of messages doesn't reach the desired batchSize?

@krsoninikhil

krsoninikhil: > what happens if one of the messages in the batch fails

We should let consumer decide on how they want to handle it. They can commit or read again from the last commit. Let me know if there is a better approach to handle this.

> if one of message failed, we can ack all messages before failed

This sounds good, we can do this. My only concern if consumer is processing batch by batch, it might confusing behavior that a part of the batch committed it's neither abort nor fully committed.

> What happens if the number of messages doesn't reach the desired batchSize?

I see, if the current code changes look okay, I can add a ticker with a timeout for a maximum wait time. So it would be returning if there are some messages available but not the full batch.

max107: @krsoninikhil in my opinion

> This sounds good, we can do this. My only concern if consumer is processing batch by batch, it might confusing behavior that a part of the batch committed it's neither abort nor fully committed.

It's absolute normal behavior. So if some process can't handle message correctly then should raise panic / return error / stop consuming messages and ack last successfully processed message. Whats next? Restart consumer? Raise panic? It's based on developer decision.

So we can imagine next situation - we receive 3 of 10 (batchSize) messages, we are not exceed deadline, we successfully handle first 2 messages, but failed on 3th message. We know, we can't handler 3 message so we can ack 1 and 2. In next attempt we receive 3,4,5... etc messages and can try again.

> I see, if the current code changes look okay, I can add a ticker with a timeout for a maximum wait time. So it would be returning if there are some messages available but not the full batch.

also sounds very good, because all "batch processing" it's compromise between timeout and batchSize

Sorry, english is not my first language.

max107: So, my simple batch consumer looks like:

```go
package kafkamux

import (
	"context"
	"errors"
	"sync"
	"time"

	"github.com/rs/zerolog/log"
)

var (
	ErrSmallQueueCapacity = errors.New("batch lower than queue capacity")
)

func NewBatchConsumer(
	reader Reader,
	batchSize int,
	duration time.Duration,
) (*BatchConsumer, error) {
	if r, ok := reader.(*kafka.Reader); ok && r.Config().QueueCapacity < batchSize {
		return nil, ErrSmallQueueCapacity
	}

	return &BatchConsumer{
		reader:    reader,
		batchSize: batchSize,
		messages:  make([]kafka.Message, 0, batchSize),
		duration:  duration,
	}, nil
}

type BatchConsumer struct {
	reader    Reader
	batchSize int
	duration  time.Duration
	l         sync.Mutex
	messages  []kafka.Message
}

func (b *BatchConsumer) flush(ctx context.Context, fn handler.BatchCallback) error {
	l := log.Ctx(ctx)

	b.l.Lock()
	defer b.l.Unlock()

	if len(b.messages) == 0 {
		return nil
	}

	if err := fn(ctx, b.messages); err != nil {
		l.Err(err).Msg("error in callback")
		return werr.Wrap(err)
	}

	if err := b.reader.CommitMessages(ctx, b.messages...); err != nil {
		l.Err(err).Msg("error in commit messages")
		return werr.Wrap(err)
	}

	b.messages = make([]kafka.Message, 0, b.batchSize)

	return nil
}

func (b *BatchConsumer) Listen(ctx context.Context, fn handler.BatchCallback) error {
	l := log.Ctx(ctx)

	errCh := make(chan error, 1)

	msgCh := make(chan kafka.Message, b.batchSize)

	ticker := time.NewTicker(b.duration)
	defer ticker.Stop()

	go func() {
		defer close(msgCh)

		for {
			msg, err := fetchMessage(ctx, b.reader)
			if err != nil {
				errCh <- err
				return
			}

			msgCh <- msg
		}
	}()

	for {
		select {
		case readErr := <-errCh:
			l.Err(readErr).Msg("read message error, stop main loop")
			return nil

		case <-ctx.Done():
			l.Debug().Msg("context done, stop main loop")
			return nil

		case <-ticker.C:
			l.Debug().Int("messages_count", len(b.messages)).Msg("ticker flush")
			if err := b.flush(ctx, fn); err != nil {
				l.Err(err).Msg("error flushing messages")
				return werr.Wrap(err)
			}

		case msg, ok := <-msgCh:
			if !ok {
				continue
			}

			b.messages = append(b.messages, msg)

			if len(b.messages) < b.batchSize {
				l.Debug().Int("messages_count", len(b.messages)).Msg("not enough messages, wait")
				continue
			}

			l.Info().Int("messages_count", len(b.messages)).Msg("main loop flush")
			if err := b.flush(ctx, fn); err != nil {
				l.Err(err).Msg("error flushing messages")
				return werr.Wrap(err)
			}

			ticker.Reset(b.duration)
		}
	}
}

```

SeaRoll:
This has been stuck for 1 year. any updates on this?

PR #1405, opened 2025-08-21
[AUTHR-376] Add Oauthbearer support

twleescobar:
Implement the `OAUTHBEARER` mechanism for `SASL` authentication.
This mechanism requires:
- Token
- Zid (Optional)
- Extensions (key value pair map) (Optional)


twmb:
Looks familiar to [my own](https://github.com/twmb/franz-go/blob/master/pkg/sasl/oauth/oauth.go), a fair bit..?

PR #1410, opened 2025-11-13
sasl/aws_msk_iam: update to aws-sdk-go v2

kevinburkesegment:


petedannemann:
We already have a v2, v1 is just left here for backwards compatibility https://github.com/segmentio/kafka-go/tree/main/sasl/aws_msk_iam_v2

PR #1411, opened 2025-11-17
correct lengthRemain handling into message_read.readMessageV2

Dmitriy-Kulagin:
In topic with log compaction the incorrect handling of lengthRemain into readMessageV2 lead to endlessly reading of the same compacted batch. It is happens because skipping offset to batch.lastOffset assumes that lengthRemain ==0 but without this correction lengthRemain became negative and comparison with zero doesn't work.

### DIFF
--- a/message_reader.go
+++ b/message_reader.go
@@ -292,6 +292,10 @@ func (r *messageSetReader) readMessageV2(_ int64, key readBytesFunc, val readByt
 			// stack. here we set the parent count to 0 so that when the child set is exhausted, the
 			// reader will then try to read the header of the next message set
 			r.readerStack.parent.count = 0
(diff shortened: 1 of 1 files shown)


Dmitriy-Kulagin:
I'm pretty sure that this MR can fix several different issues that I've described in this [issue](https://github.com/segmentio/kafka-go/issues/1416)

Dmitriy-Kulagin:
The squasing of commits helped to pass all checks!

Dmitriy-Kulagin:
@petedannemann @hhahn-tw please review it please 🙏  or suggest somebody else who can finally merge it. 
issue [1416](https://github.com/segmentio/kafka-go/issues/1416) has been opened months ago without any reaction.

PR #1431, opened 2026-04-21
fix: describe groups should not fail on an issue with a single group

petedannemann:
## Summary
- https://twilio-engineering.atlassian.net/browse/TUBE-3799
- `DescribeGroups` loops through all groups returned by the broker and decodes member metadata and assignments for each. Previously, a decode failure on any single group's member would cause the entire call to return an error, preventing all other groups from being returned
- This change scopes decode errors to the individual group by setting `group.Error` instead of returning a top-level error. Groups that decode successfully are unaffected by failures in other groups.
- Additionally, when a decode error occurs partway through a group's member list, any previously decoded members are cleared to avoid returning a partial member list alongside an error.

## Testing
- Added below unit tests:
  - `TestDescribeGroupsErrorHandling` — one bad group among three, valid groups still returned with members
  - `TestDescribeGroupsInvalidAssignments` — valid metadata but invalid assignments sets group error
  - `TestDescribeGroupsPartialMembersCleared` — multi-member group where member 2 fails; verifies member 1 is cleared

### DIFF
--- a/describegroups.go
+++ b/describegroups.go
@@ -135,11 +135,13 @@ func (c *Client) DescribeGroups(
 		for _, member := range apiGroup.Members {
 			decodedMetadata, err := decodeMemberMetadata(member.MemberMetadata)
 			if err != nil {
-				return nil, err
+				group.Error = fmt.Errorf("failed to decode member metadata for group %s: %w", apiGroup.GroupID, err)
+				break
 			}
 			decodedAssignments, err := decodeMemberAssignments(member.MemberAssignment)
 			if err != nil {
-				return nil, err
+				group.Error = fmt.Errorf("failed to decode member assignments for group %s: %w", apiGroup.GroupID, err)
+				break
 			}
 
 			group.Members = append(group.Members, DescribeGroupsResponseMember{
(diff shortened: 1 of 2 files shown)


describegroups.go
```diff
@@ -135,11 +135,13 @@ func (c *Client) DescribeGroups(
 		for _, member := range apiGroup.Members {
 			decodedMetadata, err := decodeMemberMetadata(member.MemberMetadata)
 			if err != nil {
-				return nil, err
+				group.Error = fmt.Errorf("failed to decode member metadata for group %s: %w", apiGroup.GroupID, err)
```
pazhang-25, line 138: So with the current implementation, it looks like there will be a partially set `group.Members` list in the case of failures (assuming that there are multiple members and the first member in a given group succeeds, but then another member in this group fails).

Is this the expected behavior that we want, or do we just want to set the `group.members` list empty in the case of even one error? Either way, I think we should write a test for this scenario with multiple members and only one of them has a failure; so for example, maybe one group that has 3 members, where member 1 and 3 are ok, but member 2 has a failure.

pazhang-25 (commented):
Can you please provide a PR description? 😁

describegroups.go
pazhang-25: I've decided to clear the members list b/c since we're just logging the error and skipping the group entirely, there's no point in having a partially maintained list

erushing (approved):

PR #1434, opened 2026-04-26
feat: implement KIP-392 (consumer rack-aware fetching)

GlebShipilov:
## What

Adds support for KIP-392 (follower fetching), which lets a consumer
fetch records from the closest in-rack replica instead of always going
to the partition leader. This cuts cross-AZ traffic and tail latency in
multi-rack Kafka clusters.

### KIP-392 core (wire format + routing)

- `fetch.go`: added `RackID` field on `FetchRequest`,
  `PreferredReadReplica` field on `FetchResponse`.
- `reader.go`: added `Rack` option on `ReaderConfig`. When set, the
  reader passes it through to every fetch.
- `dialer.go`: added `RackID` on `Dialer` so the low-level `Conn` API
  can be rack-aware too.
- `conn.go`: handles the v11 fetch response header, decodes the
  `PreferredReadReplica` field, and exposes it on `Batch`.
- `batch.go`: new `PreferredReadReplica()` accessor on `Batch`.
- `read.go`, `write.go`, `protocol.go`: small plumbing changes to carry
  the new field through the wire format.
- `protocol/fetch/fetch.go`: tagged `RackID` on the request and
  `PreferredReadReplica` on the response per partition for v11+.

### Drain in-flight batch before reconnecting

When the `PreferredReadReplica` in a fetch response differs from the
cached value, upstream would immediately close the batch and connection,
discarding unread records. This PR sets a `prrChanged` flag, delivers
every record from the batch first, then signals `errPreferredReadReplicaChanged`
to the run loop. The reconnect happens between batches, not mid-batch.

### ISR-only replica selection

`dialPreferredReplica()` resolves the preferred replica against `p.Isr`
(not `p.Replicas`). A broker that is in the replica set but out of ISR —
because it is offline or lagging — is skipped. This matches the Java
client's behavior and prevents fetching from a stale follower.

### Clear preferred replica on `NotLeaderForPartition`

When a `NotLeaderForPartition` error arrives while connected to a
preferred follower, the reader now clears the preference and adds the
replica to the cooldown cache before falling back to the leader.
Without this, `initialize()` would re-dial the same failed follower
in a tight loop.

### Clear preferred replica on follower `OffsetOutOfRange`

An `OffsetOutOfRange` from a follower is not authoritative because
follower offsets can lag behind the leader's. The reader now drops
the preference and lets `initialize()` reconnect through the leader,
whose first/last offsets are the only ones trusted for the reset
decision.

### Negative cache for failed preferred replicas

A per-reader `preferredReadReplicaCooldown` map records an "ignore
until" timestamp for each recently-failed replica. Subsequent fetch
responses that re-advertise the same broker id are silently ignored
until the 5-minute cooldown expires. Without this the broker
re-suggests the same dead replica on every fetch and the reader
live-locks in a dial→fail→fallback→re-suggest loop.

### `SeekDontCheck` on follower connections

In `initialize()`, when the reader is about to seek on a KIP-392
follower, it uses `SeekAbsolute|SeekDontCheck` instead of plain
`SeekAbsolute`. This skips the `ListOffsets` validation call, which
followers reject with `NotLeaderForPartition` (only leaders serve
`ListOffsets`). The offset has already been validated by the leader
on the prior fetch.

### Resolve `FirstOffset`/`LastOffset` via leader before follower dial

When the reader needs to resolve `FirstOffset` or `LastOffset` (e.g.
after a group rebalance), it now dials the leader first to call
`ReadOffsets`, materializes the concrete offset, then re-dials the
preferred follower for the actual fetch path. This avoids sending
`ListOffsets` to the follower.

### Follower's `PreferredReadReplica = -1` is a heartbeat

Per KIP-392, `-1` from a follower the reader is already pinned to means
"stay with me" (the follower does not advertise itself as its own
preference). `-1` from the leader means "no preferred replica, stay on
me." The reader now tracks `connectedToFollower` to disambiguate these
two cases: when connected to the follower, `-1` refreshes the TTL
instead of triggering a reconnect. Without this the reader oscillates
between leader and follower every few hundred milliseconds.

## How it was tested

Ran the full test matrix locally against Kafka 2.7.0, 2.8.1, and 3.7.0
(the same versions CircleCI uses):

- `go test -race -cover ./...`
- `go test -tags=unsafe -race -cover ./...`
- `cd sasl/aws_msk_iam && go test -race -cover ./...`

All tests pass except `TestReaderConsumerGroup`, which is a pre-existing
nil-pointer panic on `main` unrelated to this change.

### Chaos testing on a live cluster

Ran a structured chaos protocol against a 10-broker, 3-rack Strimzi
cluster (KRaft mode, RF=3, MIRR=2) with a `go-kafka-consumer` highlevel
consumer reading 10 partitions. The protocol covers every failure mode
the resilience fixes are designed for:

| step                   | action | exercises |
|------------------------|---|---|
| 1 — negative control   | kill broker with 0 partitions in fetch path | baseline sanity |
| 2 — follower-only kill | kill broker we follower-fetch 1 partition from | batch draining, ISR check, cooldown |
| 3 — leader kill        | kill partition leader (we fetch from follower) | `NotLeaderForPartition` clearing, KIP-392 proof point |
| 4 — biggest blast      | kill broker with 5 affected partitions (leader + follower roles) | all of the above under stress |
| 5 — cooldown re-test   | re-kill same broker as step 2 within 5 min | negative cache suppression |
| 6 — KRaft controller   | kill 1 of 3 controllers | quorum resilience (no consumer impact) |
| 7 — full-rack outage   | kill all 5 in-rack brokers simultaneously | cross-rack fallback + re-pin after recovery |

**Results:** zero `Not Leader For Partition` errors, zero
`fetch_message_failed`, zero routing oscillations, all partitions
resumed committing within 60 s of each kill. Full report in
[`KAFKA_OUTAGE_TEST_REPORT.md`](https://github.com/exness/kafka-go/blob/main/KAFKA_OUTAGE_TEST_REPORT.md).

## Notes on the `PreferredReadReplica = -1` edge case

This is the part that's easy to get wrong. Per KIP-392, a value of `-1`
means different things depending on which broker sent it:

- `-1` from the leader means "no preferred replica for you, stay on the
  leader" — the client must drop any cached follower.
- `-1` from a follower that the client is already cached on means
  "stay with me" — the client must keep the cache.

If both cases are treated the same, the consumer oscillates between
leader and follower on every fetch and the optimization is silently
broken. The `connectedToFollower` flag disambiguates the two cases,
matching the Java client behavior.

## Backwards compatibility

The new fields default to zero values, so consumers that don't set
`RackID` see no behavior change. Brokers older than Fetch v11 simply
ignore the field. All resilience mechanisms are inert when
`preferredReadReplica` is `-1` (the default).

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -28,7 +28,11 @@ type Batch struct {
 	partition     int
 	offset        int64
 	highWaterMark int64
-	err           error
+	// preferredReadReplica is the broker id (>=0) the broker would prefer
+	// the consumer fetch from for this partition (KIP-392). -1 means no
+	// preference (fetch from leader).
+	preferredReadReplica int32
+	err                  error
 	// The last offset in the batch.
 	//
 	// We use lastOffset to skip offsets that have been compacted away.
@@ -49,6 +53,17 @@ func (batch *Batch) Throttle() time.Duration {
 // HighWaterMark returns the current highest watermark in a partition.
 func (batch *Batch) HighWaterMark() int64 {
 	return batch.highWaterMark
+}
+
+// PreferredReadReplica returns the broker id of the replica the kafka broker
+// recommends the consumer fetch from for this partition, as introduced in
+// KIP-392 (Allow Consumers to Fetch from Closest Replica). A value of -1
+// means no preference and the consumer should keep fetching from the
+// partition leader. The value is only populated when the connection
+// negotiated Fetch v11 or higher and the broker chose a preferred replica;
+// otherwise it is -1.
+func (batch *Batch) PreferredReadReplica() int32 {
+	return batch.preferredReadReplica
 }
 
 // Partition returns the batch partition.
--- a/conn.go
+++ b/conn.go
@@ -54,6 +54,7 @@ type Conn struct {
 	fetchMinSize  int32
 	broker        int32
 	rack          string
+	clientRack    string
 
 	// correlation ID generator (synchronized on wlock)
 	correlationID int32
@@ -90,6 +91,13 @@ type ConnConfig struct {
 	Partition int
 	Broker    int
 	Rack      string
+
+	// ClientRack is the consumer's rack id used by the broker to determine
+	// the closest replica to fetch from (KIP-392). When set and the broker
+	// supports Fetch v11 or higher, the rack id is sent in fetch requests
+	// and the broker may direct the consumer to read from a follower
+	// replica that is closer to the consumer.
+	ClientRack string
 
 	// The transactional id to use for transactional delivery. Idempotent
 	// deliver should be enabled if transactional id is configured.
@@ -179,6 +187,7 @@ func NewConnWith(conn net.Conn, config ConnConfig) *Conn {
 		partition:       int32(config.Partition),
 		broker:          int32(config.Broker),
 		rack:            config.Rack,
+		clientRack:      config.ClientRack,
 		offset:          FirstOffset,
 		requiredAcks:    -1,
 		transactionalID: emptyToNullable(config.TransactionalID),
@@ -777,7 +786,7 @@ func (c *Conn) ReadBatchWith(cfg ReadBatchConfig) *Batch {
 		return &Batch{err: dontExpectEOF(err)}
 	}
 
-	fetchVersion, err := c.negotiateVersion(fetch, v2, v5, v10)
+	fetchVersion, err := c.negotiateVersion(fetch, v2, v5, v10, v11)
 	if err != nil {
 		return &Batch{err: dontExpectEOF(err)}
 	}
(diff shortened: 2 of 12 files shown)


GlebShipilov:
@petedannemann, can you have a look, please?
I believe rack awareness functionality should be useful for kafka-go library

aartal:
Hi, we're running `segmentio/kafka-go` against Amazon MSK with rack awareness enabled and hit this exact gap. When broker-side rack awareness is switched on, consumers immediately throw `Inconsistent Group Protocol` errors. The existing `RackAffinityGroupBalancer` handles client-side assignment but doesn't cover the KIP-392 wire-level `RackID` in Fetch requests — which is exactly what this PR addresses.

Would love to see this reviewed and merged. Happy to help test against MSK if useful. 🙏

PR #1441, opened 2026-07-10
fix: bound decoded array length to remaining frame bytes

### LINKED ISSUE #1439: Malformed or non-Kafka broker response can cause OOM in protocol.decodeArray
### What version of kafka-go are you using?

  Observed with `github.com/segmentio/kafka-go v0.4.48`.

  I also checked `v0.4.51`, and the relevant decoder path still appears to allocate arrays directly from the decoded length:

  - `protocol.decodeArray` calls `makeArray(elemType, int(n))`
  - `makeArray` calls `reflect.MakeSlice(..., n, n)` or `unsafe_NewArray(..., n)`

  ### What version of Kafka are you testing against?

  Apache Kafka `3.9.0`.

  ### What happened?

  A Kafka broker was misconfigured with `advertised.listeners=PLAINTEXT://localhost:9092`.

  The kafka-go writer was bootstrapped from a remote address, for example:

  ```go
  writer := &kafka.Writer{
      Addr:  kafka.TCP("10.x.x.x:9092"),
      Topic: "test-topic",
  }

  The initial metadata request succeeded, but the broker metadata advertised localhost:9092.

  As a result, kafka-go later tried to connect to localhost:9092 from the client host. In our environment that port was not Kafka; it was another local service. The bytes
  returned by that non-Kafka service were then decoded as a Kafka response.

  Instead of returning a protocol/connection error, the process crashed with OOM.

  Relevant stack:

  fatal error: runtime: out of memory

  github.com/segmentio/kafka-go/protocol.makeArray
  github.com/segmentio/kafka-go/protocol.(*decoder).decodeArray
  github.com/segmentio/kafka-go/protocol.(*Decoder).Decode
  github.com/segmentio/kafka-go/protocol.ReadResponse
  github.com/segmentio/kafka-go.(*Transport).RoundTrip
  github.com/segmentio/kafka-go.(*connGroup).connect
  github.com/segmentio/kafka-go.(*Writer).WriteMessages

  ### Why this looks like a kafka-go robustness bug

  The broker configuration was definitely wrong, but a malformed/non-Kafka response should not be able to make the client allocate an unbounded slice and crash the process.

  In protocol/decode.go, decodeArray reads an int32 length from the response and allocates before validating whether that many elements can fit in the remaining frame:

  func (d *decoder) decodeArray(v value, elemType reflect.Type, decodeElem decodeFunc) {
      if n := d.readInt32(); n < 0 {
          v.setArray(array{})
      } else {
          a := makeArray(elemType, int(n))
          for i := 0; i < int(n) && d.remain > 0; i++ {
              decodeElem(d, a.index(i))
          }
          v.setArray(a)
      }
  }

  makeArray then allocates directly from n.

  ### Expected behavior

  If the response is malformed, truncated, or not a Kafka protocol response, kafka-go should return an error such as invalid response / malformed response / unexpected EOF.

  It should not allocate memory proportional to an untrusted length field before validating it.

  ### Possible fix direction

  Before allocating arrays, validate the decoded array length against one or more bounds, for example:

  - remaining response frame bytes
  - minimum element encoded size where known
  - configured maximum response size
  - a sane protocol-level maximum

  If the length is impossible or unreasonable for the remaining frame, return a decode error instead of allocating.

  ### Additional context

  There is an old PR that protected invalid header counts from causing huge allocations:

  https://github.com/segmentio/kafka-go/pull/319

  This issue seems similar in spirit, but affects general protocol array decoding rather than only message headers.

Socialpranker:
Fixes #1439

## Summary

`decodeArray` and `decodeCompactArray` in `protocol/decode.go` allocate a slice directly from the array-length field read off the wire (`makeArray(elemType, int(n))`), before checking that field against the bytes actually remaining in the frame. As described in #1439, a malformed or non-Kafka response — for example a misconfigured broker whose advertised address happens to route to an unrelated local service — can carry an arbitrary 32-bit length. `reflect.MakeSlice` then tries to allocate an array of that size before a single element has been decoded, which crashes the process with an out-of-memory error instead of returning a decode error.

There's a precedent for this kind of guard in #319, which capped header counts at a fixed constant (1024). This PR uses a tighter, self-adjusting bound instead of a magic number: every encoded Kafka protocol element occupies at least one byte, so a declared array length greater than `decoder.remain` (the exact number of bytes left in the frame) can only come from a corrupt or non-Kafka payload. Both `decodeArray` and `decodeCompactArray` now clamp the length to `remain` via a new `boundArrayLen` helper before allocating. Decoding still proceeds against the clamped length — it will naturally hit a decode error once real elements run out, since the loop already bails out when `d.remain` reaches 0 — rather than allocating unbounded memory up front.

## Test plan

- Added `protocol/decode_test.go` with a test per encoding (`decodeArray` and the compact/flexible `decodeCompactArray`), each constructing a small buffer that declares far more elements than it has bytes for and asserting the resulting array is bounded to the buffer size.
- Added a direct unit test for the new `boundArrayLen` helper covering the boundary cases (equal, over, under, zero remain).
- Ran `go test ./protocol/...`: all existing tests across every protocol message type pass unchanged, confirming no regression to normal (well-formed) decoding.
- Ran `go vet ./protocol/...` and `gofmt -l` on the changed files: clean.
- Did not run the full repository test suite (`go test ./...`), since the rest of the module depends on packages unrelated to this change; the fix and its tests are isolated to `protocol/decode.go`.

### DIFF
--- a/protocol/decode.go
+++ b/protocol/decode.go
@@ -110,8 +110,9 @@ func (d *decoder) decodeArray(v value, elemType reflect.Type, decodeElem decodeF
 	if n := d.readInt32(); n < 0 {
 		v.setArray(array{})
 	} else {
-		a := makeArray(elemType, int(n))
-		for i := 0; i < int(n) && d.remain > 0; i++ {
+		count := boundArrayLen(int(n), d.remain)
+		a := makeArray(elemType, count)
+		for i := 0; i < count && d.remain > 0; i++ {
 			decodeElem(d, a.index(i))
 		}
 		v.setArray(a)
@@ -122,12 +123,25 @@ func (d *decoder) decodeCompactArray(v value, elemType reflect.Type, decodeElem 
 	if n := d.readUnsignedVarInt(); n < 1 {
 		v.setArray(array{})
 	} else {
-		a := makeArray(elemType, int(n-1))
-		for i := 0; i < int(n-1) && d.remain > 0; i++ {
+		count := boundArrayLen(int(n-1), d.remain)
+		a := makeArray(elemType, count)
+		for i := 0; i < count && d.remain > 0; i++ {
 			decodeElem(d, a.index(i))
 		}
 		v.setArray(a)
 	}
+}
+
+// boundArrayLen clamps a decoded array length to the number of bytes left in
+// the frame. Every encoded element occupies at least one byte, so a length
+// greater than remain can only come from a malformed or non-Kafka response;
+// allocating an array of that size would let an untrusted length field drive
+// unbounded memory allocation before a single element is validated.
+func boundArrayLen(n, remain int) int {
+	if n > remain {
+		return remain
+	}
+	return n
 }
 
 func (d *decoder) discardAll() {
--- a/protocol/decode_test.go
+++ b/protocol/decode_test.go
@@ -1,0 +1,70 @@
+package protocol
+
+import (
+	"bytes"
+	"encoding/binary"
+	"reflect"
+	"testing"
+)
+
+// newTestDecoder builds a decoder over data, with remain set to len(data), the
+// same invariant Unmarshal establishes before decoding a response frame.
+func newTestDecoder(data []byte) *decoder {
+	d := &decoder{reader: bytes.NewReader(data)}
+	d.remain = len(data)
+	return d
+}
+
+// TestDecodeArrayMalformedLengthDoesNotOOM reproduces a malformed or
+// non-Kafka response whose declared array length vastly exceeds the bytes
+// actually available in the frame. Before the length is bounded, makeArray
+// allocates directly from the untrusted length field, so a tiny buffer
+// claiming billions of elements can exhaust memory before a single element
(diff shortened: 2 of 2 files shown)


Socialpranker:
Friendly ping — this has been sitting since 10 July with CI green and no conflicts.

It bounds the array length in `decodeArray`/`decodeCompactArray` against the bytes actually left in the frame, so a malformed or non-Kafka response can't drive a huge allocation before anything validates it (#1439, still open).

No urgency if the queue is long — just flagging it's ready whenever someone has time to look. Happy to rebase or split it if that helps review.

Socialpranker:
Same as the sibling PR (#1442) — still green a few weeks later, bumping both together.

Socialpranker:
Closing for now — repo hasn't had a maintainer push in ~4 months despite a sizeable open PR/issue backlog, so I don't expect this to get reviewed soon. Fix is still valid if anyone wants to revive it.

PR #1442, opened 2026-07-10
fix: reject negative record count in v2 record batch

### LINKED ISSUE #1438: kafka-go Fetch response negative record count panic
**Describe the bug**

kafka-go can panic while parsing a broker-controlled Fetch response record batch. In the Fetch response path, a v2 record batch's signed `numRecords` field is accepted after the batch length and CRC checks, then used directly as the length for `make([]optimizedRecord, numRecords)`. A malicious or compromised broker can send a structurally valid Fetch response with `numRecords=-1` and terminate an application that does not recover panics around the kafka-go fetch path. This may have some security efffects.

**Kafka Version**
Pinned ref: 2e0b3968aa51b16beb4e221876499a6ff816cd91

**To Reproduce**

```go
package compress

import (
	"encoding"
	"fmt"
	"io"
	"strconv"
)

type Compression int8

const (
	None   Compression = 0
	Gzip   Compression = 1
	Snappy Compression = 2
	Lz4    Compression = 3
	Zstd   Compression = 4
)

func (c Compression) Codec() Codec {
	if i := int(c); i >= 0 && i < len(Codecs) {
		return Codecs[i]
	}
	return nil
}

func (c Compression) String() string {
	if c == None {
		return "uncompressed"
	}
	return strconv.Itoa(int(c))
}

func (c Compression) MarshalText() ([]byte, error) { return []byte(c.String()), nil }

func (c *Compression) UnmarshalText(b []byte) error {
	if string(b) == "none" || string(b) == "uncompressed" {
		*c = None
		return nil
	}
	i, err := strconv.ParseInt(string(b), 10, 64)
	if err == nil && i >= 0 && i < int64(len(Codecs)) {
		*c = Compression(i)
		return nil
	}
	return fmt.Errorf("compression format must be none or numeric, not %q", b)
}

type Codec interface {
	Code() int8
	Name() string
	NewReader(io.Reader) io.ReadCloser
	NewWriter(io.Writer) io.WriteCloser
}

var (
	_ encoding.TextMarshaler   = Compression(0)
	_ encoding.TextUnmarshaler = (*Compression)(nil)

	Codecs = [...]Codec{
		None:   nil,
		Gzip:   nil,
		Snappy: nil,
		Lz4:    nil,
		Zstd:   nil,
	}
)
```

```sh
#!/usr/bin/env bash
set -euo pipefail

script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
source_dir="$script_dir/source"

UPSTREAM_REPO_URL="https://github.com/segmentio/kafka-go"
UPSTREAM_PINNED_SHA="2e0b3968aa51b16beb4e221876499a6ff816cd91"

_silent() {
  local label="$1"; shift
  local log="$script_dir/.${label}.log"
  if ! "$@" > "$log" 2>&1; then
    echo "[run.sh] $label failed; log follows:" >&2
    cat "$log" >&2
    return 1
  fi
}

setup() {
  if [[ -d "$source_dir/.git" ]] \
     && [[ "$(git -C "$source_dir" rev-parse HEAD 2>/dev/null)" == "$UPSTREAM_PINNED_SHA" ]]; then
    echo "[run.sh] reusing existing $source_dir at $UPSTREAM_PINNED_SHA"
    return 0
  fi
  _setup_inner() {
    rm -rf "$source_dir"
    mkdir -p "$source_dir"
    cd "$source_dir"
    git init -q
    git remote add origin "$UPSTREAM_REPO_URL"
    if git fetch --depth 1 origin "$UPSTREAM_PINNED_SHA" -q 2>/dev/null; then
      git checkout -q FETCH_HEAD
    else
      cd "$script_dir"
      rm -rf "$source_dir"
      git clone -q "$UPSTREAM_REPO_URL" "$source_dir"
      git -C "$source_dir" checkout -q "$UPSTREAM_PINNED_SHA"
    fi
    local head
    head="$(git -C "$source_dir" rev-parse HEAD)"
    if [[ "$head" != "$UPSTREAM_PINNED_SHA" ]]; then
      echo "setup: HEAD=$head but expected $UPSTREAM_PINNED_SHA" >&2
      exit 1
    fi
  }
  _silent setup _setup_inner
  # One-line confirmation reaches the terminal — mirrors the
  # 'Pinned ref:' line in the report README so the maintainer can see
  # exactly which commit they're exercising before the proof emits.
  echo "[run.sh] reproducing against $UPSTREAM_REPO_URL @ $UPSTREAM_PINNED_SHA"
}

build() {
  _silent build bash -c '
    set -euo pipefail
    scratch="$1/inputs/build-scratch/kafka-go"
    rm -rf "$scratch"
    mkdir -p "$(dirname "$scratch")"
    cp -a "$2" "$scratch"
    sed -i "s/^go 1\\.23.*/go 1.22.2/" "$scratch/go.mod"
    cp "$1/inputs/compress_stub.go" "$scratch/compress/compress.go"
    cd "$1/inputs/poc"
    GOTOOLCHAIN=local go build -o "$1/inputs/kafka-go-negative-record-count-poc" .
  ' _ "$script_dir" "$source_dir"
}

trigger() {
  set +e
  "$script_dir/inputs/kafka-go-negative-record-count-poc" > "$script_dir/.trigger.log" 2>&1
  status=$?
  set -e
  if [[ "$status" -eq 0 ]]; then
    echo "trigger: PoC returned successfully; expected panic" >&2
    cat "$script_dir/.trigger.log" >&2
    exit 1
  fi
  grep -F "panic: runtime error: makeslice: len out of range" "$script_dir/.trigger.log" \
    || { echo "trigger: no reproduction proof in run output" >&2; cat "$script_dir/.trigger.log" >&2; exit 1; }
}

cmd="${1:-all}"
case "$cmd" in
  setup)         setup ;;
  build)         setup; build ;;
  trigger|all|"") setup; build; trigger ;;
  *)
    echo "usage: $0 {setup|build|trigger|all}" >&2
    exit 2
    ;;
esac
```

run with
```
bash ./poc/run.sh
```

**Expected Behavior**

No crash with the error handled.

**Observed Behavior**

```text
panic: runtime error: makeslice: len out of range
```

The panic line is the verified oracle fingerprint: it appears only after the crafted Fetch v4 response reaches the v2 record-batch parser and the negative record count is used as a slice length. A setup or build failure would not produce this exact runtime signature.

**Additional Context**

this fix can suppress the panic
```diff
diff -ruN '--exclude=.git' a/protocol/record_v2.go b/protocol/record_v2.go
--- a/protocol/record_v2.go	2026-06-11 19:41:40.381844166 +0000
+++ b/protocol/record_v2.go	2026-06-18 07:51:45.421110964 +0000
@@ -66,6 +66,10 @@
 	dec.reader = buffer
 	dec.remain = recordsLength
 
+	if numRecords < 0 {
+		return Errorf("invalid record batch with negative record count (%d)", numRecords)
+	}
+
 	records := make([]optimizedRecord, numRecords)
 	// These are two lazy allocators that will be used to optimize allocation of
 	// page references for keys and values.
```

Socialpranker:
Fixes #1438

## Summary

`readFromVersion2` in `protocol/record_v2.go` reads the v2 record batch's `numRecords` field (a signed `int32`) and passes it directly into `make([]optimizedRecord, numRecords)`, after the batch length and CRC checks but without validating the count itself. As reported in #1438, a structurally valid Fetch response — one that passes both the length and checksum checks — can still carry a negative `numRecords`, for example from a malformed or compromised broker. That value reaching `make()` as a slice length panics with `runtime error: makeslice: len out of range`, which crashes any process that doesn't recover panics around the fetch path instead of surfacing a decode error.

The reporter's own proof-of-concept confirms the exact panic against the pinned commit, and included a diff proposing the same guard used here.

## Change

Rejects the batch with a regular error as soon as `numRecords` is read to be negative, in the same place the existing length/CRC checks already happen, instead of letting the value reach the allocation.

## Test plan

- Added `protocol/record_v2_test.go`: builds a valid v2 record batch through the existing `WriteTo` path, patches the serialized `numRecords` field to `-1`, recomputes the CRC32 checksum the same way `writeToVersion2` does (so the batch stays otherwise well-formed and passes the length/CRC checks), then asserts `ReadFrom` returns an error instead of panicking.
- Confirmed the new test panics with the reported `makeslice: len out of range` against the pre-fix code, and passes cleanly with the fix — verifying the test actually exercises the bug rather than passing vacuously.
- Ran `go test ./protocol/...`: all existing tests pass unchanged, including `fetch`, `produce`, and `rawproduce`, which exercise record batch encoding/decoding directly.
- Ran `go vet ./protocol/...` and `gofmt -l` on the changed files: clean.
- Did not run the full repository test suite (`go test ./...`); the change and its tests are isolated to record batch v2 decoding.

### DIFF
--- a/protocol/decode.go
+++ b/protocol/decode.go
@@ -110,8 +110,9 @@ func (d *decoder) decodeArray(v value, elemType reflect.Type, decodeElem decodeF
 	if n := d.readInt32(); n < 0 {
 		v.setArray(array{})
 	} else {
-		a := makeArray(elemType, int(n))
-		for i := 0; i < int(n) && d.remain > 0; i++ {
+		count := boundArrayLen(int(n), d.remain)
+		a := makeArray(elemType, count)
+		for i := 0; i < count && d.remain > 0; i++ {
 			decodeElem(d, a.index(i))
 		}
 		v.setArray(a)
@@ -122,12 +123,25 @@ func (d *decoder) decodeCompactArray(v value, elemType reflect.Type, decodeElem 
 	if n := d.readUnsignedVarInt(); n < 1 {
 		v.setArray(array{})
 	} else {
-		a := makeArray(elemType, int(n-1))
-		for i := 0; i < int(n-1) && d.remain > 0; i++ {
+		count := boundArrayLen(int(n-1), d.remain)
+		a := makeArray(elemType, count)
+		for i := 0; i < count && d.remain > 0; i++ {
 			decodeElem(d, a.index(i))
 		}
 		v.setArray(a)
 	}
+}
+
+// boundArrayLen clamps a decoded array length to the number of bytes left in
+// the frame. Every encoded element occupies at least one byte, so a length
+// greater than remain can only come from a malformed or non-Kafka response;
+// allocating an array of that size would let an untrusted length field drive
+// unbounded memory allocation before a single element is validated.
+func boundArrayLen(n, remain int) int {
+	if n > remain {
+		return remain
+	}
+	return n
 }
 
 func (d *decoder) discardAll() {
--- a/protocol/decode_test.go
+++ b/protocol/decode_test.go
@@ -1,0 +1,70 @@
+package protocol
+
+import (
+	"bytes"
+	"encoding/binary"
+	"reflect"
+	"testing"
+)
+
+// newTestDecoder builds a decoder over data, with remain set to len(data), the
+// same invariant Unmarshal establishes before decoding a response frame.
+func newTestDecoder(data []byte) *decoder {
+	d := &decoder{reader: bytes.NewReader(data)}
+	d.remain = len(data)
+	return d
+}
+
+// TestDecodeArrayMalformedLengthDoesNotOOM reproduces a malformed or
+// non-Kafka response whose declared array length vastly exceeds the bytes
+// actually available in the frame. Before the length is bounded, makeArray
+// allocates directly from the untrusted length field, so a tiny buffer
+// claiming billions of elements can exhaust memory before a single element
+// is decoded.
+func TestDecodeArrayMalformedLengthDoesNotOOM(t *testing.T) {
+	buf := make([]byte, 4)
+	binary.BigEndian.PutUint32(buf, uint32(1<<31-1)) // math.MaxInt32 elements
+	d := newTestDecoder(buf)
+
+	v := valueOf(new([]int32))
+	d.decodeArray(v, reflect.TypeOf(int32(0)), (*decoder).decodeInt32)
+
+	if d.remain != 0 {
+		t.Fatalf("expected all bytes consumed, got remain=%d", d.remain)
+	}
+	if n := v.val.Len(); n > len(buf) {
+		t.Fatalf("array length %d was not bounded to the %d bytes available in the frame", n, len(buf))
+	}
+}
+
+// TestDecodeCompactArrayMalformedLengthDoesNotOOM is the compact-encoding
+// (flexible version) counterpart of TestDecodeArrayMalformedLengthDoesNotOOM.
+func TestDecodeCompactArrayMalformedLengthDoesNotOOM(t *testing.T) {
+	buf := []byte{0xff, 0xff, 0xff, 0xff, 0x0f} // unsigned varint for MaxUint32
+	d := newTestDecoder(buf)
+
+	v := valueOf(new([]int32))
+	d.decodeCompactArray(v, reflect.TypeOf(int32(0)), (*decoder).decodeInt32)
+
+	if n := v.val.Len(); n > len(buf) {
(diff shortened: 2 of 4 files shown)


Socialpranker:
Friendly ping — open since 10 July, CI green, no conflicts.

It rejects a negative `numRecords` in `readFromVersion2` before it reaches `make([]optimizedRecord, numRecords)`, which currently panics on a response that passes both the length and CRC checks (#1438, still open).

This is a sibling of #1441 in the same decode path, so they may be easiest to look at together. No rush — happy to rebase whenever.

Socialpranker:
Another ping on this — still green, still no conflicts, just been quiet since late July.

Socialpranker:
Same as #1441 — closing for now given the repo's current inactivity. Fix is still valid if useful to reopen later.

Issue #1358, opened 2024-12-18
Goroutines left behind after writing message even after closing writer


fjagugar:
**Describe the bug**

While debugging a high CPU consumption in my project, I noticed that each time an event is sent to Kafka, the number of goroutines running in my project was increasing.

The code sending the event is:

```
func (k *Client) Push(ctx context.Context, topicName string, message *kafka.Message) error {
	writer := &kafka.Writer{
		Addr:  kafka.TCP(k.brokers...),
		Topic: topicName,
		Balancer: &kafka.Murmur2Balancer{},
		Transport: &kafka.Transport{
			TLS: k.dialer.TLS,
		},
	}

	defer func() {
		err := writer.Close()
		if err != nil {
			logger.Warnf("Error closing writer for topic %s: %v", topicName, err)
		}
	}()
	return writer.WriteMessages(ctx, *message)
}
```

After running that, I see in Intellij debug console that the number of goroutines increases:

![image](https://github.com/user-attachments/assets/d4898a5b-eac8-49aa-a9e3-c2cc0b6ad6a0)

The more events are sent, the more goroutines keep running.

Checking the code in the transport file I see that a pool of connections is initialized with a `refc: 2` [here](https://github.com/segmentio/kafka-go/blob/main/transport.go#L240). When calling to the `ref()` function, it is increasing that value [here](https://github.com/segmentio/kafka-go/blob/main/transport.go#L322). When calling the `unref()` function, it [decreases the value](https://github.com/segmentio/kafka-go/blob/main/transport.go#L326) and if it's 0 it properly [closes the context of that goroutine](https://github.com/segmentio/kafka-go/blob/main/transport.go#L335). The `unref()` function is properly called when [closing the transport connections](https://github.com/segmentio/kafka-go/blob/main/transport.go#L140).

Adding breakpoints I see that, when writing an event, a pool is created, the code calls to `unref()`, `ref()`, `unref()`, `ref()`, `unref()`, leaving the value of refc as `1` but never properly canceling the context and leaving the discover goroutine running forever. On each write operation, there's a new pool created and a new discover goroutine  left behind running.

I suspect that `refc` attribute is being used for tracking if the pool is being used or not. I don't fully understand why the code is using a `uintptr` for this instead of a `int` or any other type nor why it is being initialized with `2`.

**Kafka Version**

[...]

PR #1447, opened 2026-08-11
Document Transport lifecycle ownership on Client, Writer, and Transport

om7057:
## Summary

Fixes the documentation gap behind #1358 ("Goroutines left behind after writing message even after closing writer").

The leak reported in #1358 isn't a bug in the connection pool's ref-counting (`transport.go`'s `refc`/`ref`/`unref`), that logic is correct. The actual cause is that `Writer.Close()` only calls `CloseIdleConnections()` on a `Transport` it created *internally*, via the deprecated `NewWriter(WriterConfig)` constructor. A `Writer{Transport: &kafka.Transport{...}}` built via struct literal (the common, non-deprecated way, and the way the bug report's repro does it) never gets its `Transport` cleaned up by `Close()`, so the background `discover()` goroutine that `Transport` starts runs forever unless the caller calls `CloseIdleConnections()` themselves.

This turns out to be intentional design, confirmed by @achille-roussel in #599 (2020): `Transport` follows the same ownership model as `net/http.Transport`, long-lived, meant to be shared and reused across many `Writer`/`Client`/`Reader` values, with the caller responsible for closing it. The README already documents the correct "shared transport" pattern (in the SASL section), but:
- none of the godoc comments on `Transport`, `Writer.Transport`, `Client.Transport`, or `CloseIdleConnections` state this contract, and
- the README's plain "Direct Writer creation" example (right above the SASL section) creates a fresh `&kafka.Transport{}` inline with no warning, modeling the exact anti-pattern from the bug report.

## What this PR does

- Adds doc comments to `Transport`, `Transport.CloseIdleConnections`, `Client.Transport`, `Writer.Transport`, and `Writer.Close` making the ownership/lifecycle contract explicit (mirrors `net/http`; caller must call `CloseIdleConnections` or reuse a shared `Transport`).
- Fixes the README's "Direct Writer creation" example to construct and share a `Transport` the same way the SASL examples already do, instead of silently modeling the leak.

## What this PR intentionally does not do

- No behavioral/code change to `Transport`, `Writer`, or `Client`. An automatic-cleanup change would require guessing at `Transport` ownership, which the existing design (per #599) deliberately avoids since `Transport` values are meant to be shared across multiple owners. I scoped that out rather than attempt a heuristic that could silently break shared-Transport usage.

Updates #1358 (documentation-only; does not change runtime behavior).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` on changed `.go` files (clean)
- [x] Ran the full test suite locally against a live broker via `docker-compose up` (`KAFKA_VERSION=3.7.0 go test -cover ./...`); pre-existing failures (`TestConn`, `TestReaderConsumerGroup/verify_offset_committed`) are unrelated flakes from single-node broker auto-topic-creation races, present before this change.
- [x] Ran `TestWriter`/`TestClient`/`TestTransport` explicitly against a live broker after the change - all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,13 +570,21 @@ r := kafka.NewReader(kafka.ReaderConfig{
 Direct Writer creation
 
 ```go
+// Transports are responsible for managing connection pools and other
+// resources, so it's generally best to create a single Transport and share
+// it across the Writer (and Client) values in your application, rather than
+// creating one per Writer. A Transport that is never shared and never
+// closed (via its CloseIdleConnections method) will leak the goroutine it
+// starts to discover the cluster layout.
+sharedTransport := &kafka.Transport{
+    TLS: &tls.Config{},
+}
+
 w := kafka.Writer{
     Addr: kafka.TCP("localhost:9092", "localhost:9093", "localhost:9094"), 
     Topic:   "topic-A",
     Balancer: &kafka.Hash{},
-    Transport: &kafka.Transport{
-        TLS: &tls.Config{},
-      },
+    Transport: sharedTransport,
     }
 ```
 
--- a/client.go
+++ b/client.go
@@ -41,6 +41,12 @@ type Client struct {
 	// A transport used to communicate with the kafka brokers.
 	//
 	// If nil, DefaultTransport is used.
+	//
+	// The Client never closes the Transport; the caller retains ownership of
+	// it and is responsible for calling its CloseIdleConnections method once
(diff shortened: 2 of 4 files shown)


om7057:
@petedannemann tagging you here since this PR addresses the root cause related to Transport/Writer lifecycle discussions. Would appreciate a look when you have a chance.

